PR #9, opened 2017-06-07
Updated and refactored how items and characters are chosen for intera…

Faerdan:
…ctions in the world. The new system has more precision, with items under the cursor taking priority (this helps with small items which are in front of large items). A new method CanInteractWith(Item) has been added to the Character class which does all validation of interaction with items, including new distance calculations based on bounding boxes. This gives more consistency in selecting and interacting with all sizes and shapes of items. Because of this pickdistance has been removed from all items (in their XML config files) and the default PickDistance for all items is now 120.

### DIFF
--- a/Barotrauma/Content/Items/Artifacts/artifacts.xml
+++ b/Barotrauma/Content/Items/Artifacts/artifacts.xml
@@ -3,8 +3,7 @@
   <Item
     name="Skyholder Artifact"
     category="Alien"
-    Tags="alien"
-    pickdistance="150">
+    Tags="alien">
 
     <Sprite texture="artifact.png" depth="0.7" sourcerect="59,0,60,61"/>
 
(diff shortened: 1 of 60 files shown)


Faerdan:
JAG (on Discord) found an issue, the Hatch button inside of the docking hatch on the Aegir cannot be clicked, for some reason on my implementation the physics Raycast is failing to hit the button item. I will look into this further in the morning.

Regalis11:
It seems that the issue on Aegir is due to the button being in a different submarine (the docked shuttle). CanInteractWith doesn't work correctly if the item is not inside the same sub as the character, so now it's also impossible to press airlock buttons from the outside. Transforming the position of the item to the same coordinate system as the player before doing the raycast should fix this, something along the lines of:
`if (Submarine == null && item.Submarine != null)
{
    //character is outside, item inside
    itemPos += item.Submarine.SimPosition;
}
else if (Submarine != null && item.Submarine == null)
{
    //character is inside, item outside
    itemPos -= Submarine.SimPosition;
}
else if (Submarine != item.Submarine)
{
    //character and the item are inside different subs
    itemPos += item.Submarine.SimPosition;
    itemPos -= Submarine.SimPosition;
}`

Faerdan:
Ah brilliant, thanks for clearing that up.

This week got very busy, but I should be able to implement the fix this evening.

Faerdan:
I am going to create a new pull request for this update.

PR #14, opened 2017-06-18
Updated and refactored how items and characters are chosen for intera…

Faerdan:
…ctions in the world. The new system has more precision, with items under the cursor taking priority (this helps with small items which are in front of large items). A new method CanInteractWith(Item) has been added to the Character class which does all validation of interaction with items, including new distance calculations based on bounding boxes. This gives more consistency in selecting and interacting with all sizes and shapes of items. Because of this pickdistance (not interactdistance) has been removed from all item configurations (in their XML config files) and the default InteractDistance for all items is now 120.


Faerdan:
The change which allows interactions with items not on the same submarine has been implemented. This is not ready to merge.

PR #55, opened 2017-12-02
fix start pop up saying the host is the target if host becomes traitor

Crystalwarrior:
Right now, if you're hosting and you toggled "play yourself", if you become the traitor the target on the pop-up will be you instead of your actual target, though the server logs will correctly tell you your actual target.


Crystalwarrior:
Also made this partly to test pull requestings for some helpful developing shazam

PR #79, opened 2017-12-05
Chat tweaks and fixes

Crystalwarrior:
* Prevents the chat keybind from deactivating the chat
* Allows the use of chat keybind during lobby
* Prevents server from receiving messages from unconscious clients. Previously it was possible to keep the chat box open and send messages while KO'd if you compiled yourself.
* Changes the private message color to be toxic green for a distinction from radio chat
* Adds a [PM] tag to private messages for an even easier distinction

default


Crystalwarrior:
This is merge-ready

PR #81, opened 2017-12-05
Adds ragdoll button. Because why not.

Crystalwarrior:
Also adds admin ragdoll command to force people into ragdoll mode.
The networking/syncing wasn't yet tested for this but hopefully everything should work.
Updates config.xml to add default keybinds for radio chat and ragdoll

TODO: Decrease impact resistance when stunned/ragdolled/unconscious
Make it easier to carry ragdolls through stairs for security purposes
EDIT: Turns out both could be considered bugs. Fixed them both in the latest commit.

### DIFF
--- a/Barotrauma/BarotraumaClient/Source/Characters/Animation/Ragdoll.cs
+++ b/Barotrauma/BarotraumaClient/Source/Characters/Animation/Ragdoll.cs
@@ -13,7 +13,7 @@ namespace Barotrauma
         {
             float volume = Math.Min(impact - 3.0f, 1.0f);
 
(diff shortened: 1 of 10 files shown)


Crystalwarrior:
Holy shit as a side effect of the fixed ragdoll things, impact damage for ragdolls is increased as it should be.
However, there's also a side benefit - if you lay down before impact when sinking, your damage will be reduced since you won't bounce all over the place as much. It's a viable strategy!

Crystalwarrior:
Fixed carrying unconscious people being a pain, dunno what else to add at this point.
This is merge-ready

PR #123, opened 2017-12-10
Syringe gun + projectile tweaks + several status effects/item components tweaks

Crystalwarrior:
***WARNING: REQUIRES https://github.com/Regalis11/Barotrauma/pull/120 TO BE MERGED FIRST!***

Syringe Gun woot!!
default

https://puu.sh/yDv6Z/ff91294280.webm
Accepts syringes as ammo. Has custom sprite and sound.
Made syringes have a "projectile" component for syringe guns
Adds several missing signal components to the fabricator, as well as medical syringe to medical fabricator
Adds support for status effects applied to characters from onImpact
Removed "contained.Condition = 0.0f;" in Projectile.cs due to it limiting the possibilities of the .xml content creation
Added a couple null checks in StatusEffect.cs due to crashes caused by the new projectile changes - might need to debug further, perhaps there's a root cause?
Added itemlist console command


Crystalwarrior:
Updated the sprite to show the syringes inside
tweaked a few things
Fixed crash bug with itemlist command

Issue #187, opened 2017-12-20
"InWater" status effect triggers on ledges/platforms


Crystalwarrior:
When making potassium explode when submerged I encountered this issue:
https://puu.sh/yKlZv/53e6093a5d.webm
volvo, pls fix

PR #188, opened 2017-12-20
Motion detectors, ducts, more chem effects and chems, fabricator/deconstructor overhaul, more footsteps, clown hitsounds...

Crystalwarrior:
Adds:
* Liquid Oxygenite now has OnImpact explosion that's weaker than Oxygenite Shards but still can toss you around quite a bit. Throwing two bottles at someone's feet might also completely destroy them: https://puu.sh/yKskB/33d0abced2.webm
https://puu.sh/yKso8/b5d911a5a7.webm
https://puu.sh/yKsCb/52d32c1611.webm
* Liquid Oxygenite is no longer craftable in medical fabricator, however, it can now be deconstructed. Reason behind this is that oxygenite shards are way too precious to just get *one* liquid oxygenite from them.
* However, Liquid Oxygenite now takes 0.5 HP per tick as opposed to 0.1 so stacking it is far more lethal.
* Added a new poison called Amanitin. It is a poison with a delayed effect. On initial injection it takes -5 health and -15 oxygen, however this is only more of a tell for the victim than anything. But after 20 seconds it takes effect and causes -10 oxygen and -1 health for the duration of 60 seconds, stopping oxygen regeneration on first injection and being lethal with more injections. Best countered by Liquid Oxygenite.
Amanitin is craftable from Erythrozine, Flash Powder and Sulphuric Acid with 60 medical skill
* Made Calyxanide reduce the infection instead of straight-up curing it so it might take the whole bottle to cure husk infection at later stages.
* Nerf Auxiliorizine (aka super regenerative godmode chemical) slightly.
* Made Uranium Powder induce rapid bleeding instead of being a Morbusine reskin
* Nerfed Morbusine to have same stats as original Uranium Powder
* Made Fibrinozine cure a bit of health as well to make it counteract the bleeding a tiny bit. It's not much but it helps.
* Adds Motion Detector, Duct Blocks and Switch based on Barotrauma Extended
* Fixed monsters who are not supposed to bleed still taking bleeding damage from poisons, etc.
* Add Cause of Death to Health HUD
* Get bike horns from clown masks and clown suit because nothing related to clowns makes sense.
* So many more things it'll be hard to list them all in here

Bugs:
* ~~Delayed status effects only care about the last "character", previous "character" is ignored. For example, injecting a delayed poison in 4 people will only take effect on the 4th person.~~ Fixed by overhauling DelayedEffect
* ~~InWater is completely bugged https://github.com/Regalis11/Barotrauma/issues/187~~ Managed to fix this one


Barotrauma/BarotraumaShared/Source/Items/Item.cs
```diff
@@ -777,11 +777,13 @@ public AttackResult AddDamage(Character attacker, Vector2 worldPosition, Attack
 
         private bool IsInWater()
         {
+            if (parentInventory != null && parentInventory.Owner != null)
+                FindHull();
```
Regalis11: The update method keeps calling FindHull as long as the item is moving, so why is this necessary? Now all items, including immovable ones without a body, keep calling it constantly (twice per frame if moving).

Crystalwarrior: Yeah this might not be neccessary at all now that I think about it.
The only reason I assumed this was neccessary is if it used incorrect hull, however I'm not sure if that's even the case.

Regalis11: Disregard my previous comment actually, I didn't notice it's only done for items in an inventory. 😅 In any case I don't think this is an appropriate place for the hull checks, IMO it would make more sense to ensure the hull is up-to-date in the Update function.

Crystalwarrior:
Forgot to mention but fixed monsters who are not supposed to bleed still taking bleeding damage from poisons, etc.

Crystalwarrior:
This is merge-ready

Crystalwarrior:
TODO: Item combination tag so you can combine chems, etc.

Crystalwarrior:
TODO: a conditional for xml

Barotrauma/BarotraumaShared/Source/Characters/StatusEffect.cs
```diff
@@ -209,6 +225,80 @@ public virtual bool HasRequiredItems(Entity entity)
...
+                    float? floatValue = null;
+                    if ((type == typeof(float) || type == typeof(int)) && (property.GetValue() is float || property.GetValue() is int))
+                    {
+                        floatValue = Convert.ToSingle(value);
+                    }
+
+                    switch (op)
+                    {
+                        case "==":
+                            if (!(property == value))
```
Regalis11: Shouldn't this be comparing the value of the property and not the property itself? Also, you should note that comparing boxed value types won't correctly if you do it this way. Meaning that something like ((object)2) == ((object)2) would equate to false.

https://stackoverflow.com/questions/6205029/comparing-boxed-value-types

Barotrauma/BarotraumaShared/Source/Characters/StatusEffect.cs
```diff
@@ -138,7 +141,20 @@ protected StatusEffect(XElement element)
                             " - sounds should be defined as child elements of the StatusEffect, not as attributes.");
                         break;
                     default:
-                        propertyAttributes.Add(attribute);
+                        object attributeObject = XMLExtensions.GetAttributeObject(attribute);
+                        Type type = attributeObject.GetType();
+
+                        string op = ((string)attributeObject).Substring(0, 2);
```
Regalis11: This will throw an exception if the value of the attribute is a number, a boolean or a too short string. Also, you should probably use GetAttributeString because you're expecting the value to be a string.

Barotrauma/BarotraumaShared/Source/Characters/StatusEffect.cs
```diff
@@ -32,6 +32,8 @@ public enum TargetType
         public string[] propertyNames;
         private object[] propertyEffects;
 
+        List<Tuple<string, string, object>> propertyConditionals;
```
Regalis11: I'd rather implement the conditionals as a separate class that handles the comparisons, operator parsing, etc. Would make it easier to reuse the conditional logic and List<PropertyConditional> would be less of an eyesore than List<Tuple<string, string, object>>. :)

Crystalwarrior: Yeah for sure. Ideally it should be like an <if target="Character" Health="< 50">, however that's problematic as it doesn't let you apply StatusEffect things like Health="50", etc.

Maybe turn it into its own class and add a new component that does exactly that? Since <if> component would be EXTREMELY useful for cases like this:
```xml
<if target="This" Condition="> 0.0">
  <StatusEffect="Always" Condition="-0.1"/>
</if>
<if target="This" Condition="<= 0.0">
  <StatusEffect="Always" .../>
</if>
```

Barotrauma/BarotraumaShared/Source/Characters/StatusEffect.cs
Crystalwarrior: So I take it I should use .Equals(value) then?

Barotrauma/BarotraumaShared/Source/Characters/StatusEffect.cs
Crystalwarrior: But I can't use GetAttributeString because "attribute" not an XElement, it's an XAttribute. Maybe I should do something like `string op = (attributeObject.ToString()).Substring(0, 2);` instead?

Regalis11: Ah yeah you're right, GetAttributeString won't work here. There's still no reason to parse the value of the attribute to an object and then cast it back to a string - if you want the string representation of the value you could simly use attribute.Value. And make sure it's long enough before calling substring.

Barotrauma/BarotraumaShared/Source/Characters/StatusEffect.cs
Regalis11: Yep, along with a null check to make sure calling Equals doesn't throw an exception.

Barotrauma/BarotraumaShared/Source/Characters/StatusEffect.cs
Regalis11: It shouldn't need to apply any StatusEffect things, the StatusEffect could simply call the relevant method in the PropertyConditionals to check if all the conditions are fullfilled before doing its StatusEffect business.

I'm not sure if there's a need for a separate component that only does those conditional checks - I think you could get the same result by simply putting the conditional and the StatusEffect inside any ItemComponent or use an empty base ItemComponent if there's no need for any additional ItemComponent functionality.

Crystalwarrior: I suppose so, yeah. I think my current implementation will need some testing to determine if it's robust enough.

Crystalwarrior:
OH GODDAMNIT.

```xml
    <Throwable canBeCombined="true" slots="Any,RightHand,LeftHand" throwforce="4.0" aimpos="35,-10">
      <StatusEffect type="OnUse" target="Character" Health=">50" Health="-2" duration="10">
        <RequiredItem name="Medical Syringe" type="Container"/>
      </StatusEffect>
      <StatusEffect type="OnUse" target="Character" Health="<=50" Health="-5" duration="10">
        <RequiredItem name="Medical Syringe" type="Container"/>
      </StatusEffect>
    </Throwable>>
```

Barotrauma/BarotraumaShared/Source/Characters/StatusEffect.cs
Regalis11:
```
<if target="This" Condition="> 0.0">
  <StatusEffect="Always" Condition="-0.1"/>
</if>
```
I don't mean there's anything wrong with this approach, IMO it'd be a more clear way to define the conditionals and would take care of those "duplicate attribute" errors. I just meant the conditionals don't need to be a separate ItemComponent type, code-wise they could work just like they do now even if they were defined as separate XML elements.

Crystalwarrior:
> Illegal characters in XML attributes
> 
> Some special characters are not permitted in XML attribute values.
> 
> Note that the ampersand (&) and less-than (<) characters are not permitted in XML attribute values. Since XFDL computes appear in a compute attribute, these must be escaped with character or entity references (e.g. the entity references &amp; for the ampersand and &lt; for the less-than character). Hence, the less-than-or-equal symbol (<=) could be encoded as ' &lt;='.

YOU ARE TEARING ME APART XML

Crystalwarrior:
@Regalis11 I'll need help figuring out how to check for every public variable on a target so you can check for things like SpeciesName with Conditionals. Will that require adding things to SerializableProperties list, and what will be the implications of exposing so much info?

Crystalwarrior:
OH WOW, I haven't even realised it but removing coroutines and using UpdateAll instead fixed the chemicals not being affected by pausing the game!

Regalis11:
> @Regalis11 I'll need help figuring out how to check for every public variable on a target so you can check for things like SpeciesName with Conditionals. Will that require adding things to SerializableProperties list, and what will be the implications of exposing so much info?

I can think of a few options: either expanding SerializableProperties to support fields in addition to properties or changing the fields you want to check into properties. I think the latter would be the preferable option, because it gives us more control over which values the modders are allowed to mess with. Giving everyone access to any public field would most likely cause a ton of bugs since the usage or implications of some of the fields may not be that easy to understand without delving deeper into the code.

Another option would be a separate reflection-based system that only allows getting the values of the fields but not modifying them - that'd make them somewhat safer to use in conditionals. However, I'm still leaning towards changing the fields into serializable properties. No need to implement any additional logic, and we could easily choose which values the modders are allowed to access.

Barotrauma/BarotraumaShared/Source/Characters/StatusEffect.cs
```diff
@@ -272,6 +262,66 @@ protected StatusEffect(XElement element)
...
+                            string[] splitString = atStr.Split(' ');
+                            string op = splitString[0];
+                            if (splitString.Length > 0)
+                            {
+                                for (int i=1; i<splitString.Length; i++)
+                                {
+                                    atStr = splitString[i] + (i > 1 && i < splitString.Length ? " " : "");
+                                }
+                            }
+                            //thanks xml for not letting me use < or > in attributes :(
```
Regalis11: How about using InnerXML for this? Something like `<Conditional>variable>=5</Conditional>`

Crystalwarrior: I'm not sure even that'll work, honestly.

Crystalwarrior:
Just noticed a new, strange bug happening: https://puu.sh/yPuai/eca2a059e2.webm
Problem is, this should've been happening before even regardless of my changes. So why did it only start happening now? That'll have to be looked into.
The reason why it should be happening in theory is that in character's UpdateOxygen method, it takes hull oxygen and OxygenAvailable. If OxygenAvailable was 0, it'd take the hull, which would likely be 100. If OxygenAvailable was -100, it would take the hull, which would make OxygenAvaialbe 0.
However, it seems like something happened and instead of setting the property it now decreases it, and I don't know why? In theory it should SET OxygenAvailable to -100 but it seems like its performing a substraction now.

Crystalwarrior:
TODO: Test if booleans work with conditionals. I never typecast booleans, so probably not.

Crystalwarrior:
https://puu.sh/yQ8fs/61e12f15c6.webm
yes.

Crystalwarrior:
From what I tested this PR is now merge-ready, however most of the testing was done in single-player.

Barotrauma/BarotraumaClient/Source/Sounds/Sound.cs
```diff
@@ -87,6 +87,10 @@ public static Sound Load(string file, bool destroyOnGameEnd = true)
                 DebugConsole.ThrowError("File \"" + file + "\" not found!");
                 return null;
             }
+
+            Sound dupe = Sound.loadedSounds.Find(s => s.filePath == file);
```
Regalis11: Why is this needed? Won't it cause problems if a sound is removed but other objects are still using the same sound?

Crystalwarrior: I believe this is necessary due to me getting rid of DamageSoundType and using tags instead so as not to recreate the same sound multiple times in case that causes memory leaks or something.
I'll admit tho I don't exactly know what I'm doing when it comes to that

Regalis11: I'm not sure how replacing the DamageSoundType enums with strings would require this change, as the damage sounds are still loaded and freed the same way as before. Sound objects don't actually hold the audio data anyway, they only reference the OggSound that holds it and sounds sharing the same clip share a reference to the same OggSound. So basically I think the issue you're trying to solve has already been taken care of, and this change only adds a new issue because removing a sound will break other sounds sharing the same audio clip.

PR #359, opened 2018-04-08
Make Captain's Jacket able to hold 5 things

Kampfkarren:
It's stupid that all jobs can hold stuff in their jumpsuits except Captain.


Regalis11:
The slots in clothes may be removed altogether and replaced with a job-specific toolbox/medkit/tool belt (partially implemented in the dev branch), but for now this is a useful change even if we may not end up needing it in the end.

PR #401, opened 2018-05-07
Added oxygenite tank, ancient weapon changes

BlueTheKing:


Regalis11:
I'm not sure if it's a good idea to prevent players from using oxygenite shards as a replacement for oxygen tanks when exploring ruins. Oxygenite tanks and the option to use them as plasma cutter fuel make sense though.

BlueTheKing:
I was thinking of making it a tank since breathing out of the shard didn't make much sense to me.
I couldn't get the cutter to work so i didn't add that one.

ClasticM:
Empty tank + Oxygenite shard = craftable Oxygenite tank?

BlueTheKing:
My idea was for you to have to deconstruct the shard and get the liquid oxygenite and steel bar to then craft the tank.

ClasticM:
Nice. Even better idea

BlueTheKing:
Im still unable to get the plasma cutter to use the oxygenite tank, seems bugged or something. If anyone else is willing to they can add it.

ClasticM:
I can look at it later for you

ClasticM:
Do you get any errors?

BlueTheKing:
Its as if it doesn't save the changes i made, keeps saying you need an oxygen tank.

Kampfkarren:
where's the ancient weapon changes? i don't see anything

BlueTheKing:
Its the first file edited and its mentioned in [this](https://github.com/Regalis11/Barotrauma/pull/401/commits/100d1176f8ff93bfae894c1f488e072f1a19cf91) changelog.

BlueTheKing:
Yeah true, thinking smart about shards in this way would pay off.
But since currently the ancient weapon basically destroys itself after using up a shard, i don't think thats needed just yet.

PR #409, opened 2018-05-17
Removes Awful Game Mode

chesse20:
Every round of this is just a greifer murdering you and your team and then not being able to respawn in the respawn shuttle because theres no respawn shuttles during combat missions so you just have to sit around while you watch as the enemy team struggles to figure out how to go back to their home base for 20 minutes.

### DIFF
--- a/Barotrauma/BarotraumaShared/Content/Missions.xml
+++ b/Barotrauma/BarotraumaShared/Content/Missions.xml
@@ -225,7 +225,7 @@
     </Items>
   </CargoMission> 
   
-  <CombatMission
+  <!--<CombatMission
     name="Search and Destroy"
     
     descriptionneutral="A Coalition vessel has been dispatched from [location1] to destroy a renegade vessel at the outskirts of [location2]."
@@ -244,5 +244,5 @@
     
     <message header="Crew eliminated" text="The renegade crew has been defeated! The Coalition vessel has to make its way back to [location1]."/>
     <message header="Crew eliminated" text="The Coalition crew has been defeated! The renegade vessel has to make its way back to [location2]."/>
-  </CombatMission>
+  </CombatMission> -->
 </Missions>


Kampfkarren:
░░░░░░░░░░░░▄▄                     
░░░░░░░░░░░█░░█                  
░░░░░░░░░░░█░░█                  
░░░░░░░░░░█░░░█                  
░░░░░░░░░█░░░░█                  
███████▄▄█░░░░░██████▄ 
▓▓▓▓▓▓█░░░░░░░░░░░░░░█ 
▓▓▓▓▓▓█░░░░ Regalis, ░░░░░█ 
 ▓▓▓▓▓▓█░░░░Add This░░░░█  
▓▓▓▓▓▓█░░░░Please!░░░░░█ 
▓▓▓▓▓▓█░░░░░░░░░░░░░░█ 
▓▓▓▓▓▓█████░░░░░░░░░█   
██████▀░░░░▀▀██████▀

Regalis11:
How about playing other game modes instead? I agree that the combat missions could use some rethinking. Having to watch the other team return back to base is not fun at all and there's too much downtime for many of the players in general. But I wouldn't go as far as removing it - personally I've had some of the most entertaining rounds playing combat missions on (griefer-free) private servers.

Regalis11:
I understand your frustration, but I feel at the moment it might be a good idea to complain to the admins or moderators of those servers. BlueTheKing's idea is also a way better way to deal with this than removing the combat missions altogether.

BlueTheKing:
Does that mean it could be a possible feature in the future?

chesse20:
i think the idea of just ending the round when one team wins would fix this really well

Regalis11:
@BlueTheKing I'm on it (see f931d81)
@chesse20  Or maybe both? I think there should be the option to disable combat missions because of how frustrating they can be on poorly moderated public servers even if the boring "return back to base" part is removed.

PR #425, opened 2018-05-24
Diving suit Locker, Fire extinguisher brackets

ClasticM:
divesuitlocker

Adds a small locker that can contain a (1) Diving suit. Diving suit locker has hideitems="false" to visually signify if the locker has a suit or not.

divesuitlocker

Also adds a fire extinguisher bracket and horizontal facing wall mount, which can be seen pictured in the above photo. Useful for making this equipment immediately available in sensitive areas that would require them  

extinguisherbracket

### DIFF
--- a/Barotrauma/BarotraumaShared/Content/Items/Containers/containers.xml
+++ b/Barotrauma/BarotraumaShared/Content/Items/Containers/containers.xml
@@ -41,6 +41,17 @@
       <Containable name="medical"/>
     </ItemContainer>
   </Item>
+   
+  <Item
+    name="Dive Suit Locker"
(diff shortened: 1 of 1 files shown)


BlueTheKing:
I think keeping to just the bracket or mount would be good since having two items that do the same thing is kind of unnecessary.

ClasticM:
The horizontal facing mount is to give sub makers more options if they run out of room. It also makes it possible to put it on top of terminals or under beds

BlueTheKing:
Oh, in that case sure.

JJRcop:
This pull request has a strange lack of sprites

PR #437, opened 2018-06-05
Allow the `spawnitem` command to give items directly to the inventory of a specific player

### LINKED ISSUE #436: Give item to specific player via commands
Currently, there is no way for an admin or a someone with access to the server CLI to give a specific player an item.  The closest thing that can be managed is using the `cursor` feature to drop the item at their feet, a solution which

 * Allows other players to intercept the item
 * Doesn't guarantee that the player will actually notice and pick up the item
 * Can't be done without visually seeing the player (i.e. needs to be done while logged in using the client and in sight of the player)

I propose that a command is added or modified to allow server operators to add an item directly to a player's inventory.  For example, if the `spawnitem` command were modified to take a clientid as an alternative to `inventory` or `cursor`, an admin could give an oxygen tank to a player with clientid 3 using the command `spawnitem oxygen tank 3`.  Of course, there are other ways of implementing this, such as adding a new command, or giving items based on name instead of ID, but this is just an example.

Tazial:
This is one possible implementation of a solution to #436.  I chose to modify the `spawnitem` command to take a clientid in place of `inventory` or `cursor`, as in the issue's example, because this seemed to be the simplest way implement this without creating multiple redundant commands or complexities when separating player names from item names.  It is possible that this could cause problems if a player attempts to spawn an item with a name that ends in an integer in a random location, but since there are currently no items with names ending in an integer, this seems like a niche enough case to dismiss, and could be circumvented by using `cursor` as the last argument.    If you would recommend another implementation, to this issue, I would be happy to change my solution to meet it.  This just seemed like a good path to start with.

Closes #436

### DIFF
--- a/Barotrauma/BarotraumaShared/Source/DebugConsole.cs
+++ b/Barotrauma/BarotraumaShared/Source/DebugConsole.cs
@@ -2140,8 +2140,23 @@ namespace Barotrauma
                     spawnInventory = Character.Controlled == null ? null : Character.Controlled.Inventory;
                     break;
                 default:
-                    extraParams = 0;
+                    int id = 0;
+                    int.TryParse(args.Last(), out id);
+                    if(id > 0) {
+                        var client = GameMain.Server.ConnectedClients.Find(c => c.ID == id);
+                        if (client != null && client.Character != null) {
+                            spawnInventory = client.Character.Inventory;
+                        }else{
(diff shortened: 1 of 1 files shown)


Tazial:
Pull request being recreated as #441 to the dev branch instead of master

Tazial:
Pull request being recreated as #441 to the dev branch instead of master

PR #459, opened 2018-07-03
Fixes Release build

JJRcop:
A few things were looking for GameMain.Config.UseSteam regardless of DEBUG, and if DEBUG is not defined, then UseSteam isn't either.

This defines UseSteam if DEBUG isn't set, and sets it to true.


JJRcop:
@Regalis11 Now that I think about it, should GameMain.Config.UseSteam be bound to Steam.SteamManager.USE_STEAM when DEBUG isn't defined? I do understand that Steam.SteamManager.USE_STEAM is bound in reverse to GameMain.Config.UseSteam when DEBUG is defined, what do you think?

PR #501, opened 2018-07-19
Revolver, Plasma Cutter, Oxygen Generator, Oxygenite Tank changed

BlueTheKing:
+ Oxygen Generator - If on fire its condition will deteriorate and it will eventually break, takes construction skill to fix + it can have oxygenite tanks contained in it, but they won't refill
+ Plasma Cutter - If skill isn't high enough oxygenite tank depletes 3 times as fast (1.5)
+ Revolver - New sprite, can now hold 6 bullets, the captain will now spawn with 6 bullets
+ Oxygenite Tank - When on fire they need to have condition greater than 20% to explode, are a bit more dangerous than oxygen tanks


NilanthAnimosus:
Just a note but the oxygen generator can expload due to EMP damage with these changes (If this isn't wanted, the .

I don't think a revolver should start with 6 rounds. I am totally for its capacity being increased but it was a very strong weapon with even 2 rounds.

The revolver round deals 40 damage with 3 bleed icons, and 100 structure damage. Collectively thats 240 damage, 18 bleed icons or 600 structure damage. This can destroy 6 windows, a wall or two - even heavy ones, a mantis (200HP), or just several crawlers and human characters. I just think this is too devastating for multiplayer campaign (As is) to be given for free each round, I think either 3 or 2 rounds is the ideal count to begin with. submarines may add additional rounds to their design or when shopping.

Perhaps (Much like the old barotrauma behaviour from older versions) Oxygenite tanks (Or at least shards) placed into a plasma cutter should have a chance to simply expload on insufficient skill (But not oxygen tanks), instead of depleting them like a normal oxygen tank. they are advanced items that would more likely carry a penalty than regular tanks at least.

Issue #506, opened 2018-07-20
Change campaign purchase menu to include quantity field, add - + buttons and auto stack on additions


NilanthAnimosus:
A simple idea, essentially instead of listing 20 bandages as individual items, add a count and/or sort the list alphabetically on adding items. clients would need less data sent to them (Only the prefab + count), the menu would feel far less cluttered while much improving the ease of use of the menu to work out exactly what you are buying.

If you add 5 bandages, then add 3 bandages later after buying other items, one could forget and wind up buying 8 bandages given  the levels of scrolling sometimes required.

As for adding + - buttons, this should make is easy to either add or remove additional counts of the item, instead of just only removing them.

NilanthAnimosus:
im seeing if I can work on this at the moment as a note. figured I'd give it a try.

PR #507, opened 2018-07-20
Redesign campaign buymenu to use a quantity field

NilanthAnimosus:
Maybe it needs some refactoring/renaming or generally looking neater/more optimized but I implemented a new buymenu for the singleplayer and multiplayer campaign, I however did have to edit the methods using spawncargo.

This should work with multiplayer and single, and write less data as it writes 2 ints / item type.

The ability to click an item to sell it was commented out (Selecting a child element removed the item).

the buymenu list sorts itself by name then category to mimic the buy category (Easier to find/use).

Additionally added a fix for #509


NilanthAnimosus:
For issue #506 
Image of the changes:

PR #523, opened 2018-07-24
Game fails to start on store over buy

ursinewalrus:
I don't know why it says there are so many changes, only real changes start inside the inner for loop. I just re add a container of the last needed type should one not exist in the list availibleContainers due to it being removed earlier for being full. I also add an additional check to make sure when you try to remove a itemContainer from the list of availibleContainers, it actually exists in there currently. Not having that check is what caused the error in the first place.


Regalis11:
Merged. I left out the last commit and added the comment manually, because all the line endings had been changed in the last commit, making it difficult to see which lines have actually changed.

PR #545, opened 2018-07-30
GameMain.Config.UseSteam determined at compile time on release builds

JJRcop:
Based off of Steam.SteamManager.USE_STEAM, similar to how it uses GameMain.Config.UseSteam on debug builds.


rico6822:
Does it prevent barotrauma from using steam's version of the game at all time, when debug is disabled?

Regalis11:
@rico6822 Yes, if you're using a Steam version (= a version bought from Steam) you cannot change it to a non-Steam version. Although unlike most Steam games, you can still play the game if Steam authentication fails (e.g. if you're not logged in to Steam or playing a pirated version), all Steam functionality is just disabled.

PR #703, opened 2018-08-18
Fixed the VOIP not being reset when the item isn't drawn in Mario's hands.

CommanderMark:


juanjp600:
What am I going to do with all this audio? After all, I do build it up for twelve merges.

Issue #832, opened 2018-09-29
Make a message appear on your screen if you get killed by a traitor who had you as his target


BillCombs:
Since griefers exist, sometimes it may be quite hard to understand if the person, who killed you, did it because he had to, or just because he wanted to. Sometimes players were getting really angry at me for killing them, even though i was just doing my job as a traitor. And most of the time they try to votekick me, which leads to other players seeing the votekick message, getting suspicious, finding the body and immediately realizing that i was the one who killed the person. That kinda ruins the fun of being a stealthy killer. I think there should be a message telling players that they were killed by a traitor who had them as his targets, so they would know that it was not just a griefer killing them for no reason.

Perhaps the way it could work is checking if the traitor did any damage or stun their targets recently, and, if the target died soon enough after that, count it as getting killed by the traitor. That won't help those smart traitors, who manage to convince their crew that their target is the traitor and should be killed, but at least it would help those who do killing the usual way; i think most traitors kill their targets themselves, so that would help the majority of them.

ZealanL:
This is a good idea, but when a player re-spawns they would know that person was a traitor.

That's the only visible flaw - besides that, this works.

BillCombs:
The problem is that if you kill a guy for to noticeable reason, you are usually either a traitor or a griefer. So if the player, who got killed by you, did not know if you were a traitor, he would most likely want to revenge kill you anyway. It's not like anybody, who got killed for no noticeable reason, would just immediately forget about it.

The message should probably contain some words about new life rule(like it's made in some roleplay kind of games), since once you are dead, it still counts as traitor's victory even if you respawned. Since there is nothing else for the traitor to do once he completed the task, besides helping the crew, the message should probably tell the player that they should just try to work with the traitor like if nothing ever happened. Unless the crew witnessed the murder and is trying to kill or arrest the traitor. In that case, even if you have a new life, you could still be informed about a traitor on the sub and get an order to kill or arrest him.

The whole thing is mostly needed to help players to distinguish griefers from traitors, so traitors won't get as much hate just for doing their job. If players are dishonest, they will try to revenge kill their killer in any case.

Regalis11:
[...]

PR #951, opened 2018-11-29
Vote kicking is now not allowed during a round

Kampfkarren:
Vote kicking is a flawed system in a game of deceit like Barotrauma. 

Assume this scenario in a public server:

1. The server is traitors maybe/yes.
2. You are a traitor and you have a target to kill
3. You kill your target.
4. Your target assumes you're a griefer and makes a vote kick on you.
5. The entire server gangs up against you because they think you're a griefer and use this meta information to stay far away from you.

A naive solution would be to not show vote kicks in chat, but this would be a bad idea because if there WAS a griefer then some players might not vote kick them.

An ever naiver solution would be to not vote kick traitors, but this is a really bad idea for hopefully obvious meta reasons.

Instead, this PR makes it so you can only vote kick while a round is in the waiting stage. This means that players will have the opportunity to see the traitors end game panel to say "Oh, that guy wasn't griefing, he was just a traitor".

### DIFF
--- a/Barotrauma/BarotraumaShared/Source/Networking/Voting.cs
+++ b/Barotrauma/BarotraumaShared/Source/Networking/Voting.cs
@@ -8,9 +8,17 @@ namespace Barotrauma
 {
     partial class Voting
     {
-        private bool allowSubVoting, allowModeVoting;
+        private bool allowSubVoting, allowModeVoting, allowVoteKick;
 
-        public bool AllowVoteKick = true;
+        public bool AllowVoteKick {
+            get {
+                return this.allowVoteKick && !GameMain.NetworkMember.GameStarted;
+            }
+
+            set {
+                this.allowVoteKick = value;
+            }
+        }
 
         public bool AllowEndVoting = true;
 


Regalis11:
I'm not convinced that this is the best way to solve the problem. It would mean that griefers cannot be votekicked during the rounds, and if there's no moderator/admin around to do the kicking, a griefer can ruin the entire round.

Personally I feel #832 would be a better way to solve this issue. The only downside is having to revealing the traitor to the target, but IMO it's better that the respawning victim knows the killer was a traitor and not a griefer.

Kampfkarren:
> I'm not convinced that this is the best way to solve the problem. It would mean that griefers cannot be votekicked during the rounds, and if there's no moderator/admin around to do the kicking, a griefer can ruin the entire round.

From my experience, I've seen people quick to vote to end rounds when there's a griefer because they're all dead and know they probably can't fix whatever terrible state the submarine is in.

> IMO it's better that the respawning victim knows the killer was a traitor and not a griefer.

Players will meta grudge either way, and power gamers will be told who the traitors are so they can kill them. I'm not convinced that's a better solution.

chesse20:
Yeah this happens a lot, i've seen plenty of traitors get kicked because they tried to kill their target and then got a kick against them

Kampfkarren:
Good point by chesse: that solution only works if the traitor actually manages to successfully *kill* their target. If they don't, then they're seen as a griefer and vote kicked.

thegrb93:
Imo the only solution is detecting grief automatically, e.g. karma, and have the griefer auto commit suicide if their karma is too low. Then this PR would solve the rest of the issues.

It's impossible for users to differentiate between grief and traitor so it has to be left up to a neutral party, i.e. the machine.

Kampfkarren:
> It's impossible for users to differentiate between grief and traitor so it has to be left up to a neutral party

At the end of the round there is: it explicitly says it. And whatever system there'll be will not only unlikely to be perfect (I've seen some clever trolls), but players won't trust the system. If they see a traitor but think they're just griefing, they're more likely to assume that the system doesn't work than that they're a traitor.

Regalis11:
> From my experience, I've seen people quick to vote to end rounds when there's a griefer because they're all dead and know they probably can't fix whatever terrible state the submarine is in.
True, I've seen this happen a lot. And I believe we need to give the players better means to deal with griefers to prevent this kind of stuff from happening - removing the ability to votekick during rounds sounds like doing the exact opposite.

> Good point by chesse: that solution only works if the traitor actually manages to successfully kill their target. 
That's a very good point though.

PR #995, opened 2019-01-17
New Submarine for Consideration - Leatherback

crackbone:
Included is a new submarine for inclusion in the game for developer's build current on January 16, 2019, the Leatherback.

[Leatherback.zip](https://github.com/Regalis11/Barotrauma/files/2767014/Leatherback.zip)


Regalis11:
Something seems to have gone wrong when making the pull request - 250 commits and it seems you haven't committed the new submarine. But anyway, I was able to get it from that download link. Looks quite nice, we'll give it a try and see if it'd be usable as a vanilla sub! 👍

Issue #392, opened 2018-04-26
Logic and Math Wiring Operators


Umbrias:
**Idea Request**
Could be pretty low priority, but in general, subs would be highly optimized and more accessible if wiring logic was introduced. A variety of very basic signals modules, generally all single usage but very basic math and logic operators.
Ideas, because why not.

- [x] Adder
Adds values inputted together, negative or positive, outputs sum.

- [ ] Sign Flip
Flips the sign of a value inputted, outputs value.

- [x] Multiplier
Outputs product of inputs.

- [ ] Inverter
Outputs inverse of input.

- [ ] Power
Outputs say, input 1 to the power of input 2

- [x] A > B
outputs boolean
- [ ] A >= B
outputs boolean
- [x] A = B
outputs boolean
A and B both inputs

- [ ] Constant
Outputs a constant, can be modified by a set.

- [ ] Function
Takes one or more inputs and runs it through a function, i.e. f(x). Nothing could ever go wrong implementing this.

It certainly isn't a high priority feature, however, it would be incredibly helpful for sub design and for sub optimization, and nobody seems to have officially suggested it yet, so this is really to just put it on the radar.

Regalis11:
We're moving feature requests from the issue tracker to the Barotrauma forums. From now on only actual issues (bugs, crashes, design flaws, oversights) should be reported here. We're doing this because the issue tracker has started to become extremely cluttered - it's getting harder for us to find the more critical issues when they're mixed in with hundreds of ideas and feature requests. And even though we are closing the existing feature requests, we are still planning on implementing many of them further down the line. 

So from now on, please post your feature requests on the [Collaboration section on the Barotrauma forums](https://undertowgames.com/forum/viewforum.php?f=19&sid=3ef7ed701ca460dc3503e2ab4e3cf7c9).

PR #1000, opened 2019-01-22
Expansion of Barotrauma's logic system.

kuraiookami:
This is the first time I have tried to contribute to a Github project.
If I have done something wrong with git please explain my mistake so I can learn.

Changed:
- AdderComponent and children can clamp their output
- Powercontainer signals for charge,charge% and charge rate

Added:
- ColorComponent: Dynamic signals for light set_color inputs
- MemoryComponent: Stores and sends a signal that is edge latched
- DivideComponent: Standard division
- MultiplyComponent: Standard multiplication
- SubtractComponent: Standard subtraction
- XorComponent: Exclusive or
- EqualsComponent: Equals comparison
- GreaterComponent: Greater than comparison

### DIFF
--- a/Barotrauma/BarotraumaShared/BarotraumaShared.projitems
+++ b/Barotrauma/BarotraumaShared/BarotraumaShared.projitems
@@ -1470,7 +1470,15 @@
     <Compile Include="$(MSBuildThisFileDirectory)Source\Characters\DamageModifier.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)Source\Events\Missions\MissionPrefab.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)Source\GameAnalyticsManager.cs" />
+    <Compile Include="$(MSBuildThisFileDirectory)Source\Items\Components\Signal\ColorComponent.cs" />
+    <Compile Include="$(MSBuildThisFileDirectory)Source\Items\Components\Signal\EqualsComponent.cs" />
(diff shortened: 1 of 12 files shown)


Regalis11:
Nice work, these would be useful for sure! I didn't test these yet but I believe there's at least one small issue: I think you should be using invariant culture when parsing floats or converting them to strings, otherwise the same circuit may behave differently depending on which decimal separator the player's system is using.

I'll take a closer look at this and merge it as soon as I have my hands a little less full with getting the dev branch ready for closed beta. :)

kuraiookami:
I tested this in-game in every way I could think of but I missed that.
This fault is my own doing, every time I have used GitHub code in the past it has been for personal modifications or learning. I have never really thought about the consequences of things like decimal separators that are foreign to me.

I am unsure of the proper way to test that. Does changing the Windows region->format settings count as sufficient? Or are there other things that need to be done to test this in a complete manner?

Regalis11:
Yeah, that's something very easy to miss (I've done the same mistake several times even after realizing it can make a difference in this game). :)

I believe this could be tested by changing your region to something where a decimal comma is used (see https://en.wikipedia.org/wiki/Decimal_separator). The problem is pretty easy to fix, you just need to use CultureInfo.InvariantCulture when parsing the float signals and when converting the result of the calculation back to a string.

CommanderMark:
Seems like this addresses a lot of #392.

rico6822:
Can we turn "to do" stuff in here into task list so we can know what remains undone?

Regalis11:
Added in 26e89c6

PR #1084, opened 2019-02-06
Splash screen rewrite

juanjp600:
- [x] Test on Windows
  - [ ] Move libvlc to a folder of its own alongside the plugins?
- [x] Test on MacOS
- [ ] Update install script (libvlc5)
- [x] Fix audio callbacks? ~~Might need to modify LibVLC itself or write a callback in C~~ Had to do with delegates being GC'd
- ~~Consider just rewriting this whole darn thing using ffmpeg instead~~ libvlc seems to work well enough now


juanjp600:
Seems to be working perfectly on all platforms.

Regalis11:
I'm guessing the plugins folder should be copied to output directory as well? Now this causes the game to crash in the Video constructor.

Regalis11:
It also seems something went wrong with the merge in b4242e0. The molochbaby files are missing from BarotraumaShared.projitems, and git shows all lines as being changed (maybe something to do with line endings?) so it's hard to see what exactly has been changed.

Regalis11:
Force-pushed because I accidentally pushed d0ec94d into this branch

juanjp600:
Whoops, you're right, the plugins folder should be copied.

Also I really have no idea what's causing every line to change on my end, the most I really did was add the Video and VideoSound classes.

Issue #1110, opened 2019-02-10
Add a hotkey for the info tab (plus some general reworking of the defaults to include this new key).


CommanderMark:
Hotkey that when held down brings up the info tab, so it isn't a toggle.

This would probably make the most sense being bound to the tab key, since most games use that for bringing up player info. The default chat key could then be moved over to T, the radio moved over to Y, and then just have deselecting chat bound to ESC.

Issue #1108, opened 2019-02-10
Make the fact that the chat window has been selected more obvious.


CommanderMark:
I tend to wind up in a lot of situations where my character abruptly stops responding to my keyboard input, only to realize that I somehow enabled typing in chat while performing some other action.

Regardless of why I keep accidentally hitting the key, there should be a visual and/or audio cue that occurs when the chat box is opened. Maybe the chat window should flash a bright yellow?

rico6822:
The entire lower screen would shine in green by a short range from it until chat is left.

Issue #1017, opened 2019-02-01
[dev] Game starts in windowed mode when fullscreen is enabled


Deelite34:
- [x] I have searched the issue tracker to check if the issue has already been reported.

**Description**
I've launched the game, with full screen setting enabled. There was window bar on top, as if game was in windowed mode. Minimizing the game then maximizing fixed the issue. 
Same thing sometimes happens, when changing screen mode from any to any setting (fullscreen, fullscreen borderless, windowed)

**Steps To Reproduce**
If possible, describe how the developers can reproduce the issue. It is often extremely hard to fix a bug if we don't know how to make it happen.

**Version**
Which version of the game did the bug happen in. Also, please include the operating system you're using (Windows/Linux/Mac).
Windows 10 steam closed beta

**Additional information**
screenshot of issue: https://i.postimg.cc/k4Yx4WVG/Fullscreen.png

Issue #915, opened 2018-11-16
Can't bind ingame controls to mousewheel key and additional mouse keys(mouse4, mouse5, etc.)


BillCombs:
Pressing those mouse keys seems to do nothing when trying to change the controls.

You can work around that by making mouse keys work as certain keyboard keys and then binding the controls to those keys, but it would be more convenient if you could bind the mouse keys right away.

PR #1132, opened 2019-02-12
[WIP] Misc. hotkey stuff.

CommanderMark:
Replaced the info button in a game session with a keybind, alongside reworking some of the default chat bindings. This brings the key layout closer to what most other multiplayer games do with chat bindings.

The defaults are:
- Tab for bringing up the info frame.
- T for entering chat.
- Y for entering radio chat.

Note that the keybind for selecting chat is no longer a toggle, instead you have to press escape or click outside the window to deselect it.

To-do:
- [x] Address #818.
    - [ ] Test on Linux.
    - [x] Test on macOS.
- [x] Address #1017.
- [x] Address #1108.
- [x] Address #1110.
- [x] Address #915
    - [x] Make scroll wheel bindable.
    - [x] Make Mouse4 and Mouse5 bindable.


CommanderMark:
[So apparently support for Mouse4 and Mouse5 was only added to MonoGame in 3.7](https://github.com/MonoGame/MonoGame/pull/6418/commits/9f257694b916fa841a025f4595752a66506a0fd9). Either gonna try updating to that version or just call it quits for now.

Issue #1411, opened 2019-04-08
[Mac] [Bug] Game still doesn't launch from Steam, probably due to launch script error or mono missing


hubismal:
Probably a good one to fix before the Mac testers arrive: The mac version of the game doesn't run when double clicking play from steam. I think it has something to do with the script using mono64 or something (I got an error when trying to run it on mac). This was tested yesterday so unless there was a patch for mac in that short amount of time this should be relevant.
This might have to do with what's installed on the base machine (I don't think mono is manually installed on my Mac), but either way a new user won't be able to run it.

hubismal:
I should also mention my mac is like a 2012-13(?) Mac (I don't buy all the latest apple products anymore). It runs every single other steam game fine, but it would be helpful if the two other Mac Gamers remaining on this section of the solar system verified the non-functionality

CommanderMark:
> but it would be helpful if the two other Mac Gamers remaining on this section of the solar system verified the non-functionality
You rang?

So yeah the game doesn't launch directly inside of Steam on macOS. Navigating over to the game's files and running the launch script manually does seem to work though.

Regalis11:
Thanks for the notice! I think launch settings were wrong on Steamwork's side, could you try if it works now?
![asefrew](https://user-images.githubusercontent.com/14351070/55757815-ae0c2280-5a5d-11e9-8479-31f75c190e58.png)

hubismal:
>could you try if it works now?

Was I supposed to recieve some kind of update? Nothing is different for me.
Trying to run it outside of steam prints:

>Support/Steam/steamapps/common/Barotrauma/barotrauma_launch.command: line 3: mono64: command not found

then logout... etc etc, [Process completed]

CommanderMark:
It still doesn't work for me, but in addition to that my build is still on v0.8.9.6. Is the Mac repo up to date?

hubismal:
My build is still way behind too, it's on 0.8.9.7. I just think virtually nobody has been pushing updated Mac builds for steam.

CommanderMark is able to run his because he has mono64 installed. I don't. 
I don't think many players are going to come in with mono64 automatically installed (especially the casual spectrum of Mac players), so if there's a way to automatically install mono, that should probably be configured (assuming it's straightforward).

I'm also going to uninstall mono on linux and see if the game still runs.

Regalis11:
[...]

PR #1426, opened 2019-04-19
MonoKickstart on macOS.

CommanderMark:
Addresses #1411. Extends the portable Mono setup on Linux to Mac.

It's almost working, but for some reason SharpFont refuses to detect the freetype dylib on computers without Mono installed. Once that's figured out, it should be good to go.


CommanderMark:
So for whatever reason, it seems as though freetype was bundled inside of `libvlc.dylib`, and that appears to be the only thing that a no-Mono-installed Mac will accept in order to run the game.

Regardless of whether or not it's running on magic at this point, it appears to be working properly. One other thing to note is that I had to test these changes on the master branch since I don't have the dev branch content folder. Doesn't seem like it matters at all but I was blindly editing csprojs in the final half of this PR so...fingers crossed I didn't break anything.

Regalis11:
Cherry-picked in 1e02bf9..25526ab

PR #1502, opened 2019-05-23
Moved "ReadWrite" files to save folder.

CommanderMark:
If the macOS version's resources are going to be inside of an app bundle, then it can no longer be guaranteed that files (or paths rather) relative to the program are read-writable, they have to be read-only. So I moved the config files and crash reports to the save folder (Library/whatever on Mac, appdata/whatever on Windows).

Also did these:
- Launch arguments are passed to GameMain. At the moment it is only used for when someone writes "skipintro" as an argument to disable the splash screen.
- Crash reports are opened automatically when they occur so a user doesn't have to navigate to hidden folders each time the game crashes to find a report.
- The game no longer includes a default config and instead generates it programatically.

It seems like the last "ReadWrite" files that are still relative are the "Mod" "Data" and "Submarine" folders. I decided not to touch those for the time being since I'm not sure what your plans for them are.


CommanderMark:
Just realized that there's also the "Data", "Mods" and "Submarines" folders which are still relative to the exe path. They should be moved to the Save folder.

However, it seems like those need to be bundled with the game itself, so maybe it should copy those files over to the save folder on launch when it doesn't detect them, and then use the save folder directory from now on?

Not sure if you guys would have a different approach. Thoughts?

CommanderMark:
Alright so it seems the Mac port after the Steam release is already inside of a .app bundle that can be written to. As long as code signing never becomes an issue then this PR is no longer necessary.

PR #1589, opened 2019-06-11
AI ladder fixes

0hm:
This patches the current behavior, but the real problem is in heightFromFloor not calculating correctly (instead seeing a different floor further down).

This is trivial to reproduce in the current master and dev build: start on the typhon sub, then tell the AI to man the top left (dorsal) coilgun. Alternatively, walk into the room and tell the AI to follow you. They consistently get stuck on the ladder in the access shaft and refuse to let go. I can provide some screenshots with debugdraw on if it's helpful.

### DIFF
--- a/Barotrauma/BarotraumaShared/Source/Characters/AI/IndoorsSteeringManager.cs
+++ b/Barotrauma/BarotraumaShared/Source/Characters/AI/IndoorsSteeringManager.cs
@@ -277,6 +277,11 @@ namespace Barotrauma
                         }
                         currentPath.SkipToNextNode();
                     }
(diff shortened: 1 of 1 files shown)


itchyOwl:
Thanks for investigating and reporting the issue. The heightFromFloor should now be properly calculated, so there's no need for the workaround (although it did work in this case). The issue was due to a condition (Ragdoll.cs:1443, the case on CollisionPlatform), which made the raycast to ignore the platform (and actually it did ignore all the platforms on top of it, because IgnorePlatforms is enabled when climbing up the ladders.

0hm:
Awesome! Much appreciated, I'll close this out. :)

PR #1612, opened 2019-06-12
Load GameMode setting from serversettings.xml in dedicated server

tjhowse:
Built and validated the fix in https://github.com/tjhowse/barotrauma_server_docker
Saving the setting to the file doesn't work, but nor does saving any other setting I've tested :)


Regalis11:
This has been fixed in the recent versions

PR #1635, opened 2019-06-14
Game automatically opens crash logs and passes command line arguments to GameMain.

CommanderMark:
* When a crash occurs the game now opens the crash log.

* Command line arguments are passed to GameMain. At the moment this is only used to skip the splash screens when a user writes "skipintro" but it can probably be used for more in the future.


Regalis11:
Merged to the public repo

PR #1774, opened 2019-07-29
Crouch button on reactor for fine adjustments

JJRcop:
Lets you hold the crouch button when using movement keys on the reactor to slow it down by 60%. If you hold the sprint button to speed it up, the sprint button takes priority and overrides the crouch.


Regalis11:
Merged in the private repo

PR #2116, opened 2020-02-02
Steering Movement Keys uses Run bind for speed up

JJRcop:
Rather than hard coding the shift key to speed up the input rate, the key bound to run is used instead, which by default is shift


Regalis11:
Thanks, this had completely slipped past us! We merged the fix in the private repo.

PR #2332, opened 2020-02-20
Change AFK Kick Timer to 10 minutes

LucasMolander:
This was a much-needed change. Many times, our crew would have a member go to the bathroom and they're kicked by the time they come back.

Can't we just leave the extreme AFK cases to vote kicking, complemented by this 10 minute hard cap?

### DIFF
--- a/Barotrauma/BarotraumaShared/Source/Networking/ServerSettings.cs
+++ b/Barotrauma/BarotraumaShared/Source/Networking/ServerSettings.cs
@@ -742,7 +742,7 @@ namespace Barotrauma.Networking
             private set;
         }
 
-        [Serialize(120.0f, true)]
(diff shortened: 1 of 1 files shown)


Rokvach:
There's another ticket about afk kicking where we would allow the host to define if the afk kick is enabled and how long it will last

LucasMolander:
@Rokvach Gotcha - would it have a default value, still? If so, this commit might make sense to keep.

Can you link me to that ticket? I couldn't find it

Regalis11:
>Can you link me to that ticket? I couldn't find it
Looks like that ticket is in our private repository.

But in any case, I think increasing the default AFK kick time is still a good idea, so I changed it in the private repo. Thank you for the PR!

PR #2707, opened 2020-04-25
Returned method DrawUserIndexedPrimitives back

BurykinNikolay:
After last update (commit ac37a3b0e4e7b8fadb1bedbdddb386f64fe46592)  (0.9.9.0)
I see a strong drop in performance on different video adapters.
Memory leak occurs after update of SpriteBatcher.cs.
You switched to the DrawIndexedPrimitives method, but this method requires very precise work with dispose of buffer.
As a quick solution, I suggest returning the old method

### DIFF
--- a/Libraries/MonoGame.Framework/Src/MonoGame.Framework/Graphics/SpriteBatcher.cs
+++ b/Libraries/MonoGame.Framework/Src/MonoGame.Framework/Graphics/SpriteBatcher.cs
@@ -14,7 +14,7 @@ namespace Microsoft.Xna.Framework.Graphics
     /// sent to the GPU). 
     /// </summary>
 	internal class SpriteBatcher : IDisposable
-	{
+    {
         /*
          * Note that this class is fundamental to high performance for SpriteBatch games. Please exercise
          * caution when making changes to this class.
@@ -41,7 +41,7 @@ namespace Microsoft.Xna.Framework.Graphics
         /// Index pointer to the next available SpriteBatchItem in _batchItemList.
         /// </summary>
         private int _batchItemCount;
-        
+
         /// <summary>
         /// The target graphics device.
         /// </summary>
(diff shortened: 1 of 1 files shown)


juanjp600:
I don't see how the changes made to SpriteBatcher are causing a memory leak that didn't exist in the previous version; the implementation of DrawIndexedPrimitives does mostly the same work as DrawUserIndexedPrimitives, with the notable exception of the latter handling uploading of vertex and index data to the GPU. The change was made to minimize index data updates, which has yielded a significant performance improvement in our testing.

Additionally, the changes you've made in this pull request make the SpriteBatcher far less efficient than either previously used method, as it retains the new vertex and index buffer that were introduced by the update, but then reuploads all of the vertex and index data in DrawUserIndexedPrimitives, thus uploading the vertex data twice for no good reason.

I've checked both your branch and current master to see how they compare, master seems to perform better on average. The performance hit the latest update has seen is likely related to one of the many other features we've implemented, and is unlikely to be related to the changes made to the SpriteBatcher.

That being said, there are a few things that could be improved to our current solution and I'll be pushing them soon. If you can demonstrate that there is indeed a memory leak or otherwise clearly less efficient aspect of the SpriteBatcher, I would appreciate it if you profiled the difference and demonstrated what is causing the performance hit (`showperf` and `fpscounter` on Release builds is usually good enough if the difference made is large and consistent enough, though running Visual Studio's profiler is preferable as it can further narrow down potential issues)

BurykinNikolay:
@juanjp600  
Good day. Ok, now I will take quick screenshots with performance on different methods:
DrawIndexedPrimitives:
IMAGE 2020-04-26 23:38:58


DrawUserIndexedPrimitives:
IMAGE 2020-04-26 23:39:33


On Monday, I will try to give more information.

juanjp600:
Update: DrawUserIndexedPrimitives does seem to have an optimization that's fairly significant; when using a buffer that's larger than the amount of vertices to draw, DrawUserIndexedPrimitives will use various offsets within the buffer to allow the driver to return immediately without risking overwriting data. This optimization has now been ported into the SpriteBatcher.

juanjp600:
The SpriteBatcher is being temporarily reverted to its previous state until we can be sure that the new implementation is stable.

PR #3131, opened 2020-06-18
Hash function is now case-insensitive. (fixes #3114)

### LINKED ISSUE #3114: [v0.9.10.0] Folder case-sensitivity issues for Workshop and Server/Client hashing.

**Description**
First off, Windows is case-insensitive to the point where tools like the Steam Workshop, and Git (with certain settings) can't detect if the folder name has updated casing, and doesn't update it. This prevents mod creators from updating folder casing since it won't actually propagate to users unless they delete their entire mod folder.

However, Barotrauma is case-sensitive in the Hash function, so even if their mod is working perfectly fine (due to not caring about case-sensitivity), the server will kick them for not having the correct hash.

**Steps To Reproduce**
Change the casing of a folder in a mod on the Steam Workshop, notice it doesn't update the casing for clients!
Barotrauma, and Linux is sensitive to these changes though, so a Linux server downloading the mod will have the correct folder casing, and kick all clients with the wrong casing out.

Any clients with the wrong folder case will get kicked with the error message `Missing content package. You need the content package "x". (hash XXXXXX)`

**Version**
v0.9.10.0, Steam, Windows (Client) and Linux (Dedicated server).

**Additional information**
Had a hard time figuring out why some of my friends could join the server, and others couldn't. Finally pinned it down on one of the mods updating the casing of a folder so that it works on Linux.
I'm sure this problem will happen again since it's pretty common for a mod creator to accidentally use the wrong casing, then updating it to work on Linux after people complain.

Not sure if this should actually be fixed, though it does cause a lot of mysterious issues for mod developers using typical version control systems (Steam workshop, git on windows).

naelstrof:
Tested on my personal Dedicated Linux server and it indeed fixes #3114 .

This does create a different issue where if a mod creator somehow made _specifically_ a creature mod with two identical creature names with different cases it has a 50% chance of failing a hash check. Although Windows won't actually allow this by default (since it would see them as identical folders anyway):

animation of folder conflict message that windows gives

By the way, it was **extraordinarily** easy to compile from source and test! Thanks for the great open-source game, I might mess around more in my free time as I get more familiar with the code-base.

### DIFF
--- a/Barotrauma/BarotraumaShared/SharedSource/ContentPackage.cs
+++ b/Barotrauma/BarotraumaShared/SharedSource/ContentPackage.cs
@@ -525,7 +525,13 @@ namespace Barotrauma
                 {
                     using (MD5 tempMd5 = MD5.Create())
                     {
-                        filePaths = filePaths.OrderBy(f => ToolBox.StringToUInt32Hash(f.CleanUpPathCrossPlatform(true), tempMd5)).ToList();
+                        // TODO: ToLower() fixes discrepencies between windows folder name capitalization for server/client interactions.
+                        // The proper fix would probably be to save both a case-sensitive and non-casesensitive hash on the server,
+                        //     and only compare case-sensitive clients with the case-sensitive hash, and case-insensitive clients with the
+                        //     case insensitive hash.
+                        // Though ultimately, the only case where this will cause issues is if someone made a mod with identical folder names in the same path.
+                        // Windows... 😩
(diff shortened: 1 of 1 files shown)


Regalis11:
Thank you for the pull request! However, this fix doesn't work as intended, because ToLower is culture-sensitive and the results depend on your system language. For instance, "Items" isn't "items" when converted to lower case in Turkish, so this would introduce hash mismatches even when the clients are using the same OS.

naelstrof:
Ah gotcha, in that case I think an alternative solution would be to rely on the xml's literal strings for sorting rather than trying to sort based on the result of `CleanUpPathCrossPlatform`.

Feel free to close this pull request, I'll try again unless you think you think otherwise.

PR #3278, opened 2020-07-03
Add Concatenation Component

Jlobblet:
Text concatenation has been a feature I've seen requested by Discord users multiple times, so I thought I'd write a solution for it.

I took inspiration from the `ArithmeticComponent` class as a powerful base class to use for a wide variety of applications and made the class `StringComponent`. It's effectively just a trimmed down `ArithmeticComponent` (no need to clamp values) and with types changed from `float` to `string`. 

From here, implementing the concat component is as simple as writing a `Calculate` method that just returns the concatenated string, and is handled by `ConcatComponent.cs`.


Regalis11:
Thanks, this will definitely be useful! Merged to our private repo so I'll close the PR

PR #3782, opened 2020-09-09
Fixed a bug in IsValidTarget where targets that dont have their ident…

Mezzokoko:
…ifiers already lower-cased dont get matched


Mezzokoko:
Replaced with new PR with fix on a separate branch

PR #3783, opened 2020-09-09
IsValidTarget bugfix: target tags that weren't lowercased wouldn't match

Mezzokoko:


Regalis11:
Thank you for the PR! And I'm glad you spotted this issue, I hadn't run into it before. However, I don't think ToLowerInvariant should be used here, because it allocates memory and this method is called very frequently.

Mezzokoko:
If performance is an issue, perhaps Identifiers should be turned into a struct, with the original and a lowercased version available for this exact purpose, it would only need to be setup once on initialization.

Mezzokoko:
Or better yet, hashing the lowerinvariant of the identifier so that you can skip the string compare entirely, thats how most modern game engines handle that kinda stuff.

Mezzokoko:
It seems C# checks string references at first when checking for being equal, but otherwise falls back to a conventional string compare, which means there probably is alot of potential for performance improvement by replacing them with simple hash compares. The struct I just committed should take care of that, aswell as potential collisions. I'll try and observe if it's gonna be buggy so far in the coming days.

Regalis11:
Thank you for all the effort you've put into this, but I'm afraid we can't merge this PR. 

- The approach you've taken is pretty risky because the hash returned by String.GetHashCode is not stable, meaning that it can vary between different systems, .NET versions or even subsequent runs of the program.

- I also feel this is overkill for solving the bug in question. I mentioned performance issues with ToLowerInvariant because the method generates garbage, and due to how often the IsValidTarget method is called, it'd have a measurable effect. However, the cost of doing the string comparisons is negligible, and I don't feel it justifies introducing this much additional complexity. It's also risky to do such extensive changes to the identifier system at this point of development: we would need to allocate a ton of QA time to make sure this doesn't break anything.

- This makes identifiers case-sensitive, while we've generally treated them as case insensitive, so this could cause some unintended side-effects or break some mods. The original bug could also be fixed by ignoring the case in the string comparisons using StringComparison.OrdinalIgnoreCase (which is very fast and doesn't require instantiating new string objects using ToLower or ToLowerInvariant).

Regalis11:
Fixed in https://github.com/Regalis11/Barotrauma-development/commit/f3fc0b43106dae40e27eb610dfb5a705736cfe1a

Mezzokoko:
The hashing algorithm can be easily replaced by something else, I just went with .GetHashCode() at first because it was easily available and I assumed it would be good enough. It also doesn't matter if it varies between platforms because it only needs to hash strings on a local instance, the resulting hash table isn't shared or saved between instances or sessions.

As for performance, while I agree that single stringcompares on their own aren't all that necessary to omit (which is why I focussed only on ones that would pop up in the debugger repeatedly at playtime), things like ticking status effects and AI evaluations/inventory queries could be made lighter to run, so I believe it's still worthwhile considering. My StringTag class still allows to compare against raw strings, which means not all of them need to be changed to use StringIdentifiers.

As for the rejection itself, fair enough, I almost expected that it might not make it through, but I was curious to see if I could make it work anyway, and so far I haven't run into any issues yet. Even without the string hashing there is alot of duplicate code related to tag management, I think it makes sense to consolidate it all into one place like in my branch to prevent possible future inconsistencies.

PR #3817, opened 2020-09-12
Fixed AI not updating when not ragdolled, but having a miniscule amount of stun

Mezzokoko:
Bug Recreation:

Have a character (in my case a mudraptor) with an affliction with a constant stun effect of strength=1 (e.g. late stage Cyanide Poisoning), then inject them with hyperzine. They will not be incapacitated, but their stun amount will be non-zero, causing the AI controller update loop not to be called.

devenv_8iXAkq3lVp

https://drive.google.com/file/d/1lqvwlCMRgpu15QKOtWxrdPz5CZ-X_nbz/view?usp=sharing

### DIFF
--- a/Barotrauma/BarotraumaShared/SharedSource/Characters/AICharacter.cs
+++ b/Barotrauma/BarotraumaShared/SharedSource/Characters/AICharacter.cs
@@ -47,7 +47,7 @@ namespace Barotrauma
             base.Update(deltaTime, cam);
 
             if (!Enabled) { return; }
-            if (IsDead || Vitality <= 0.0f || Stun > 0.0f || IsIncapacitated)
+            if (Vitality <= 0.0f || IsCurrentlyRagdolled)
             {
                 //don't enable simple physics on dead/incapacitated characters
                 //the ragdoll controls the movement of incapacitated characters instead of the collider,
--- a/Barotrauma/BarotraumaShared/SharedSource/Characters/Character.cs
+++ b/Barotrauma/BarotraumaShared/SharedSource/Characters/Character.cs
@@ -444,6 +444,11 @@ namespace Barotrauma
             }
         }
 
+        public bool IsCurrentlyRagdolled
+        {
+            get; private set;
+        }
+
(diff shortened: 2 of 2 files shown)


Mezzokoko:
Well it turns out the fix wasn't quite this simple, while the stun related stuff was fixed, paralyzed AI are still ticking, I'll try and figure out how to amend this.

Mezzokoko:
Alright it turns out that the Hyperzine anti-stun effect gets applied entirely before the character updates happen, which sets the stun successfully to zero, but has the rest of the effect essentially discarded, so that when the afflictions update later, they just apply the stun effect from cyanide ontop of zero again, as opossed to ontop of the remaining negative stun. Why this still doesn't cause the Character to ragdoll I haven't figured out yet.

Regalis11:
Thank you for the PR, but I don't think this is a good way to solve the issue.
1. The value of this new "IsCurrentlyRagdolled" is not necessarily correct in all situations: for instance, if Update is not being called due to the Character being disabled, or if the character has been killed/incapacitated/stunned in between the Update calls. While currently it doesn't matter because it's only used in that one check in AICharacter, it'd be very easy to accidentally use it incorrectly especially now that the inconsistent behavior of the property isn't explained anywhere.
2. I think the property adds a lot of confusion: in my opinion it doesn't make sense that there are 2 properties, IsCurrentlyRagdolled and IsRagdolled, that don't mean the same thing.

Mezzokoko:
I agree on point 2, but from what I gathered it seems IsRagdolled describes a player's intent to ragdoll via the space bar, not the actual state of being ragdolled, which initially confused me aswell.

Regalis11:
As far as I can tell, the way IsCurrentlyRagdolled is implemented doesn't describe the actual state of being ragdolled either; it's true if the character was dead, incapacitated or stunned (but not manually ragdolled) the last time it was updated, even though it may no longer be.

Mezzokoko:
True, but the property is only used to decide whether or not an AI Character loop should play, in which case manual ragdolling shouldn't be relevant.

Issue #3904, opened 2020-09-25
Suggestion: Water Volume Detection


Eskandaer:
Currently water detectors' signal_out is an output with a binary option of strings from whether the hitbox is in water or not.

Connecting this signal, however, to a status monitor provides the status monitor with percentage of flooding in containing hull.

It would be exceptionally useful for more creative circuits/logic functions if the water detectors had an additional signal node output for percentage of water volume in hull.  This would provide sub creators the ability to determine flood volume and introduce this variable into their custom circuits.

Regalis11:
Implemented in https://github.com/Regalis11/Barotrauma-development/commit/d5e7f3e869dc375b28ae2826c8462197692fa98a

Jlobblet:
Tested, working correctly.

Rokvach:
Closing

PR #3916, opened 2020-09-26
Add "water_%" output signal to water detector

rasicz:
#3904


Regalis11:
This causes a crash if the detector is outside a hull

rasicz:
Added condition to test if hull is not null.

Regalis11:
Some additional issues:
- The output is inconsistent, because the conversion to string is culture-sensitive (e.g. the decimal separator can be a dot or a comma depending on the user's language).
- The format of the output can vary wildly: e.g. "4", "1.45E-6f", "8.34561"... I think the output would be much easier to use if it was an integer.

rasicz:
This shoud hopefully fix all issues.

Regalis11:
Thanks! Couldn't spot any additional problems, merged this to our private repo now.

PR #4000, opened 2020-10-04
Extract player details dialog out of NetLobbyScreen.

someone972:
Multiple locations were calling the SelectPlayer function of NetLobbyScreen.
This function created a details dialog for the given player client. Since
this is used outside of NetLobbyScreen and has no specific ties to it, it
should be a separate class. Created a PopUpDialog class which tracks active
dialogs for activation purposes, and is intended to act as a root for dialogs
that block out the background. The SelectPlayer function has been refactored
into PlayerDetailsDialog with a focus on making the code more readable by
splitting it into functions handling the different sections.

Expected behavioral changes:
- Previously clicking the background behind the dialog did not close the
  dialog due to elements being out of order, this has been fixed so that it
  closes the dialog.
- When checking/unchecking a permission, if it matches a preset then the rank
  dropdown will be set to that preset.


Regalis11:
Finally got around to reviewing this PR. Thank you for all the effort you've put into this!

However, unfortunately we decided not to merge this. Even though I agree that NetLobbyScreen is not the most appropriate place for the SelectPlayer method, especially now when it's being called from so many places, I don't think it warrants this extensive refactoring (we would need to allocate a lot of testing and QA time to merge this).

Another issue with the PR is that PopUpDialog is essentially a duplicate implementation of GUIMessageBox: they mostly serve the same purpose.

someone972:
No worries. I did find later that GUIMessageBox was an alternative but didn't bother updating the PR. Maybe you guys can tackle it sometime in the future.

PR #4001, opened 2020-10-04
Use PlayerName instead of server name for player when hosting server.

### LINKED ISSUE #3868: P2P Servers - Use defined Player Name in Server Browser as the Host's Character Name instead of Server Name for the Character when they host

**Description**
When hosting P2P Servers, hosts will have their character name as the server name, which is quite annoying, I think it'd be nice if it uses the defined Player Name in Server Browser as their character by default instead.

**Version**
~0.10.5

someone972:
Change to use the player name from Config when hosting a P2P server. Previously
it was setting the player's name to be the server's name instead of what the
user configured (or the steam name if none was set previously).

Fixes #3868


Regalis11:
Thank you for the PR! We merged this to our private repo, closing this one

PR #4078, opened 2020-10-14
Fix water flow sound disappearing over time due to bad position.

someone972:
The issue was introduced in the latest unstable. A change was made so
that the targetFlow arrays were not allocated on every update, but
the code was relying on the new allocation to have 0.0f values as
starting state. This fix resets the values to 0.0f where it used to
be allocating the arrays.


Regalis11:
Thank you for spotting this! Merged to our private repo.

PR #4079, opened 2020-10-14
Fix disposed sounds constantly getting dispose called again.

someone972:
I noticed when debugging the water flow sound issue that when
returning to the menu, sounds which were disposed already were still
getting dispose called on every update loop. This fix sets the
playingChannels reference to null after disposing of the sound so
that it no longer wastes cycles on disposed sounds.

It looks like when a SoundChannel is disposed, it can never start
playing again, and if any exceptions occur in SoundChannel.Dispose a
crash occurs, so no point in calling Dispose more than once.


Regalis11:

PR #4115, opened 2020-10-21
Update ArithmeticComponent.cs

KalinaWildfire:
Fixed signal_in_2 never timing out if signal_in_1 timed out. 
In-game, the issue translates as being able to output if sending a signal to signal_in_2 and then signal_in_1 even though the signals weren't received at the same time.

### DIFF
--- a/Barotrauma/BarotraumaShared/SharedSource/Items/Components/Signal/ArithmeticComponent.cs
+++ b/Barotrauma/BarotraumaShared/SharedSource/Items/Components/Signal/ArithmeticComponent.cs
@@ -52,15 +52,12 @@ namespace Barotrauma.Items.Components
 
         sealed public override void Update(float deltaTime, Camera cam)
         {
-            for (int i = 0; i < timeSinceReceived.Length; i++)
+            for (int i = 0; i < timeSinceReceived.Length; i++) { timeSinceReceived[i] += deltaTime; }
+            if (timeSinceReceived[0] > timeFrame && timeSinceReceived[1] > timeFrame) 
             {
-                if (timeSinceReceived[i] > timeFrame) 
-                {
-                    IsActive = false;
-                }
-                timeSinceReceived[i] += deltaTime;
+                IsActive = false;
+                return;
             }
-            if (!IsActive) { return; }
             float output = Calculate(receivedSignal[0], receivedSignal[1]);
             if (MathUtils.IsValid(output))
             {
--- a/Barotrauma/BarotraumaShared/SharedSource/Items/Components/Signal/ArithmeticComponent.cs
+++ b/Barotrauma/BarotraumaShared/SharedSource/Items/Components/Signal/ArithmeticComponent.cs
@@ -53,10 +53,7 @@ namespace Barotrauma.Items.Components
         sealed public override void Update(float deltaTime, Camera cam)
         {
             for (int i = 0; i < timeSinceReceived.Length; i++) { timeSinceReceived[i] += deltaTime; }
-            if (timeSinceReceived[0] > timeFrame && timeSinceReceived[1] > timeFrame) 
-            {
-                IsActive = false;
-            }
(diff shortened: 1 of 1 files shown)


Barotrauma/BarotraumaShared/SharedSource/Items/Components/Signal/ArithmeticComponent.cs
```diff
@@ -52,15 +52,9 @@ public ArithmeticComponent(Item item, XElement element)
...
-                if (timeSinceReceived[i] > timeFrame) 
-                {
-                    IsActive = false;
-                    return;
-                }
-                timeSinceReceived[i] += deltaTime;
-            }
+            for (int i = 0; i < timeSinceReceived.Length; i++) { timeSinceReceived[i] += deltaTime; }
+            if (timeSinceReceived[0] > timeFrame && timeSinceReceived[1] > timeFrame)  { IsActive = false; }
+            if timeSinceReceived[0] > timeFrame || timeSinceReceived[1] > timeFrame) { return; }
```
Jlobblet, line 57: The issue here is that the early return causes the later timeSinceReceived values to stay unupdated.  
To illustrate, let's consider a timeFrame = 1.0f.  
A single signal is sent to both inputs. The component calculates the operation over the next second, with both values in timeSinceReceived increasing on each frame.  
Once 1s has passed, the early return causes only the first timeSinceReceived value to be updated. Now, if a signal is sent to input 2, it's timeSinceReceived will continue to not be increased, because input 1 is still greater than timeFrame and exiting the function.  
Now, if we wait another second for signal 2 to time out and send a signal to input 1, there should be no output. However, because signal 2's timeSinceReceived is still 0, we get an erroneous output.

The easy solution here is to move the return from inside the for loop to just after it.
```suggestion
            for (int i = 0; i < timeSinceReceived.Length; i++)
            {
                if (timeSinceReceived[i] > timeFrame) 
                {
                    IsActive = false;
                }
                timeSinceReceived[i] += deltaTime;
            }

            if (!IsActive)
            {
                return;
            }

```

Jlobblet (changes requested):

Barotrauma/BarotraumaShared/SharedSource/Items/Components/Signal/ArithmeticComponent.cs
KalinaWildfire: I did that earlier (see very first file, I also apologize for the multiple updates after that) but I realized that if we do that, then it works for adder with timeframe = 0.00, but the issue remains for timeframe != 0.00 and if signals aren't sent at the same time.

For example, a timeframe of 3s and 1 signal on each input being = 1 and they aren't sent at the same time. It only takes 1 signal to time-out to stop the component from updating while we still need it to update until every signal is timed-out
(do note that that only we are sending inputs from a button and only once at a different time)

at runtime =  0
1) Input 1 = 1 (time = 0s)
2) Input 2 = timed-out (no input)
3) Adder does not output

at runtime = 1s
1) Input 1 = 1 (time = 1s)
2) Input 2 = 1 (time = 0s)
3) Output  = 2

at runtime = 3s
1) Input 1 = 1 (time = 3s)
2) Input 2 = 1 (time = 2s)
3) Output = 2

at runtime 3.01s
1) Input 1 = timed-out
2) Input 2 = 1 (time = 2.01s)
3) Adder does not output

at runtime >3.01s
1) Input 1 = timed-out
2) Input 2 = 1 (time = 2.01s)
3) Adder does not output

at runtime > 10s
1) Input 1 = 1 (another signal is sent)
2) Input 2 = 1 (time = 2.01s)
3) Adder outputs even if signal_2 should've timed-out

alright so here the problem appears. input 2 is still not timing-out anymore because Update() has been disabled with (IsActive = False). The only time we want to stop the component from Update() is after both signals timed-out hence the && > timeframe on the entire array. we do not want to output if either of the signal timed-out hence the early return with || > timeframe

Regalis11:
Thank you for the PR! Merged to our private repo now.

PR #4124, opened 2020-10-23
Fix NPE when can't find containerPrefab while spawning cargo.

someone972:
Fixes a small oversight which was causing the code to blow up if a
cargocontaineridentifier could not be found.


Regalis11:
Merged to our private repo in https://github.com/Regalis11/Barotrauma-development/commit/d6211bc34bd1399cacaabd315e851bb17b4d1ad5

Issue #3944, opened 2020-09-29
Feature Request: Local Voice Hot Mic


Blu3wolf:
At present, users can select voice chat disabled, voice chat with push to talk, or voice chat with voice activation. In voice activated mode, voice is sent to the radio whenever detected. With the push to talk mode, two push to talk keys are set: one to use the radio, and one to use the local chat (voice heard only nearby the player, rather than using the radio). 

I propose a mode be added which allows the player to have a hot mic (always active) for local chat, and a push to talk to use the radio.

ezjamsen:
This is a good idea, I think we should be able to support it pretty easily.

Regalis11:
Implemented in https://github.com/Regalis11/Barotrauma-development/commit/bb872e9

Rokvach:
Tested the UI side of the implementation, working correctly there. Rest needs to be tested in the next Unstable.

Rokvach:
Tested, working correctly. Closing.

Blu3wolf:
Hi guys, coming back to Barotrauma after a break. Has this feature been removed from the game? There appears to no longer be a way to have voice activated local chat, and push-to-talk radio chat.

PR #4197, opened 2020-11-02
Add a setting to use local voice chat for voice activity

Nexum72:
This change adds a config setting to allow players to use the voice activity mode for local voice chat, while keeping the option to use the radio with the push-to-talk key.

This is part one of the modifications our group uses in our games, and I think it might be a good addition to the base game.
The second part is in a seperate PR.


Nexum72:
Existing Feature Request: #3944

Regalis11:
Thank you for the PR! Merged this to our private repo now.

PR #4198, opened 2020-11-02
Add a setting to toggle directional voice for radio

Nexum72:
This change adds a config setting to allow players enable or disable the directional voice for radio chat.

This is part two of the modifications our group uses in our games, and I think it might be a good addition to the base game.
The first part is in a seperate PR.


JoneKone:
Surely it should be a server side toggle.

Regalis11:
Super late to taking a look at this PR, but I agree with JoneKone, I think this should be a server-side setting.

Issue #4140, opened 2020-10-27
[Feature Request] Concatenation component assets implementation


Working-Joe:
It's been a few months since Jlobblet's code for the string concatenation component has been merged with the repo, but we've yet to receive the item in game. Given that it's probably pretty low prio, I've made the assets and XMLs to do some testing with the component in game myself a while back. I thought I might as well [share them here](https://drive.google.com/drive/folders/1UwPmNhyJwxyPWDHof7xAgb9_2IVpV6f2?usp=sharing) in case it may save some time in the implementation, although I'm not sure this is the right place given that it's not code.

Regalis11:
Thank you, these files made it really quick to add! :) Implemented in https://github.com/Regalis11/Barotrauma-development/commit/44ae999ab4bd42cb3bc703b903b61a40cecde246

Working-Joe:
All seems to work as intended in V0.1100.0.1. Thanks for the addition!

Rokvach:
Closing.

PR #4265, opened 2020-11-09
Add String Partitioning Components

nullmaton:
### Reasoning
Building off of JLobblet's String Component Class from [Pull#3278](https://github.com/Regalis11/Barotrauma/pull/3278), I thought that more string-based operations are needed. One thing that isn't possible in the game, is to take a string input and split it, which really limits the use of the terminal object for ship-based commands. 

An example command from a terminal that a sub builder could implement would be `battery <x>` where `<x>` is some number inbetween 0 - 100, with the purpose of setting a sub's battery recharge rate. With the current electronic components, the sub builder would have to check for every case of `battery 0`,`battery 1`,...,`battery 100`, which would require 100 signal check/regex components. By introducing a method of grabbing a substring from a string, it would lead to there being no need to check each case, as the sub builder could just get the substring `<x>` from the terminal output signal `battery <x>`, then send that as a signal to the battery. 

Fundamentally, it would mean a system which would require 100 repeated components, would (if this was implemented) now be possible with a single component.

### Components Details
The two components I've added are the HEAD and TAIL components, both take inputs *signal_in1* (the text) and *signal_in2* (the divisor), and output *signal_out* (all characters to the left/right of the divisor in the text). Example inputs and their results are below.
##### Head Component
```
"battery 56" -> signal_1in   HEAD   signal_out -> "battery" 
" "          -> signal_2in
```
##### Tail Component
```
"battery 56" -> signal_1in   TAIL   signal_out -> "56" 
" "          -> signal_2in
```

As you can see the head component outputs a substring of everything to the right of the divisor, and tail outputs a substring of everything to the left. 

### Assets
The assets for these components are based on WorkingJoe's Assets from [Issues#4140](https://github.com/Regalis11/Barotrauma/issues/4140), with the sprite and inventory icon following the same font as the Concatenation Component (ModeNine), and design similar. The assets are inside the google drive [here](https://drive.google.com/drive/folders/1Ek4G-iz30iM6MfEmhS9Y69Tcx2zEB1j8?usp=sharing).

### DIFF
--- a/Barotrauma/BarotraumaShared/SharedSource/Items/Components/Signal/HeadComponent.cs
+++ b/Barotrauma/BarotraumaShared/SharedSource/Items/Components/Signal/HeadComponent.cs
@@ -1,0 +1,47 @@
+using System;
+using System.Xml.Linq;
+
+namespace Barotrauma.Items.Components
+{
+    class HeadComponent : StringComponent
+    {
+        /*
+            signal_in1 - Input String
+            signal_in2 - Input Limit/Divider
+            signal_out - Output all characters in String LEFT of the first Limit
+        */
+        
+        public SplitComponent(Item item, XElement element)
+            : base(item, element)
+        {
+        }
+
+        protected override string Calculate(string signal1, string signal2)
+        {
+            int index = 0;
+            for (int iter = 0; iter < signal1.Length; iter++)
+            {
+                if (index >= signal2.Length)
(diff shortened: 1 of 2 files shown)


nullmaton:
I've just been informed about the Regex Capture Group feature thanks to JLobblet, so this pull request may not be necessary.

PR #4372, opened 2020-11-22
Make steamID work with multiplayer campaign on dedicated servers

spaivaras:
This fixes the issue that on a dedicated server all clients have steamID set to 0. Those in multiplayer campaign save files characters are basically bound to IP addresses in "endpoint". For players with non-static IP addresses, it's painful.

The issue it seems is that Initial Lidgren connections have a default steamID of 0. And nothing is setting it to real ones after clients pass steam auth. Furthermore Barotrauma.Networking.NetworkConnection has a setter for steamID in protected mode.

I am not a c# developer, so this is just a concept that works in my tests.
IMHO steamID binding should be by default. Or if IP binding with steamids = 0 is still needed, this could be implemented on a switch


Regalis11:
Thank you for the PR! We actually fixed this in our private repository just recently (https://github.com/Regalis11/Barotrauma-development/commit/5689228981bc9da76db3dd23d7a9e789a858d5ef), and the fix is currently included in the unstable build and coming up in the next update, so I'll close the PR.

PR #4404, opened 2020-11-29
Added max output length to ConcatComponent

Working-Joe:
The concatenation component can cause considerable memory usage and networking issues if it is placed in a feedback loop. Doing this can increase the length of a string each update leading to huge signals being sent around.

The concatenation component is the only component which has this problem as all other components already have an intrinsically limited output length.

To not limit the applications of the concatenation component, the maximum output length can be adjusted in the editor. It can not be adjusted in game to prevent griefing. 

I believe addinig this limit to the concatenation component only is a better solution than limiting all signal lengths. Limiting all signals will break existing circuits and limits the possibilities of contraptions that can be made.


Regalis11:
Nice catch, it didn't occur to me that the concat component could be exploited in this way. Merged to our private repo!

PR #4465, opened 2020-12-10
Change Reactor key controls to work anywhere in the UI

Coryf88:
The movement keys can be used to control the Reactor's Fission Rate and Turbine Output only when the cursor on the scrollbars.

This PR changes it so the cursor can be anywhere on the Reactor screen.

### DIFF
--- a/Barotrauma/BarotraumaClient/ClientSource/Items/Components/Machines/Reactor.cs
+++ b/Barotrauma/BarotraumaClient/ClientSource/Items/Components/Machines/Reactor.cs
@@ -53,6 +53,8 @@ namespace Barotrauma.Items.Components
 
         private GUIFrame inventoryContainer;
 
+        private GUILayoutGroup paddedFrame;
+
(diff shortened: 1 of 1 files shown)


mLuby:
A similar thing would be nice for manual control mode in the Navigation Console, since currently wasd only works when te cursor's inside the circular sonar scope.

Rokvach:
I think this would affect being able to run away if the reactor happens to catch on fire, as currently you can still move whilst using the reactor if you don't have your mouse over the sliders, so now you would need to bring the mouse all the way off from the reactor UI in order to be able to run away without first having to close the UI. Same with the sonar, as there are always moments when you need to be able to run without first trying to focus on closing whatever you are viewing.

Coryf88:
It would, however can easily close UIs with right click. Also, you can't actually run away until the UI is closed, since you face towards the UI's object.

Regalis11:
Thank you for the PR! Finally got around to reviewing this, merged to our private repo in https://github.com/Regalis11/Barotrauma-development/commit/b6035b72e6c23a468293cbc0cfd113eb74e9e43a

PR #4531, opened 2020-12-15
Allow option In-Game to have Wifi Component output to chat (with Anti-Greifing Option)

xrabohrok:
### The Scenario
In one of my friend-only game servers, we spent the time and effort to rewire a Humpback to output its charge level over one of the channels of the radio, but nothing happened.  It wasn't until later that I found out that the  "Allow Wifi Chat" is only accessible in the ship editor.  Dang.

### The Proposal
Pretty straight forward: if the server-owner wills it, any player who can get at a wifi component can allow it to output to radio chat.  However, due to its... _Highly abusable_ nature, it the option will be off by default and only reveal itself if the server owner manually turns it on.

### The Changes
Player facing changes are threefold:

1.  Allow the server admin to enable/disable the players being able to set wifi components being able to output over chat in the Anti-griefing tab of the host window.
image

 2. If the setting is enabled (default: disabled), then the setting is shown to any player who physically accesses that component.
image

3. If the setting is disabled (by default), then the setting does not display at all to any player.  
image

### Technical Notes 
Due to the addition to the griefing tab, a text entry needs to be added to the localization files under the tag `ServerSettingsAllowWifiChat` 

I had to do some vaguely unholy things to make this work, I would love feedback to make this adhere to your code standards better.  Most of the weirdness comes from the fact that I can't pass very interesting things to class tags, as they must be computable at compile time.  

Basically, I added a new tag: `[ConditionallyEditable]` .  It takes in an enum that gives it a hint as to what server property it should consult, and then can be evaluated at runtime with `isEditable()` .  Unfortunately, this strategy requires that I find all calls of `GetProperties<InGameEditable>()` and make it somehow take my new tag into account, so it is a bit more messy than I would have liked.  

Anyways, thanks for the consideration.

### DIFF
--- a/Barotrauma/BarotraumaClient/ClientSource/Items/Item.cs
+++ b/Barotrauma/BarotraumaClient/ClientSource/Items/Item.cs
@@ -696,7 +696,8 @@ namespace Barotrauma
                 if (inGame)
                 {
                     if (!ic.AllowInGameEditing) { continue; }
-                    if (SerializableProperty.GetProperties<InGameEditable>(ic).Count == 0) { continue; }
+                    if (SerializableProperty.GetProperties<InGameEditable>(ic).Count == 0 &&
+                        SerializableProperty.GetProperties<ConditionallyEditable>(ic).Any(p => p.GetAttribute<ConditionallyEditable>().isEditable())) { continue; }
                 }
                 else
                 {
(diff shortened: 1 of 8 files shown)


subarotrauma:
I hope this goes through, I had the same issue a few days back!

xrabohrok:
Hmm, I neglected to test this with the ship editor side of things, I'll need to revisit this.

Regalis11:
Finally got around to merging this. I did spot a couple of bugs while merging this: 
- The check for editable properties in Item.CreateEditingHud was inverted, and prevented the editing hud from showing up if the item _has_ conditionally editable properties.
- The game would crash if it tries to check the conditionally editable property when a server/client isn't running.
- The "link to chat" property wasn't editable when a client/server isn't running (e.g. in the sub editor).

Those should now be fixed in our private repo though. I also did some changes to the formatting and renamed some fields/properties to make them a bit more descriptive.

Thank you for the pull request!

PR #4794, opened 2021-01-16
Adds "Hull toggles" for automagical features (Nav Blue rooms, AI Forbidden)

Saevon:
This (should) add editable toggles for:
 * `wetrooms` for rooms that are expected to flood, such as ballast, airlock, bilge, etc.
 * `forbidden` for rooms that should be avoided by the AI (dangerous), can be applied to killboxes, or other dangerous locations.

This allows any room name to be used without losing these two capabilities (including in other languages potentially), but also allows you to toggle this OFF for certain rooms, e.g. a ballast that might serve as a "passageway". This also adds the keyword `void`, usually used for unusable spaces which can contain flood-damage.

Notes:
 * This might work better as "hull tags"?   But I have no idea how to add that.
 * I haven't built/tested this yet, just feeling out the waters to see if this kind of stuff is mergeable.  If it seems like a good Idea ping me and I can maybe try to figure out how to fully build/test this.

(Also adds "infectable" tag for non-ballast pumps, didn't feel it was worth a separate PR for it, but can be split out if one/either feature is not-considered)

### DIFF
--- a/Barotrauma/BarotraumaClient/ClientSource/Items/Components/Machines/MiniMap.cs
+++ b/Barotrauma/BarotraumaClient/ClientSource/Items/Components/Machines/MiniMap.cs
@@ -210,15 +210,12 @@ namespace Barotrauma.Items.Components
                     GetLinkedHulls(hull, hullData.LinkedHulls);
                     hullDatas.Add(hull, hullData);
                 }
-                
+
                 Color neutralColor = Color.DarkCyan;
-                if (hull.RoomName != null)
+                if (hull.WetRoom)
                 {
-                    if (hull.RoomName.Contains("ballast") || hull.RoomName.Contains("Ballast") ||
-                        hull.RoomName.Contains("airlock") || hull.RoomName.Contains("Airlock"))
-                    {
-                        neutralColor = new Color(9, 80, 159);
-                    }
+                    // Colour the room as "supposed to be wet/flooded"
+                    neutralColor = new Color(9, 80, 159);
                 }
 
                 if (hullData.Distort)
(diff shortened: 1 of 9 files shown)


itchyOwl:
Thanks for a good suggestion!

Some remarks:
- There were some issues in the code that threw compile time errors: an assignment to "wetRoom", although the field was named "isWetRoom"; a typo "IsForbiden", but reference to it as "IsForbidden"; and an attempt to set item.isForbidden although the property doesn't have a setter, which makes it readonly.
- Forcing the IsWetRoom property value only in the RoomName property setter is not enough, because the player can then change the value in the editor. However, this change would be overridden next time the submarine loads. So it kind of works, but gives the wrong feedback that you could change the value when you can't.
- I don't think we should add a check for IsWetRoom in the pump water objective, because that would make the bots to ignore those rooms and pumps. If the pump is automatically driven (like they almost always are), the bots should ignore them anyway.
- I renamed Hull.Forbidden as Hull.AvoidStaying, because the property doesn't (and shouldn't) categorically deny the access to those rooms. It only makes the bots avoid that room while idling.
- There's no need for "forbidden" tag for the items, because they can just use "ignorebyai".
- I left "infectablepump" out, because I think it's better to keep all the pumps infectable by default. We could add a property for it that defaults true. Something to consider @3e849f2e5c ?

So, I fixed and changed the abovementioned issues and rewrote some of the code.

Implemented in 
Regalis11/Barotrauma-development@a060c1d and
Regalis11/Barotrauma-development@6131fd6

Lithosagym:
Tested, the tags of IsWetRoom and AvoidStaying can be toggled and work correctly.
(This probably should've been done a month ago)

Rokvach:
Closing.

PR #4904, opened 2021-01-26
Changes to utilize NLog as the logging backend.

someone972:
### Preamble
This is a change I worked on to help me debug issues. I've found it to be quite useful as an additional tool to the regular debugger, and it has a lot of potential to help with debugging issues from end users without having to have them use a debug build or know how to use a debuggger. I've cleaned up the code to a point where I believe it is "production ready" and so I decided to put up a pull request. Although I believe this has a lot of merits over the current logging system, the Barotrauma team can of course decide it's not necessary. I have put it up as a pull request in case it provides a value-add for the main repo. 

### Original commit message:
This includes changes to start utilizing NLog as the logging backend. Included
is instrumentation to both send existing logging from DebugConsole to the NLog
log, as well as new messages made with NLog to the DebugConsole.

This has a number of advantages, some of which are outlined below:
- Logging available from very early on in the program, can log initialization.
- Much better handling of multi-threaded logging
- Finer control of changing log level on individual components; excellent for
  debugging.
- Much more resilient; don't have to worry about log files not being saved on
  crashes.
- Integrated replacement of full path in exceptions/stacktraces with \<DEV\>. No
  need to explicitly call CleanupStackTrace on every exception.

Notable changes:
- All log settings including size, save names, save location, and log levels
  are controlled by an NLog.config file. The VerboseLogging setting in the
  main config file has been retained for the time being.
- Logs are now put in a 'Logs' folder, with 'Client', 'Server', and
  'ReceivedLogs' sub-folders.
- Server logs always save by default now. The setting to save server logs is
  instead changed to toggle sending the server log messages to clients.
- For server and client logs, logs automatically roll over at a certain size,
  and will be removed if more than a specified number of logs exist. This keeps
  the log folders from becoming too bloated, and is of course fully
  configurable.

Translation requirements:
- "ServerSettingsSaveLogs" has been changed to "ServerSettingSendLogs" and
  needs updated translated text. "Save server logs" should change to "Send
  server logs to clients" or similar.
- Not currently translated, but GameMain::ShowNLogConfigFailedToLoadMessage
  could be a potential translation candidate. ServerSource/Program::Main also
  has the same string. 
  _(Note, on further inspection this message comes before language settings are loaded, so probably not a candidate)_

### DIFF
--- a/Barotrauma/BarotraumaClient/ClientSource/DebugConsole.cs
+++ b/Barotrauma/BarotraumaClient/ClientSource/DebugConsole.cs
@@ -132,16 +132,6 @@ namespace Barotrauma
                 {
                     var newMsg = queuedMessages.Dequeue();
                     AddMessage(newMsg);
-
-                    if (GameSettings.SaveDebugConsoleLogs)
-                    {
-                        unsavedMessages.Add(newMsg);
-                        if (unsavedMessages.Count >= messagesPerFile)
-                        {
-                            SaveLogs();
-                            unsavedMessages.Clear();
-                        }
-                    }
                 }
             }
 
@@ -247,8 +237,6 @@ namespace Barotrauma
                 {
                     AddMessage(newMsg);
                 }
-
-                if (GameSettings.SaveDebugConsoleLogs) unsavedMessages.Add(newMsg);
             }
         }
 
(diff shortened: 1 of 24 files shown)


MapleWheels:
This is "officially" abandoned lol.

PR #4913, opened 2021-01-27
Use a sane step value for respawn duration.

someone972:
The respawn duration slider is annoying in that it gives random times
like 3m 7s as the steps. This changes the step for the respawn
duration to be 10 seconds per step. Would have liked for it to be
something more like 15 or 30, but with the way the slider is
programmed that leads to off numbers like 9m 55 seconds since the
minimum is 10 s. Tested with the smallest horizontal resolution and
it's still able to select at 10s intervals fine.


Regalis11:
Thank you for the PR! Merged to our private repo in https://github.com/Regalis11/Barotrauma-development/commit/cc231dc181fa58762d56ba6e29fe354ee8d40313

PR #4915, opened 2021-01-27
Improve handling of redirected output on server code.

### LINKED ISSUE #2973: Coloured log output of dedicated server is unmanageable on Linux/Unix
- [ x] I have searched the issue tracker to check if the issue has already been reported.

**Description**
Coloured log output from the dedicated server (that is, the logs from the game session, clients disconnecting etc) will only appear in an unblemished tty on unix.

**Steps To Reproduce**
```bash
./DedicatedServer 2>&1 > test.log
./DedicatedServer > test.log
./DedicatedServer 2> test.log
./DedicatedServer 2>&1 | tee test.log
```
None of these commands produce the same output in the expected place as simply `./DedicatedServer` does.

**Version**
Server reports `0.9.9.1`. Reproduced  on:

- arch linux, kernel 5.6.13
- ubuntu, kernel 4.15.0

someone972:
Rather than not outputting any text when output has been redirected, properly
wraps calls to unsupported Console properties when using redirected output.
This enables piping output to tee or docker logs without all the logs missing.
Command input is disabled for both redirected output and redirected input
scenarios.

Also, a small bugfix to reset the autocomplete on backspace being hit was made,
which enables proper autocomplete when removing characters. For example, if I
had typed "exi", but hit backspace twice, I would still only get "exit" as an
autocomplete suggestion. Adding ResetAutoComplete(); to the backspace handler
fixes this.

Fixes #2973

### DIFF
--- a/Barotrauma/BarotraumaServer/ServerSource/DebugConsole.cs
+++ b/Barotrauma/BarotraumaServer/ServerSource/DebugConsole.cs
@@ -73,17 +73,27 @@ namespace Barotrauma
             Stopwatch sw = new Stopwatch();
             sw.Start();
 
-            int consoleWidth = Console.WindowWidth;
-            if (consoleWidth < 5) consoleWidth = 5;
-            int consoleHeight = Console.WindowHeight;
-            if (consoleHeight < 5) consoleHeight = 5;
+            int consoleWidth = 0;
+            int consoleHeight = 0;
+
+            if(!Console.IsOutputRedirected)
+            {
+                consoleWidth = Console.WindowWidth;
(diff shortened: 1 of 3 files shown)


someone972:
Note: I did test this locally, but additional testing with the dedicated server might be desirable on various OSes. I see no reason why it would have problems, given that it's just Console.Write* calls though.

juanjp600:
Merged into our private repo (https://github.com/Regalis11/Barotrauma-development/commit/d585d8285fa227ab45e1503dd813e89eab77086c)

juanjp600 (commented):
No issues found.

PR #4917, opened 2021-01-27
WIP: Use ternary states for some of the server filters.

### LINKED ISSUE #3652: Server Filters Should Use Ternary Switch System
[Feature Request, Server Browser]
A good suggestion popped up on the discord, server filters should probably have YES, NO, and EITHER/DON'T CARE states. This would allow for the most filtering possible, and allow people to filter categories just the way they want, instead of the somewhat inconsistent negative and positive filters that the game currently uses.
image

someone972:
This is a preliminary attempt at making some of the server filters like Karma
and Traitors use ternary options (All/Enabled/Disabled) instead of a tickbox
which only gives you two of the states. I went with the approach of using drop
down boxes instead of trying to implement a third state into GUITickBox. I
couldn't think of a good way to graphically represent the third state, thus
the usage of drop downs.

This change is functional, but there are some spots where it needs attention.
I left TODO comments where I was unsure of how to proceed. This could be used
as a jumping-off point for the devs to implement this feature fully if they
like the approach.

Translation requirements:
- Needs a new 'servertag.X.label' for the tags affected here. For example,
  "Traitors".
- Needs 'servertag.X.true' to be modified to not include the label. For example,
  "Traitors enabled" -> "Enabled"
- Needs 'servertag.X.false' to be modified to not include the label. For example,
  "Traitors disabled" -> "Disabled"

Fixes #3652

### DIFF
--- a/Barotrauma/BarotraumaClient/ClientSource/Screens/ServerListScreen.cs
+++ b/Barotrauma/BarotraumaClient/ClientSource/Screens/ServerListScreen.cs
@@ -43,6 +43,13 @@ namespace Barotrauma
         private Dictionary<ulong, Steamworks.Ugc.Item?> pendingWorkshopDownloads = null;
         private string autoConnectName; private string autoConnectEndpoint;
 
+        private enum TernaryOption
+        {
+            Any,
+            Enabled,
+            Disabled
+        }
+
         private class FriendInfo
         {
             public UInt64 SteamID;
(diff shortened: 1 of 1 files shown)


someone972:
Here's an example of what this looks like in context:
ExampleTernary

juanjp600 (approved):
We can address the localization and styling TODOs ourselves, no other issues found.

juanjp600:
We've incorporated this as a separate branch in our private repo and will be making the final changes ourselves, thanks for your pull request!

PR #5005, opened 2021-02-07
Refactor Signal system

Jlobblet:
Apologies in advance, this is going to be a lengthy PR.

### TL;DR (or "Executive Summary")

- All parts of a signal are now bundled up in the new `Signal` struct.
- `ReceiveSignal` and `SendSignal` methods have been refactored to use this struct.
- There no changes to the behaviour of the wiring system, only refactorings to the code.

### Motivation

Currently, the concept of a signal encompasses multiple parts. This includes the signal itself, the number of steps taken, the item source, the sender, the power and signal strength, and what connection the signal is being sent to. As a result of this, the method signatures of methods that interact with signals are quite messy. 

Take, for example, the virtual `ReceiveSignal` method of `ItemComponent`:
https://github.com/Regalis11/Barotrauma/blob/c2e181ee6dd5799196cf0ea45fafe14e8b58ddbf/Barotrauma/BarotraumaShared/SharedSource/Items/Components/ItemComponent.cs#L429
`ReceiveSignal` contains all these parts. However, most overrides of `ReceiveSignal` only use a few fields.
https://github.com/Regalis11/Barotrauma/blob/c2e181ee6dd5799196cf0ea45fafe14e8b58ddbf/Barotrauma/BarotraumaShared/SharedSource/Items/Components/Signal/AndComponent.cs#L62-L78
In this case, only `signal` and `connection` are used. This makes the function signature quite cumbersome and signals themselves hard to expand on if necessary.

This problem is further compounded by the `SendSignal` method of `Item`, which contains three different overloads all doing ever so slightly different things, and the `SendSignal` method of `Connection`.

https://github.com/Regalis11/Barotrauma/blob/c2e181ee6dd5799196cf0ea45fafe14e8b58ddbf/Barotrauma/BarotraumaShared/SharedSource/Items/Item.cs#L1869
This method is responsible for sending a signal to an unknown `Connection` with a known name. It's used extensively in the outputs of components to send their signals to a `Connection` defined in the prefab. I will call this `Item.SendSignal(1)`.

https://github.com/Regalis11/Barotrauma/blob/c2e181ee6dd5799196cf0ea45fafe14e8b58ddbf/Barotrauma/BarotraumaShared/SharedSource/Items/Item.cs#L1876
This method sends signals to a known `Connection` by invoking the `Connection`'s `SendSignal`. I'll call this `Item.SendSignal(2)`.

https://github.com/Regalis11/Barotrauma/blob/c2e181ee6dd5799196cf0ea45fafe14e8b58ddbf/Barotrauma/BarotraumaShared/SharedSource/Items/Components/Signal/Connection.cs#L251
`Connection`'s `SendSignal` relays the input signal on to other `Connection`s that are wired to this one. The only thing it changes about the signal is its connection. I will call this `Connection.SendSignal`.

https://github.com/Regalis11/Barotrauma/blob/c2e181ee6dd5799196cf0ea45fafe14e8b58ddbf/Barotrauma/BarotraumaShared/SharedSource/Items/Item.cs#L1899
The final `SendSignal` overload is used to delay a signal being sent by one frame to avoid infinite loops. Notably, it changes the signal's source to the component causing the delay. I'll call this `Item.SendSignal(3)`.

To reduce the complexity of these function signatures, I decided to implement a new struct.

### Changes Made

**Signal.cs struct**

The first change, and main underpinning feature of this pull request, is Signal.cs. It contains a struct `Signal` which collects all the parts of the signals defined above into one place. Over the course of developing this PR, I made some changes to it.
The original constructor closely matched the current order of parameters in function signatures:
https://github.com/Jlobblet/Barotrauma/blob/bc5e92974ff2f1adfde0438e4b306e93bd5a71b3/Barotrauma/BarotraumaShared/SharedSource/Items/Components/Signal/Signal.cs#L13

However, as I was exploring the signal system, I noticed that not all parameters are used equally, and most typically have a default value. This made me refactor the constructor into its current form:
https://github.com/Jlobblet/Barotrauma/blob/b810ee665b93f3b8457528ae19e64908e9008ea0/Barotrauma/BarotraumaShared/SharedSource/Items/Components/Signal/Signal.cs#L13-L14

This version of the constructor has only a single required parameter, `value` (which replace the parameter `signal` in the original function signatures). The other values are initialised to default values. This is because the majority of the time I saw `Item.SendSignal(1)` it was used something like this:
```cs
item.SendSignal(0, "value", "signal_out", null);
```
Mostly from signal components. With this in mind, I moved the `value` of `Signal` to be the first parameter in its constructor and made `0` and `null` default values for `stepsTaken` and `sender` respectively.

With the struct containing signals organised, I moved on to the next part.

**ReceiveSignal**

My first target was `ReceiveSignal`, since it contains all the parts of a signal as is easiest to change.
Here's the new virtual `ReceiveSignal` in `ItemComponent`:
https://github.com/Jlobblet/Barotrauma/blob/b810ee665b93f3b8457528ae19e64908e9008ea0/Barotrauma/BarotraumaShared/SharedSource/Items/Components/ItemComponent.cs#L429-L452
Since `Signal` is a struct, it cannot be null. This means there is no need for defensive programming. In fact, attempting to check for null here causes a compiler error.
image
The rest of the method simple refers to `signal.value` and `signal.connection` etc. instead of `signal` and `connection`. This is repeated across all the overrides of `ReceiveSignal`.

This change means that `ReceiveSignal` is now a bit more resilient to any changes in signal structure - an additional field can be added to `Signal` and there would still be a single source of truth for what a signal is made of.

**SendSignal**

Refactoring the implementations of `SendSignal` proved to be a bit more complicated.

The easiest one to start with is `Connection.SendSignal`. The changes here are fairly minimal, just referring to fields inside `Signal` instead of arguments here.

`Item.SendSignal(1)` has been refactored to accept a `Signal` and a `string` connectionName (https://github.com/Jlobblet/Barotrauma/blob/b810ee665b93f3b8457528ae19e64908e9008ea0/Barotrauma/BarotraumaShared/SharedSource/Items/Item.cs#L1874-L1880). The `Signal`'s connection is overwritten based on the same logic as the original function. This is intended to be called with a null value for connection in the situations where the string connectionName method was used before, e.g.
```cs
item.SendSignal(new Signal("value"), "signal_out");
```
Since this pattern appeared a lot, I also created an overload (https://github.com/Jlobblet/Barotrauma/blob/b810ee665b93f3b8457528ae19e64908e9008ea0/Barotrauma/BarotraumaShared/SharedSource/Items/Item.cs#L1869-L1872) that accepts two strings - a value and a connection name. This simplifies the above to
```cs
item.SendSignal("value", "signal_out");
```
For situations where more detail about the `Signal` is required, I use the first of these two overloads. For example, I specify the sender in this snippet https://github.com/Jlobblet/Barotrauma/blob/b810ee665b93f3b8457528ae19e64908e9008ea0/Barotrauma/BarotraumaShared/SharedSource/Items/Components/Machines/Controller.cs#L432.

`Item.SendSignal(2)` now takes a single `Signal` as input and calls
```cs
signal.connection.SendSignal(signal);
```
now, passing it on to the connected `ItemComponent`s.

Finally, I renamed `Item.SendSignal(3)` to `DelaySignal` to better reflect what its purpose is. The new method is identical apart from the new name and function signature, again taking a `Signal` instead of many parameters.

### Testing the Changes

I have tried several different things to make sure that the new method offers exactly the same functionality as the last. At the time of writing this PR, I have not been able to find any differences.
I have tested:
- In-game text chat
- Power systems
- Basic wiring (vanilla submarine)
- Complex wiring (WorkingJoe's Barotronix Game of Life for example)

If I have missed anything, I am more than happy to fix it in this PR.

I have also kept a close eye on performance. I have not noticed any differences between the current vanilla patch (0.11.0.10) and this version. From what I can tell there are not any issues with memory allocation as a result of this change.

### The End

Thank you for reading my PR! I realise this is quite long - I wanted to make the intention behind these changes clear.


Barotrauma/BarotraumaShared/SharedSource/Items/Components/Machines/Controller.cs
```diff
@@ -277,7 +277,7 @@ public override bool Use(float deltaTime, Character activator = null)
                 return false;
             }
 
-            item.SendSignal(0, "1", "trigger_out", user);
+            item.SendSignal("1", "trigger_out");
```
Regalis11, line 280: The user should be passed to item.SendSignal here

Barotrauma/BarotraumaShared/SharedSource/Items/Components/Signal/ArithmeticComponent.cs
```diff
@@ -67,23 +67,24 @@ sealed public override void Update(float deltaTime, Camera cam)
             float output = Calculate(receivedSignal[0], receivedSignal[1]);
             if (MathUtils.IsValid(output))
             {
-                item.SendSignal(0, MathHelper.Clamp(output, ClampMin, ClampMax).ToString("G", CultureInfo.InvariantCulture), "signal_out", null);
+                item.SendSignal(MathHelper.Clamp(output, ClampMin, ClampMax).ToString("G", CultureInfo.InvariantCulture), "signal_out");
+                item.SendSignal(null, "");
```
Regalis11, line 71: Is this supposed to be here?

Barotrauma/BarotraumaShared/SharedSource/Items/Item.cs
```diff
@@ -1890,42 +1890,54 @@ public bool TryFindController(out Controller controller)
             return controller != null;
         }
 
-        public void SendSignal(int stepsTaken, string signal, string connectionName, Character sender, float power = 0.0f, Item source = null, float signalStrength = 1.0f)
+        public void SendSignal(string signal, string connectionName)
+        {
+            SendSignal(new Signal(signal), connectionName);
+        }
+
+        public void SendSignal(Signal signal, string conenctionName)
```
Regalis11, line 1898: Typo in the parameter name

Regalis11 (changes requested):
Really nice work, thanks for the PR! I only spotted a couple of minor issues, aside from these I think this would be ready to merge.

Barotrauma/BarotraumaShared/SharedSource/Items/Components/Machines/Controller.cs
Jlobblet: I thought I'd caught all of these - evidently I missed one. Fixed in 3f50c6a8bfbd095d4f231e2fbd11986ff072fc37.

Barotrauma/BarotraumaShared/SharedSource/Items/Components/Signal/ArithmeticComponent.cs
Jlobblet: Not sure how that got there. Fixed in ce4d2a297a90962969e167faf5ae58f7597ce9f4.

Barotrauma/BarotraumaShared/SharedSource/Items/Item.cs
Jlobblet: Oops! Fixed in 8b597c8c311232d36b684193ddf9c82752a69e28.

Regalis11:
Noticed a more major issue with this: using the SendSignal(string signal, string connectionName) method discards important information in many cases, including the steps taken by the signal, the character who sent the signal and the item originated from.

Jlobblet:
Currently waiting for the 0.1300 changes to get pushed here - happy to make the above changes (I agree with connections not being part of a signal on reflection) after it's public.

Jlobblet:
When attempting to build on 24cbef485ac6a43b166e27b3fbef5d0f8d3478d6 I get the following errors:
image
[JetBrainsLog.ReSharperBuild_2021_02_25_15_33_49_6751.log](https://github.com/Regalis11/Barotrauma/files/6044035/JetBrainsLog.ReSharperBuild_2021_02_25_15_33_49_6751.log)

Regalis11:
Couldn't find any issues after the latest changes, merged to our private repo now. Thank you!

Issue #2688, opened 2020-04-24
Constantly low FPS


Melolatius:
- [X] I have searched the issue tracker to check if the issue has already been reported.

**Description**
So, my friend has very low FPS during the whole game (10FPS, max. was about 17FPS) no matter what sub we use. His specs are
-Windows 10 Home
-CPU Intel Core i7 4720HQ @2,60gH 4 cores, 8 threads, 6mb cache,  8GB DDR3
-GPU Nvidia Geforce GTX 960M

His PC is way more better than mine, however I can play the game perfectly.

**Steps To Reproduce**
-/-

**Version**
Windows 10 Home
![grafik](https://user-images.githubusercontent.com/63376072/80208849-55400580-8631-11ea-940c-6f53288829f7.png)

**Additional information**
He has reinstalled the game already and checked it for errors, but this didnt help

BurykinNikolay:
For Rx 480 same issue

Melolatius:
So, does anyone know how to fix this?

Melolatius:
So, any new information about the issue?

EtiamNullam:
Are other games running fine?

I would start by checking both CPU and GPU temperature, make sure its not throttling due to overheating.

Then it might be worth checking where is the bottleneck, even a quick glance in the `Task Manager` could be helpful, for example:
- Is CPU usage at 100%?
- Is memory full?
- Is GPU usage at 100%?
- Is the correct GPU under load?
- Are there some background processes that slows down PC?
- Are you sure external GPU is rendering the game, and not the GPU integrated in CPU? It probably should still be enough for more than 15 FPS though.
- Are drivers up to date?

I'm not saying its impossible that its some problem with the game, but it works for a great majority here even on weaker PCs so its probably something on your end. Performance also seems better in the recent updates so maybe it already got better for you?

thegrb93:
Monitor cpu temp too. Hot cpu will throttle down.

Melolatius:
Soo

His laptop was using the CPU integrated GPU, not the external Nvidia GPU
Thanks a lot, he now has 70FPS

PR #5040, opened 2021-02-12
Enable nvidia optimus on Windows.

PJB3005:
While nvidia's only documented way for end applications to enable optimus appears to be exporting the NvOptimusEnablement symbol (not easily doable with a C# project...), it turns out that `LoadLibrary("nvapi64.dll")` before initializing any rendering stuff also works!

This isn't technically supported probably but hey, better than nothing?

For what it's worth, I've been using this [in my own project](https://github.com/space-wizards/RobustToolbox/blob/8f870403d250482ae388d7277346c7733e7a8952/Robust.Client/Graphics/Clyde/Clyde.DllMap.cs#L14-L28) with no reports of problems yet. Seems to work pretty smoothly.

This should fix problems like #2688 without requiring users to manually add an application override for Barotrauma.exe.

[Relevant StackOverflow post](https://stackoverflow.com/a/46749111/4678631)

### DIFF
--- a/Barotrauma/BarotraumaClient/ClientSource/Program.cs
+++ b/Barotrauma/BarotraumaClient/ClientSource/Program.cs
@@ -54,6 +54,7 @@ namespace Barotrauma
             executableDir = Path.GetDirectoryName(System.Reflection.Assembly.GetEntryAssembly().Location);
             Directory.SetCurrentDirectory(executableDir);
             SteamManager.Initialize();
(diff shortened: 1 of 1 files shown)


juanjp600:
Merged into our private repo (https://github.com/Regalis11/Barotrauma-development/commit/5941950619bff3e76647cd05a8e9a2adc1153a1e), thanks for your contribution!

PR #5059, opened 2021-02-14
Fix missing parens in TraitorMission.cs

Jlobblet:
While reading through the traitor code I noticed this small bug. According to https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/expressions#operator-precedence-and-associativity, the `||` operator has a higher precedence than the `??` operator, so it's evaluated first. This means the current boolean expression is the same as
```cs
traitor.Character?.IsDead ?? (true || traitor.Character.Removed)
```
which in turn is
```cs
traitor.Character?.IsDead ?? true
```
Cutting out the removed check. Adding the brackets in corrects the evaluation order.

Then I thought "better yet, make it clear what's actually being checked" and separated it out to 3 parts.

### DIFF
--- a/Barotrauma/BarotraumaServer/ServerSource/Traitors/TraitorMission.cs
+++ b/Barotrauma/BarotraumaServer/ServerSource/Traitors/TraitorMission.cs
@@ -249,7 +249,7 @@ namespace Barotrauma
                 {
                     return;
                 }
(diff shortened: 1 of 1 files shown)


Regalis11:
Nice catch! Cherry-picked this to our private repo (https://github.com/Regalis11/Barotrauma-development/commit/2a89260df6225cd50ab3b6c0928e5de662a835ee)

PR #5287, opened 2021-03-21
Added KeepCorpse bool to HuskAfflictionPrefab

Working-Joe:
To my knowledge is currently no sure-fire way for modders to spawn one character on the death of another without overriding existing characters. Overriding existing characters is usually not recommended, as it can lead to conflicts with other mods (or new content). This makes it hard to for instance make a parasitic affliction where an infected character spawns a new parasite when the host dies.

This small PR adds the KeepCorpse bool to the HuskAfflictionPrefab. If this bool is set to true, instead of copying the inventory of the original character over to a newly spawned husk character, the corpse of the old character is not removed and the husk character is simply spawned alongside it.

### DIFF
--- a/Barotrauma/BarotraumaShared/SharedSource/Characters/Health/Afflictions/AfflictionHusk.cs
+++ b/Barotrauma/BarotraumaShared/SharedSource/Characters/Health/Afflictions/AfflictionHusk.cs
@@ -183,8 +183,11 @@ namespace Barotrauma
                 yield return CoroutineStatus.Success;
             }
 
-            character.Enabled = false;
-            Entity.Spawner.AddToRemoveQueue(character);
+            if (Prefab is AfflictionPrefabHusk { KeepCorpse: false })
+            {
+                character.Enabled = false;
+                Entity.Spawner.AddToRemoveQueue(character);
+            }
 
             string huskedSpeciesName = GetHuskedSpeciesName(character.SpeciesName, Prefab as AfflictionPrefabHusk);
             CharacterPrefab prefab = CharacterPrefab.FindBySpeciesName(huskedSpeciesName);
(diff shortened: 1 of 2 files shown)


itchyOwl:
If the intention is to change the husk infection so that it can spawn a new character when the character "turns" so that the corpse would remain and the parasite would spawn as a separate monster, then you are right: there's probably no way to do it. Without actually testing it and by just looking the code quickly, the changes seems ok to me.

Could you provide the modified files so that it would be easier to test this? Regarding the testing: we need to test both the vanilla husk infection and the modified husk infection and ensure that both work correctly.

Just to make sure: there is a generic way to spawn new character(s) for example when a character dies. See (simplified) .xml code from Hammerheadmatriarch.xml: 
image

Working-Joe:
It is indeed the intention to make an affliction spawn a separate monster from any afflicted character upon death. 

To test this so far I've tested the existing huskaffliction as is and a variant with the tag `keepcorpse="true"` added. Without the tag the huskaffliction works as usual, and with the tag the corpse remains in place while a separate husk is spawned with an empty inventory. 

I am aware of the possibility to use onDeath on a statuseffect on a character, but from my testing it is not possible to apply this to (any) character via a statuseffect. For instance, using an affliction defined as 

```xml
<Affliction
		name="spawnondeath"
		identifier="spawnondeath"
		type="resistance"
		isbuff="true"
		showiconthreshold="1000"
		showicontoothersthreshold="1000"
		showinhealthscannerthreshold="0"
		maxstrength="100"
		limbspecific="false">

	<Effect minstrength="0" maxstrength="100" strengthchange="0">
		<StatusEffect target="Character" type="OnDeath">
			<!-- XML to spawn different character here -->
		</StatusEffect>
	</Effect>
</Affliction>
```

does not actually result in anything spawning upon death of the afflicted character (Note that the code for the spawning was tested to work if `type="OnDeath"` is not used). The main goal here is to allow these type of afflictions to be made by instead using a HuskAffliction which guarantees a creature spawns on death. Additionally if you want any effects . 

That being said if it were possible to make a StatusEffect of type OnDeath work in this context, that might provide as good of a solution. But I'm not familiar to say how doable this would be, since there might be some ordering issues with things like the afflictions being removed before the character dies.

Working-Joe:
I've had another quick look through the character/affliction code to see why the above Affliction would not work. 

The OnDeath statuseffects defined in the character XML are added to the `StatusEffects`  list of the character in its constructor:
https://github.com/Regalis11/Barotrauma/blob/694cdfee7bd986242b9a772cb3d1a8a65d19c29c/Barotrauma/BarotraumaShared/SharedSource/Characters/Character.cs#L964

When `Kill()` is called on the character, this calls the `ApplyStatusEffects` method to apply the StatusEffects of type `OnDeath`.
https://github.com/Regalis11/Barotrauma/blob/694cdfee7bd986242b9a772cb3d1a8a65d19c29c/Barotrauma/BarotraumaShared/SharedSource/Characters/Character.cs#L3393

However, `StatusEffects` is a `private readonly` list and therefore never altered after construction. Then the only way for Afflictions to apply StatusEffects on death would be through the use of the `OnDeathHandler OnDeath` of the character. This is also how it's done in `AfflictionHusk.cs`. However, for other afflictions it seems like code for this simply does not exist, so `OnDeath` statuseffects are never executed as they're not called by the Character's `Kill` method nor the Affliciton code.

Perhaps an alternative fix to make these Statuseffects of type `OnDeath` work in Affliction Effects is then to extend the Affliction code elsewhere to add the these StatusEffects using the target's `OnDeathHandler`. Should I make an issue for this?

Working-Joe:
Upon further inspection I think such a change might be somewhat involved. Mainly because you'd have to somehow add the effect of a StatusEffect to the `OnDeathHandler` Only once, and remove it once the statuseffect ends. This seems to be substantially different from how other StatusEffects are handled. 

A logical approach will be adding this code to Affliction.cs. You'd get code very similar to that in AfflictionHusk for subscribing `Character.OnDeath` to a method of Affliction.cs, which would apply those StatusEffects of the current effect that are of type `OnDeath`. Then once a different effect is reached the event should be unsubscribed agian.

Working-Joe:
Proposed an alternative solution in #5325. This PR can be closed if #5325 is implemented.

Working-Joe:
I believe #5325 is the better solution and as such will maintain only that PR.

PR #5391, opened 2021-03-30
Adds reactor fuel_left pin

Ecchi-Ki-Senpai:
Because we often want to know how much fuel there is left in the reactor.
The new pin will output the lowest condition of all contained rods. 

In practice, we can combine this information with fuelOut to know how much fuel there is left in the reactor and if we are using the desired amount of rods. Thus, players could design alarm systems for reactor fuel :)


Ecchi-Ki-Senpai:
Upon further thought, I don't think this is the best way to send the information we need. Will rework this to send a string containing the conditions of all rods instead of just the worst condition rod.
Something like: condition1,condition2,condition3,condition4,condition5
which can be parsed with a regex component

Oukido:
I am wondering, is item.condition a value between 0 and 100 like 100% or is it the actual durability of the item, for example thorium fuel rods with 200.

Ecchi-Ki-Senpai:
Ah yes, I should be using conditionpercentage instead

kubalakowski:
any updates on this?

Regalis11:
Implemented in https://github.com/Regalis11/Barotrauma-development/commit/ae6827838e0864280c9df0f34d25f8eb0bed75db. I went a slightly different approach, and instead made it output the sum of the condition percentages of the fuel rods. Outputting the lowest condition seems like it'd be much more complex to use - a "fuel low" warning for example would require a ton of additional logic if you had to take into account that the reactor may contain several rods in addition to the one that has a low condition.

Ecchi-Ki-Senpai:
I can think of at least one problematic case with the sum of the condition percentages in regard to wiring alarms for low_fuel warnings.

Considering 1 regular fuel rod unlocks 80% of reactor's power output potential, it can be required to have at least 2 rods for a submarine to operate at maximum capacity

Suppose you have a total_fuel_rod_percentage = 101% from 2 rods in the reactor
Knowing that the condition of a rod varies between [0,100]%, there are thus multiple combinations of condition % that would give a 101% result

Example A: Rod 1 condition = 50%, Rod 2 condition = 51 %, I think it is unanimous that this does not require a fuel low warning
Exampe B: Rod 1 condition = 100%, rod 2 condition = 1%, Players could consider to implement a warning for this situation since they might wanna keep the submarine at maximum reactor capacity

However, there are no ways to differentiate situation A from situation B simply from fuel_out and/or fuel_left data.
This is fixed if you consider to output the lowest condition of all the rods in the reactor. With this method, players can be warned that at least one fuel rod is about to run out which is the primary use I believe players want.

Moreover, simple logic can be used to not give false alarm if a rod is about to run out, but 1 rod is sufficient to operate the submarine (using the fuel_out data)

PR #5563, opened 2021-04-22
Allow player controlled monsters to eat bodies.

Emanthealmighty:
This PR request implements a few changes that allow monsters to grab dead bodies by pressing G. This will cause them to start eating the body, and they can choose to stop eating the body by pressing G again.

In order to be able to eat a body, the creature you are playing as must have a new boolean parameter set to true: "CanEat". If it is set to true, you can grab and eat bodies while controlling the monster.

Based off of my own suggestion https://github.com/Regalis11/Barotrauma/discussions/4763.


itchyOwl:
Merged to dev. Moving to QA.

PR #5582, opened 2021-04-23
Allow the husk infection to carry player control over to the husk.

Emanthealmighty:
This PR implements a hefty amount of changes (that have probably been terribly written) that allow the husk infection affliction to give players control of their husk.

Commit message

```
The husk infection has two new parameters: "ControlHusk" and "ControlException".
ControlHusk decides whether or not the player should be allowed to control their husk.
ControlException will prevent the player from controlling their husk if their species is in this list.

Added two console commands: "ignorehuskexceptions" and "restorehuskexceptions". These commands will toggle a new boolean parameter of the husk infection which lets it ignore any exception lists if set to true.

Added a new function for characters that returns the name of the player controlling them when called.
```

If merged, the standard husk infection xml should be updated to look like this.
```
  <AfflictionHusk
    name="Husk infection"
    identifier="huskinfection"
    description="Something dark and unpleasant moves in the mouth. They are rendered completely mute, save for occasional clicking sounds apparently emanating from deep within the throat."
    type="alieninfection"
    targets="human,crawler"
    huskedspeciesname="[speciesname]husk"
    causeofdeathdescription="Taken over by a husk parasite"
    selfcauseofdeathdescription="You have been taken over by the husk parasite."
    limbspecific="false"
    indicatorlimb="Torso"
    activationthreshold="0"
    showiconthreshold="40"
    karmachangeonapplied="-1"
    maxstrength="100"
    achievementonremoved="healhusk"
    controlhusk="true"
    controlexception="human"
    iconcolors="60,140,195,255;60,107,195,255;60,0,195,255">
    <Effect minstrength="0" maxstrength="100"
      maxvitalitydecrease="0"
      strengthchange="0.3"/>
    <icon texture="Content/UI/MainIconsAtlas.png" sourcerect="0,640,128,128" color="60,107,195,255" origin="0,0"/>
  </AfflictionHusk>
```
This would always allow players to control crawlerhusks if they are infected while playing as a crawler, but also allows servers that use the new command to let players play as human husks after being infected as well.

### DIFF
--- a/Barotrauma/BarotraumaShared/SharedSource/Characters/Character.cs
+++ b/Barotrauma/BarotraumaShared/SharedSource/Characters/Character.cs
@@ -87,14 +87,19 @@ namespace Barotrauma
             }
         }
 
+        /// <summary>
+        /// Returns the client that is currently controlling the character
+        /// </summary>
         public Client GetCharacterClient
         {
             get
             {
-                dynamic controller = null;
+                Client controller = null;
 #if SERVER
+                // Loop through every client in the server
                 foreach (Client c in GameMain.Server.ConnectedClients)
                 {
+                    // If the client's current controlled character is this one, then this character's controller is that client
                     if (c.Character == this)
                     {
                         controller = c;
--- a/Barotrauma/BarotraumaShared/SharedSource/Characters/Character.cs
+++ b/Barotrauma/BarotraumaShared/SharedSource/Characters/Character.cs
@@ -87,6 +87,26 @@ namespace Barotrauma
             }
         }
 
+        public Client GetCharacterClient
+        {
+            get
+            {
+                dynamic controller = null;
+#if SERVER
+                foreach (Client c in GameMain.Server.ConnectedClients)
+                {
+                    if (c.Character == this)
+                    {
+                        controller = c;
+                    }
+
+                }
+#endif
+                return controller;
+            }
+
+        }
+
         /// <summary>
         /// Is the character controlled by another human player (should always be false in single player)
         /// </summary>
--- a/Barotrauma/BarotraumaShared/SharedSource/Characters/Health/Afflictions/AfflictionHusk.cs
+++ b/Barotrauma/BarotraumaShared/SharedSource/Characters/Health/Afflictions/AfflictionHusk.cs
@@ -211,30 +211,28 @@ namespace Barotrauma
                 husk.Info.Character = husk;
                 husk.Info.TeamID = CharacterTeamType.None;
             }
-#if CLIENT
+            var CanControlCharacter = false;
             if (Prefab is AfflictionPrefabHusk huskPrefab)
             {
-                if (originalCharacter == Character.Controlled && huskPrefab.ControlHusk)
-                {
-                    if (huskPrefab.ControlException.None(s => s.Equals(originalSpecies, StringComparison.OrdinalIgnoreCase)) || IgnoreHuskExceptions == true)
-                    {
-                        Character.Controlled = husk;
-                    }
-                }
+                // Check if this specific husk infection allows the player to control their husk, then check if their species DOESN'T appear in the exception list or if the server is currently ignoring exceptions altogether.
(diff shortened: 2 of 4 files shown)


Barotrauma/BarotraumaShared/SharedSource/Characters/Character.cs
```diff
@@ -87,6 +87,26 @@ public bool IsRemotelyControlled
             }
         }
 
+        public Client GetCharacterClient
+        {
+            get
+            {
+                dynamic controller = null;
```
itchyOwl, line 94: Why the controller needs to be dynamic? Dynamic is used when the type may change or when it's unknown. Here it's Client. I'm sure there's a place for that in some applications, scripting etc, but personally I've never found any reason to use that keyword, because static typing and compile time errors are generally preferable over dynamic types and runtime errors.

Barotrauma/BarotraumaShared/SharedSource/Characters/Health/Afflictions/AfflictionHusk.cs
```diff
@@ -208,6 +211,30 @@ private IEnumerable<object> CreateAIHusk()
                 husk.Info.Character = husk;
                 husk.Info.TeamID = CharacterTeamType.None;
             }
+#if CLIENT
```
itchyOwl, line 214: There's unnecessary repetition here. I'd try to rewrite it so that the each logic statement is written once and only the parameters change. On the other hand, since the code blocks are meant for only either for client or for the server, would it make sense to move them from the shared class to dedicated client and server classes, since we do have those for AfflictionHusk? Perhaps using CreateAIHuskProjectSpecific methods, as has been the practice elsewhere in the code base?

Barotrauma/BarotraumaShared/SharedSource/Characters/Health/Afflictions/AfflictionPrefab.cs
```diff
@@ -94,6 +94,8 @@ public AfflictionPrefabHusk(XElement element, string filePath, Type type = null)
             SendMessages = element.GetAttributeBool("sendmessages", true);
             CauseSpeechImpediment = element.GetAttributeBool("causespeechimpediment", true);
             NeedsAir = element.GetAttributeBool("needsair", false);
+            ControlHusk = element.GetAttributeBool("controlhusk", true);
```
itchyOwl, line 97: Default to false?

itchyOwl (commented):
Didn't test it. Only read through the code. I think it might be a good addition to add players control the husks when they are turned. But should this be defined in the server settings?

Unless I misunderstood it, the exceptions are for preventing certain races to control the husks while others can control them. Is this related to some mod? I have difficulties imagining the purpose of it in the vanilla game.

Barotrauma/BarotraumaShared/SharedSource/Characters/Character.cs
Emanthealmighty: Changed it to "Client" instead of using dynamic.

Barotrauma/BarotraumaShared/SharedSource/Characters/Health/Afflictions/AfflictionHusk.cs
Emanthealmighty: I cleaned up the code to make the client and server code blocks as small as possible. To be completely honest I'm not sure how to go about the other method you suggested since I'm very new to this code base and C# in general.

Emanthealmighty:
> But should this be defined in the server settings?
The way I set it up for now allows xml modders to let their custom infections have players control their husked species without touching any commands or settings as well as letting them set up exceptions that can be ignored and restored on the fly through console commands depending on the current intended gameplay experience, which is (_probably?_) faster than having these changes defined by server settings instead.

itchyOwl:
We decided to only allow controlling the turned creatures of custom husk infections. Implemented in Regalis11/Barotrauma-development@7e01194.

itchyOwl:
Reopening and moving to QA.

PR #5602, opened 2021-04-24
Additional playable monsters functionality

Emanthealmighty:
This PR implements some changes that improve playable monster gameplay.

Commit message

```
Player monsters' sight is now dependent on the sight value defined in their character's AI parameters. The higher their sight the more their camera can zoom out.

Player monsters now have night vision.

Player monsters can see where submarines are if they are far away from them.

Player monsters can now manually make idle and attack noises by scrolling their mouse wheel.

```

There's really not much to say here apart from the addition of letting monsters see the submarine icon. I tried making this feature toggleable through a server setting but didn't really succeed, I'm not really sure if there should ever be a situation where player monsters shouldn't know where the submarine is, because monsters pretty much just exist to mess with the submarine, and if they can't find it, there would be nothing to do.

### DIFF
--- a/Barotrauma/BarotraumaClient/ClientSource/Characters/Character.cs
+++ b/Barotrauma/BarotraumaClient/ClientSource/Characters/Character.cs
@@ -258,8 +258,11 @@ namespace Barotrauma
                     }
                     else
                     {
-                        //increased visibility range when controlling large a non-humanoid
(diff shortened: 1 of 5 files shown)


StruggleDuck:
I really think that the players visual range should be set to that of the monster they are controlling. Manually controlling something like a Charybdis or god forbid an endworm is next to impossible because of the way the camera works.

PR #5618, opened 2021-04-26
Added new parameters to spawncharacter effect

Emanthealmighty:
This PR implements some new optional parameters for the spawncharacter effect.

Commit message

```
The spawncharacter function in status effects now accepts more parameters (all of them are set to 0 or false by default):

Stun: automatically stun each spawned character for the specified amount of time after spawning them

TransferController: automatically give control of the new character to the player character who triggered the effect

TransferAfflictions: automatically inflict every affliction the previous character had to the new character

TransferBuffs: automatically transfer every buff the previous character had to the new character

AfflictionOnSpawn: apply a specific affliction to the spawned character

AfflictionStrength: strength to give to the affliction defined in "AfflictionOnSpawn"

TransferInventory: attempt to move every item from the previous character's inventory to the new character's inventory

RemovePreviousCharacter: remove the character who triggered the effect.
```

TransferController and RemovePreviousCharacter will only be triggered, if set to true, when the last character is spawned if there are more than one being spawned.


Regalis11:
Thank you for your contribution! Finally got around to reviewing this and merging to our private repo. I'll keep the PR open until we've fully tested the additions.

NilanthAnimosus:
I've been testing this on 0.19.0.0 but honestly I am not very good at writing the XMLs as this is likely intended to be written as.

In my tests:

Transferring items to a husk results in item loss (Only ID cards transfer), but thats probably no valid slots to place the items into, but they are deleted if not transferred. - Converting a human into another human transfers items correctly however.

You cannot determine the team of the spawned character (IE. if you spawn a human, it is an enemy not an ally), this means anything spawned is considered an enemy regardless of what is done.

All transferred afflictions transfer to the chest (Instead of their respective limbs) and combine, which both changes the vitality multipliers (IE. leg to chest = worser affliction for humans) and can exceed the max for that affliction (presumably reducing the previous total amongst all limbs).

transferring control works fine (Though I didn't realize at first it had changed to "TransferControl"

Local mod I quickly whipped up testing values (Just a few items named CharTest1 to 6 I kept changing values of to try different permutations, though no in game description of what they do):
[character spawn test.zip](https://github.com/Regalis11/Barotrauma/files/9189686/character.spawn.test.zip)

Emanthealmighty:
> Transferring items to a husk results in item loss (Only ID cards transfer), but thats probably no valid slots to place the items into, but they are deleted if not transferred. - Converting a human into another human transfers items correctly however.

This is because the item transfer will only be able to occur if the characters have the same inventory slots, similarly to the husk infection's transformation (In fact, transforming a human into a humanhusk will transfer the items correctly, however the spawned human husk won't have the same name or appearance as the human it used to be). 

I noticed that failing to transfer the items actually prevents many other things from happening. The item "chartest2" doesn't actually let you control the spawned husk and delete your character, since it failed to transfer the items, and the code that was supposed to be executed gets cut off earlier by a return statement. This should probably be changed.

Additionally, if https://github.com/Regalis11/Barotrauma/pull/6542 is merged, it would be possible to remove the previous character in a separate "removecharacter" element of the status effect, and making its entire inventory drop that way, so that you can then grab whichever items you can put into your inventory after transforming.

> You cannot determine the team of the spawned character (IE. if you spawn a human, it is an enemy not an ally), this means anything spawned is considered an enemy regardless of what is done.

Specifying the team of a spawned human hasn't ever been possible with a spawncharacter effect, and this PR doesn't implement this feature either, but it could be added later on. As far as the current changes go, they are mainly meant to be used to transform humans into monsters whose "team" would be defined by their species group (as you can make monsters be friendly to humans if you set their group to "human"), and to transform monsters into other monsters while still letting the player controlling them stay in control.

> All transferred afflictions transfer to the chest (Instead of their respective limbs) and combine, which both changes the vitality multipliers (IE. leg to chest = worser affliction for humans) and can exceed the max for that affliction (presumably reducing the previous total amongst all limbs). note, this occured even if it was human to human. it makes sense for different species but it occurs even with the same species/limbs available.

I opted for transferring all the afflictions to the character's main limb because otherwise it would copy every affliction and paste it to every limb, usually killing the character instantly. By transferring every affliction to the main limb, you can get the spawned character to spawn with about as much health as the previous character, albeit a bit inaccurately. If the afflictions end up being more severe than intended after transferring them to the other character, you could always give the character some temporary healing buff through the "afflictiononspawn" parameter.

Regalis11:
All of the mentioned issues seem to be essentially missing functionality, but the ones added in this PR seem to be working correctly. Closing.

PR #5704, opened 2021-05-05
Fix some number inputs having incorrect initial button state

Smiley43210:
This PR simply fixes number inputs having an incorrect initial state.

Here are some examples of number inputs in an incorrect initial state:
image
**Expectation**: Increment buttons are disabled and decrement buttons are enabled
**Reality**: Increment buttons are enabled and decrement buttons are disabled

image
**Expectation**: Both increment and decrement buttons enabled on upper limit input
**Reality**: Only increment button is enabled on upper limit input

Here are those same inputs after the fix:
image
image

As a side effect of this change, any time the `IntValue` is set on a `GUINumberInput`, the value is clamped to [`MinValueInt`, `MaxValueInt`]. This could potentially cause issues if any inputs are used in such a way that the value is updated _before_ the bounds are.

### DIFF
--- a/Barotrauma/BarotraumaClient/ClientSource/GUI/GUINumberInput.cs
+++ b/Barotrauma/BarotraumaClient/ClientSource/GUI/GUINumberInput.cs
@@ -140,6 +140,7 @@ namespace Barotrauma
             {
                 if (value == intValue) { return; }
                 intValue = value;
+                ClampIntValue();
                 UpdateText();
             }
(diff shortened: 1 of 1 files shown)


Regalis11:
Thank you for the PR! Merged to our private repo in https://github.com/Regalis11/Barotrauma-development/commit/b5c0859a49f52e4b3358bfdd130027db85ee63c4

PR #5770, opened 2021-05-12
QTE Minigame for repairing

tobivv:
Adds a mini-game for repairing, while maintaining the original progress bar.

Gif of how it looks
[https://i.imgur.com/F9gErVW.gif](url)

How it works:

- Press Repairing button while the white cursor is over the colored section of the progress bar.
    Success: itemCondition += RelevantSkill(0-1) * 3
    Miss: itemCondition -= 1 - RelevantSkill  * 10
- The cursor turns red or green depending on the outcome and stays for a little bit (broken on the GIF but works in the commit). If the player rushes and clicks during this time, it counts as a miss to prevent mindless button mashing.
- The Cursor slows down near the end and if it reaches all the way to the left it just stops there, so even very unskilled players or players not understanding the mechanic can succeed.
- Since the player wants to do it as fast as possible misses are a probable outcome.

Tested and works on multiplayer.

### DIFF
--- a/Barotrauma/BarotraumaClient/ClientSource/Items/Components/Repairable.cs
+++ b/Barotrauma/BarotraumaClient/ClientSource/Items/Components/Repairable.cs
@@ -324,10 +324,12 @@ namespace Barotrauma.Items.Components
             {
                 return;
             }
-            RepairBoost(QTESuccess);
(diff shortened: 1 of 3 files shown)


SweptWasTaken:
The link to your gif doesn't embed.

F9gErVW

PR #5882, opened 2021-05-24
More oscillator waves

Jlobblet:
This PR adds two new waveforms to the oscillator component that are extremely useful but a bit of a pain to wire.

The first is the sawtooth wave, shown in the blue solid line on the graph. It increases linearly from 0 to 1 over the period of the wave.

The second is the triangle wave, shown in the red dashed line on the graph. It's like a sine wave but increases and decreases linearly instead, going from 0 to 1 to 0 to -1 to 0.

image


Regalis11:
Thank you for the PR! Merged to our private repo now.

PR #5892, opened 2021-05-25
Added High Pressure pin to Water Detector

Torca2001:
Addeds extra pin to water sensor named "high_pressure" outputs 1 if the current hull lethal pressure > 5.0f 
or the detector is not in a hull, else its 0.
I have done the necessary localization and modified the item xml the commit doesn't appear to transfer that.

### DIFF
--- a/Barotrauma/BarotraumaShared/SharedSource/Items/Components/Signal/WaterDetector.cs
+++ b/Barotrauma/BarotraumaShared/SharedSource/Items/Components/Signal/WaterDetector.cs
@@ -104,6 +104,8 @@ namespace Barotrauma.Items.Components
                 int waterPercentage = MathHelper.Clamp((int)Math.Round(item.CurrentHull.WaterPercentage), 0, 100);
                 item.SendSignal(waterPercentage.ToString(), "water_%");
             }
+            string highPressureOut = (item.CurrentHull == null || item.CurrentHull.LethalPressure > 5.0f) ? "1" : "0";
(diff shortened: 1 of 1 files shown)


Layarion:
Question, if the room is 100% but has no hull breach, does this still read high pressure? could you just make a "hull breach" detected pin instead? because i'm pretty sure high pressure still won't tell you if a hole exist in that room. it could be that a door was left open.

Torca2001:
> Question, if the room is 100% but has no hull breach, does this still read high pressure? could you just make a "hull breach" detected pin instead? because i'm pretty sure high pressure still won't tell you if a hole exist in that room. it could be that a door was left open.

This functions identically to the High Pressure affliction, what ever is considered high pressure by the affliction is the same for this sensor. Essentially rooms can only become high pressure whenever the room is 100% and there is still a breach to the outside. Or a room can stay high pressure even when sealed if it isn't drained at all.

Layarion:
Hmm. Well this is useful, but when I seen the title I was really hoping for a hull breach sensor...and sadly there's no amount of math, wiring, or tricks you can do with this or current tech to detect a hull breach once the room is full.

So what i said was, this is nice, but won't let my systems detect if it's safe to open the door to the next room with a pump. Really wish it did.

tiba666:
think this one should be by it self because what if you get high air pressure ^^ would it then also go off or not ? btw im total for the idear because then i could force close the door if high pressure and then add an override button to be enable to open the door

Regalis11:

PR #6212, opened 2021-06-29
Improve fabricator performance

PikaChokeMe:
This PR is made in an effort to improve fabricator performance especially in the cases where there are many linked inventories and/or possibly a lot of available to check whether they can fabricated or not. A lot of the improvements in this commit focus around the `CanBeFabricated` method, as well as converting the `GetAvailableIngredients` method to return a dictionary indexed by item identifier instead of just handing things a gigantic list to use all the time. A couple changes had to be made in a few other places to accommodate the new ingredient list format, and as an added bonus I also cleaned up the recipe display code so that it stacks required items of the same type with a count indicating how many of that item is required instead of just showing X amount of that individual item. 

I have done my best to test everything around the fabricator and how it works and I have also taken some benchmarks of the performance around this area before and after. 

Here's a benchmark of CPU and GPU usage on current master: 
image

It is quite easy to see when I am using the fabricator and the couple moments where I stop using it during the benchmark. The Hot Path down at the bottom of the diagnostic is already honing in on the `UpdateHUD` method and it's calls to the `CanBeFabricated` method.

And here's a benchmark of the CPU and GPU usage on this branch while fabricating for comparison:
image

Frame time and frame rate are both dramatically improved and it's quite possibly even impossible to tell at which point I actually opened the fabricator and started doing fabricator things. 

If we compare the CPU time more carefully between the two branches:

Main:
image
Post-Optimisation:
image

We can see that while `CanBeFabricated` is taking just over 56% of the CPU time on main, while on the post-optimised branch there is a significant reduction in CPU time being used on `UpdateHUD` and `CanBeFabricated`. `GetAvailableIngredients` actually became more than twice as expensive because of the time spent converting the list of ingredients into a dictionary from those lists, but as a result of almost doubling the cost of that extremely inexpensive method there was almost a tenfold reduction in the cost of the more expensive methods.  

Also, possibly important to note, memory usage has mostly remained the same. 

Main:
image
Post-Optimisation
image

And here's me showing off the new required item stacking and the overall improved framerate. It is important to note that items contained in linked storages aren't highlighting because I haven't opened any of those containers yet and that the number of storages linked to this fabricator are comprised of about 7 or 8 large storage lockers in total

https://user-images.githubusercontent.com/5521574/123829352-a0057f00-d8bf-11eb-8747-71b675b4b9ec.mp4

Sorry for the omega-long post. Please let me know what you think and I would be happy explain anything in more detail if there are any questions.

### DIFF
--- a/Barotrauma/BarotraumaShared/SharedSource/Items/Components/Machines/Fabricator.cs
+++ b/Barotrauma/BarotraumaShared/SharedSource/Items/Components/Machines/Fabricator.cs
@@ -290,9 +290,9 @@ namespace Barotrauma.Items.Components
                     {
                         foreach (ItemPrefab requiredPrefab in requiredItem.ItemPrefabs)
                         {
-                            if (!availableIngredients.ContainsKey(requiredPrefab.Name)) { continue; }
+                            if (!availableIngredients.ContainsKey(requiredPrefab.Identifier)) { continue; }
 
-                            var availablePrefabs = availableIngredients[requiredPrefab.Name];
+                            var availablePrefabs = availableIngredients[requiredPrefab.Identifier];
                             var availablePrefab = availablePrefabs.FirstOrDefault(potentialPrefab =>
                             {
                                 return potentialPrefab.ConditionPercentage >= requiredItem.MinCondition * 100.0f &&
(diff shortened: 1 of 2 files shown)


PikaChokeMe:
I have managed to test this branch with 4-5 levels of multiplayer campaign on a dedicated server with 5 friends all heavily using the fabricator and everything seemed to work fine.

Barotrauma/BarotraumaShared/SharedSource/Items/Components/Machines/Fabricator.cs
```diff
@@ -365,24 +372,35 @@ static void onItemSpawned(Item spawnedItem, Character user)
...
+                    var availablePrefabs = availableIngredients[requiredPrefab.Name];
+                    foreach (Item availablePrefab in availablePrefabs)
+                    {
+                        if (availablePrefab.Condition / availablePrefab.Prefab.Health >= requiredItem.MinCondition &&
+                            availablePrefab.Condition / availablePrefab.Prefab.Health <= requiredItem.MaxCondition)
+                        {
+                            availablePrefabsAmount++;
+                        }
+
+                        if (availablePrefabsAmount > requiredItem.Amount)
```
juanjp600, line 395: If I'm not mistaken, this should be `availablePrefabsAmount >= requiredItem.Amount`. Fixed on our private repo.

juanjp600:
Merged into our private repo (https://github.com/Regalis11/Barotrauma-development/commit/9ec8a8c629125b4a6cb2560b43b123bcdea42c32), thanks for your contribution!

PR #6252, opened 2021-07-05
make water detector round up for water%

Tsunder:
make water detector round up for water% (so that any amount of water will be at least 1%)


Regalis11:
Thank you for the PR! Merged to our private repo

Issue #6184, opened 2021-06-26
[Disconnect from a server] Unauthorized multithreaded access to RandSync.Server


Showdoze:
After some time playing in a multiplayer campaign with my friend (who was a host) i started to crash out of the server into a server browser every time a round ended (i.e Dock to an outpost, go on a mission, level transition, etc.) with an error message that i will post below, the issue seems like to be tied up to a save, since that didn't happen when i played by myself or on another save file, and persisted after restarting the game and even PC

Version 0.14.6.0

[event_error_log_client.log](https://github.com/Regalis11/Barotrauma/files/6720568/event_error_log_client.log)

the last part of the error log with the stack trace, it does say pirate mission, even tho the disconnect happened regardless of pirate missions existing in current round

```
6/25/2021 3:24:22 AM - Unauthorized multithreaded access to RandSync.Server
   at System.Environment.get_StackTrace()
   at Barotrauma.Rand.CheckRandThreadSafety(RandSync sync) in <DEV>\Barotrauma\BarotraumaShared\SharedSource\Utils\Rand.cs:line 47
   at Barotrauma.Rand.Range(Single minimum, Single maximum, RandSync sync) in <DEV>\Barotrauma\BarotraumaShared\SharedSource\Utils\Rand.cs:line 56
   at Barotrauma.PirateMission.GetRandomDifficultyModifiedElement(XElement parentElement, Single levelDifficulty, Single randomnessModifier) in <DEV>\Barotrauma\BarotraumaShared\SharedSource\Events\Missions\PirateMission.cs:line 142
   at Barotrauma.PirateMission.SetDifficulty(Single difficulty) in <DEV>\Barotrauma\BarotraumaShared\SharedSource\Events\Missions\PirateMission.cs:line 102
   at Barotrauma.PirateMission..ctor(MissionPrefab prefab, Location[] locations, Submarine sub) in <DEV>\Barotrauma\BarotraumaShared\SharedSource\Events\Missions\PirateMission.cs:line 90
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Object[] arguments, Signature sig, Boolean constructor, Boolean wrapExceptions)
   at System.Reflection.RuntimeConstructorInfo.Invoke(BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
[...]

PR #6278, opened 2021-07-08
Server password prompt improvements

quantus:
This PR contains two quality of life improvements to the server password prompt so that it's quicker to join a password protected server. The background story is that I experienced several disconnects due to #6184 and the password prompt really started to annoy me.


Regalis11:

PR #6329, opened 2021-07-17
"set_delay" input for Delay Components

mc-oofert:
Added a "set_delay" input to the delay component.
This input (obviously) sets the components delay to the signals value.

`<input name="set_delay" displayname="connection.setdelay" />`
Appears to be working on multiplayer.

(ignore the ".gitignore")

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@ bld/
 [Dd]ebug*/
 [Rr]elease*/
 *.o
+Content
 
 # Misc vs crap 
 *.v12.suo
--- a/Barotrauma/BarotraumaShared/SharedSource/Items/Components/Signal/DelayComponent.cs
+++ b/Barotrauma/BarotraumaShared/SharedSource/Items/Components/Signal/DelayComponent.cs
@@ -110,12 +110,13 @@ namespace Barotrauma.Items.Components
                     float val;
                     if (float.TryParse(signal.value, out val))
                     {
-                        if (signalQueue.Count > 0 && val != Delay)
+						float clampval = MathHelper.Clamp(val, 0, 60);
+                        if (signalQueue.Count > 0 && clampval != Delay)
                         {
                             prevQueuedSignal = null; //might not be the best but hey it works???
                             signalQueue.Clear();
                         }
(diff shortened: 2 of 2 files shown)


Working-Joe:
I'd love to see this change, but I think a few things are missing in this pull request. 

Your code does not seem to alter the `signalQueue` which stores the delayed signals. This could lead to problems as the size with which the queue is initialised depends on the delay. You'll see that if you now have a delay that is initially very short and you increase the delay, it can store too few signals, meaning that not all signals go through it. 

You'll also have to consider what you do with the signals that are currently in the queue. Currently existing signals just stay in the queue. This will cause weird behaviour if signals with a shorter delay are enqueued while signals with a large delay are still in the queue.

mc-oofert:
> I'd love to see this change, but I think a few things are missing in this pull request.
> 
> Your code does not seem to alter the `signalQueue` which stores the delayed signals. This could lead to problems as the size with which the queue is initialised depends on the delay. You'll see that if you now have a delay that is initially very short and you increase the delay, it can store too few signals, meaning that not all signals go through it.
> 
> You'll also have to consider what you do with the signals that are currently in the queue. Currently existing signals just stay in the queue. This will cause weird behaviour if signals with a shorter delay are enqueued while signals with a large delay are still in the queue.

Done, I think???

Working-Joe:
I think it's still incomplete. Now it clears the queue, but it still doesn't change the queue size.

mc-oofert:
> I think it's still incomplete. Now it clears the queue, but it still doesn't change the queue size.

Line 86 and 40 says otherwise

Working-Joe:
You're right, the queue size should be fine now! Now that you clear it setting `Delay` should be enough.

Did find one last bug. You currently clear the queue before you clamp the delay. This means that if your delay is at one of the clamped edges, it will be reset when a value outside the clamp in sent into `set_delay`.

e.g. wiring 70 into the delay will cause it to reset every frame because 70 != 60 and the stored delay is clamped to 60.

mc-oofert:
Done

Regalis11:
Thank you for the PR! Finally got around to merging this to our private repo.

PR #4466, opened 2020-12-10
Add Run key to toggle Sonar mode

Coryf88:
Add the Run key to toggle the Sonar's mode between Passive and Active.

PR #6395, opened 2021-08-01
revert Add Run key to toggle Sonar mode  #4466

Tsunder:
I've personally found this incredibly annoying, and several other users that aren't me have also found it annoying and about as useful as directional sonar.

reverts #4466


PaladynMinak:
How about just leaving option to bind custom key for this functionality? Personally I like it and would go for even more. I mean, manual/auto toggle should also have a key to bind and directonal ping too, so I could dedicate cursor to sonar circle all the time, and most of the time, to set course, with only other need to set direction of the ping.

Regalis11:
Agreed, merged to our private repo.

Jlobblet:
Related, it may be worth adding similar bindings for other objects that currently share keybindings, namely the reactor.

PR #6416, opened 2021-08-09
Added new "accessiblebyowner" attribute to inventories

Emanthealmighty:
This PR implements a new attribute for inventories which lets them be accessible to their owners when "accessiblewhenalive" is set to false.

Commit message

```
This new attribute is false by default and is only relevant if the "accessiblewhenalive" attribute is set to false.
If set to true, "accessiblebyowner" will allow the character to see their own inventory when player controlled, while keeping it hidden from others until death.

```

This would work well with player controlled humanoid monsters (such as husks) and with player controlled non-humanoid creatures that will eventually be allowed to hold items (such as updated fractal guardians).


Regalis11:
Thank you for the PR! Merged to our private repo in https://github.com/Regalis11/Barotrauma-development/commit/1dca835f38f8cb648db61e69e05c6cdbd8fe9e7b. Moving to QA

PR #6435, opened 2021-08-17
Fix ColorComponent HSV color calculation

quantus:
The RGB output from ColorComponent that is in HSV mode was incorrectly divided by 255. This caused the output to be way too dark.

### DIFF
--- a/Barotrauma/BarotraumaShared/SharedSource/Items/Components/Signal/ColorComponent.cs
+++ b/Barotrauma/BarotraumaShared/SharedSource/Items/Components/Signal/ColorComponent.cs
@@ -36,9 +36,9 @@ namespace Barotrauma.Items.Components
             if (UseHSV)
             {
                 Color hsvColor = ToolBox.HSVToRGB(signalR, signalG, signalB);
-                signalR = hsvColor.R / (float) byte.MaxValue;
-                signalG = hsvColor.G / (float) byte.MaxValue;
-                signalB = hsvColor.B / (float) byte.MaxValue;
+                signalR = hsvColor.R;
+                signalG = hsvColor.G;
+                signalB = hsvColor.B;
             }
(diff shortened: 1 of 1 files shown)


Working-Joe:
So you have to supply an A value in 0-1 as well, otherwise the output becomes [0-1],[0-1],[0-1],255. The lightcomponent can interpret [0-1],[0-1],[0-1],[0-1] as well as [0-255],[0-255],[0-255],[0-255] (or even hex)

quantus:
Sorry for late reply, and also for the lacking explanation in this PR.

So the issue I'm having is that when I give input `SIGNAL_R(hue)=0, SIGNAL_G(saturation)=1, SIGNAL_B(value)=1, SIGNAL_A=255` to the color component in HSV mode, I expect it to return color (R=255, G=0, B=0, A=255). Without the changes in this PR it returns R=1, G=0, B=0,A=255 which is unexpected to me.

After checking the code I think this extra division here is responsible for making the output values way too dim, and after its removal the component behaves as I would expect.

It's also possible that I have misunderstood how the signal component is supposed to behave in this case.

Working-Joe:
The problem is that by implementing this PR if you were to send in an alpha in [0-1] you end up getting a signal like [0-255],[0-255],[0-255],[0-1] which the lightcomponent will interpret as a very low alpha colour.

A better fix might be clamping the input alpha to 1 if in HSV mode. Then even if you incorrectly pass an alpha of 255 into the alpha input it will be clamped to 1 at the output and the output signal will still be in [0-1],[0-1],[0-1],[0-1]

quantus:
@Working-Joe Thank you for the comments. It seems the component works properly when the alpha is in `[0,1]` range, and so the issues here was simply just that I misunderstood how the component was supposed to work.

PR #6439, opened 2021-08-18
Added "revivecharacter" effect to status effects

Emanthealmighty:
This PR implements a new effect that can be executed by status effects called "ReviveCharacter". It is used to revive characters in a similar way to the "Revive" console command, which revives the character and puts its client back in control of their character.
This effect also accepts an attribute called "noseveredlimbs", which when set to true will prevent the character from being revived if any of their limbs have been severed.

Commit message

```
Added a new variable to clients, "DeadCharacter", which gets set to the character you were controlling prior to dying and starting to spectate.

Added a new usable effect to status effects, "ReviveCharacter", which also accepts an attribute "noseveredlimbs" (false by default).

When using the "ReviveCharacter" effect, the targeted character will be revived and whoever was controlling it prior to dying will be set in control of the character again. If "noseveredlimbs" is set to true, the character will only be revivable if it has all of its limbs intact.
```

This addition would allow the creation of new items, afflictions, etc. that can give players the ability to take back control of their character by being revived without the need for console commands.


Emanthealmighty:
Will probably redo later once I figure out some things.

Issue #3193, opened 2020-06-23
[Feature request] Make husks and huskcrawlers neutral to the infected if he/she at the later infection stages


Stallfighter:
- [+] I have searched the issue tracker to check if the issue has already been reported.

**Description**
As said in the description, i suggest to add husk neutrality IF NOT PROVOKEN to the player at stinger and probably mute stages of the infection. Its more or less logical (according to Europan, looks like this thing is capable to really sophisticated activities), adds additional flavor to the game and can be used in emergency (i mean its REALLY bad situation if you forced to intentionally infect yoursef) if necessary.

itchyOwl:
Yeah agree. Should be straightforward to do.

itchyOwl:
Implemented in Regalis11/Barotrauma-developmen@d2a805a.

Rokvach:
Tested, working correctly. Closing.

PR #6440, opened 2021-08-19
Added species group tags and affliction AI tags to AI targeting parameters

Emanthealmighty:
This PR implements some code changes to monster AI targeting that allow them to target entire species groups and/or characters with specific afflictions at specifics strengths.

The former can be done by adding a species group as a targeting tag and then adding "_group" immediately next to it to signify that it's targeting an entire group of characters.
The latter can be done by modifying an affliction's effect so that it includes a new parameter called "aitargetingtag" and setting it to something, and then referencing it in the character's targeting tags.

Commit message

```
AI targeting parameters now accept species groups as well as species names.
This means that a character can now target an entire species group instead of having to put each species name included in that group manually.
Example tags: "mudraptor_group",  "crawler_group".

Affliction effects now accept a new attribute called "aitargetingtag".
AI characters can now target affliction AI targeting tags. This means characters can react differently to other characters depending on whether or not they have an affliction at a certain effect.

Enemies can now remember who damaged them and how much damage was dealt to them by a specific character, and can also remember damage dealt to friendly characters nearby. Damage memories eventually fade away.

Targeting parameters now accept a new attribute, "MinDamageToIgnoreAfflictionTag".

If a character has an affliction tag which changes the behavior of an enemy (for example, change to idle state), this state will now be ignored by the AI if the character deals enough damage to it (depending on the value of "MinDamageToIgnoreAfflictionTag"), which will then go back to using the default state defined in the targeting parameters to deal with that character.
```

This would make it easier for mod makers to make certain groups of monsters attack each other without having to type in every single individual monster's species name present in the group.
In addition, it could allow to remove the hardcoded "husk" tag present in the AI targeting code in favor of the "husk_group" tag which should work the same way.

With the affliction targeting tags, instead, it would be possible to allow monsters to change their behavior completely when encountering a character that has a specific affliction AI targeting tag.

___

Example of a simplified crawler's AI targets using group targets. For those mudraptor species names that are defined, it will act accordingly. In case of encountering a new type of mudraptor (possibly a modded one) that is part of the mudraptor group, it will instead act according to what state is defined the mudraptor group tag.
```xml
  <ai combatstrength="100" sight="1" hearing="1.0">
    <target tag="mudraptor" state="Flee" priority="100" reactdistance="0" ignoreifnotinsamesub="True"/>
    <target tag="mudraptor_unarmored" state="Flee" priority="100" reactdistance="0" ignoreifnotinsamesub="True"/>
    <target tag="mudraptor_hatchling" state="Attack" priority="100" reactdistance="0" ignoreifnotinsamesub="True"/>
    <target tag="mudraptor_group" state="Attack" priority="100" reactdistance="0" ignoreifnotinsamesub="True"/> <!-- Generic mudraptor tag to use in case modders add new types of mudraptors-->
    <target tag="husk_group" state="Flee" priority="100" reactdistance="0" ignoreifnotinsamesub="True"/>
  </ai>
```

Example of a simplified husk's AI targets using affliction targeting tags.
```xml
  <ai combatstrength="100" sight="1" hearing="1.0">
    <target tag="huskinfected" state="Idle" priority="100" reactdistance="0" mindamagetoignoreafflictiontag="20" ignoreifnotinsamesub="True"/>
    <target tag="human" state="Attack" priority="100" reactdistance="0" ignoreifnotinsamesub="True"/>
    <target tag="crawler" state="Attack" priority="100" reactdistance="0" ignoreifnotinsamesub="True"/>
  </ai>
```

The "huskinfected" affliction targeting tag could be implemented like this.
```xml
    <AfflictionHusk
      <Effect minstrength="0" maxstrength="75" maxvitalitydecrease="0" strengthchange="0.3"/>
      <Effect minstrength="75" maxstrength="100" maxvitalitydecrease="0" strengthchange="0.3" aitargetingtag="huskinfected"/>
    </AfflictionHusk>
```
The end result of these two simplified code snippets is that when a husk encounters any character that has a husk infection affliction of strength higher or equal of 75%, they will ignore it, unless the infected character attacks the husk, in which case that husk and other nearby husks will attack the infected individual for some time until they become neutral again.

Resolves https://github.com/Regalis11/Barotrauma/issues/3193


itchyOwl:
I went through the code, but decided to rewrite it a bit. An implementation for #3193, based on your ideas can be found in Regalis11/Barotrauma-developmen@d2a805a.

I think using tags for the afflictions was a clever idea, and I took it almost as is. There was some unnecessary code, because the damage memory was already implemented in Character class. But it wasn't used by the monsters. Refactored a bit and used that instead of introducing duplicate logic.

The groups can come handy too, but I think we can just use the group names as they are so that species names have the priority over the group names. Keeps things clean and simple. Added those in Regalis11/Barotrauma-development@0e122f1

Also decided to refactor the code a bit to reduce duplicate logic. It's possible that I managed to broke something.

Issue #4856, opened 2021-01-21
[Modded] Error when transferring inventory items to husk monster when inventory sizes don't match


Emanthealmighty:
- [x] I have searched the issue tracker to check if the issue has already been reported.

**Description**
At the moment if a monster has an inventory and transforms into a husk, its items will be carried over to the husked monster.
If the husked monster also has an inventory of its own, it will try to add them to the inventory which has the items that were carried over by the non-husk monster. 
If the inventory is already full, the new items will spawn on the ground.

**If a monster that has an inventory transforms into a husk and has more or less inventory slots than the husked version, the game will throw an error.** This will not happen if one of them has no inventory at all.

Expected outcome:
If the non-husk monster has less slots than the husked one, then the items should transfer over normally like described above.
If the non-husk monster has more slots than the husked one, then much like how any items the husked monster tries to spawn when the inventory is already full, the excess items from the non-husk monster should drop on the ground.

**Steps To Reproduce**
1) Spawn a monster that has an inventory (crawler, for example)
2) Make sure the husked version doesn't have the same inventory slots as the normal version.
3) Transform the monster into a husk 

**Version**
Version: 0.11.0.10 - Public Branch. OS: Windows 10.

Emanthealmighty:
I tested this issue after removing the error, and everything seems to be working fine anyway, even if the inventories aren't transferred. Why does there _need_ to be an error that interrupts gameplay in the first place?

Regalis11:
Fixed in https://github.com/Regalis11/Barotrauma-development/commit/b42d3587af0baf65fa57b6b2abf2c5e433d01708. I decided to remove the error message altogether (not just for human characters). I think it was a remnant from back when the husk affliction was much less customizable and only usable on human characters, but it's no longer that necessary.

Rokvach:
Tested, working correctly. Closing.

PR #6444, opened 2021-08-19
Husk inventory mismatch error no longer happens for monsters

Emanthealmighty:
This PR slightly tweaks the conditions necessary for the game to throw out an error when transferring the inventory items of an infected character to an AI husk character, making it so that the error only triggers if the infected character is a human and not a monster. 

Commit message

```
The error that happens when creating an AI husk if the source character and its husked version don't have the same inventory sizes no longer happens on monsters and will only trigger for humans being turned into husks.
```

It makes sense for the error in question to appear for infected humans, due to the fact that they almost always will have valuable items that shouldn't be lost when being turned into a husk, but for monsters whose inventories are comprised of simple lootable items, it doesn't seem like their inventories should be valued the same way.

The implementation of this change would also mean that monster inventories are no longer required to have the same slots as any sort of husked variant that they have, giving more freedom to what kind of loot they can have without causing errors if one of said monsters turns into a husk.

Fixes https://github.com/Regalis11/Barotrauma/issues/4856.

### DIFF
--- a/Barotrauma/BarotraumaShared/SharedSource/Characters/Health/Afflictions/AfflictionHusk.cs
+++ b/Barotrauma/BarotraumaShared/SharedSource/Characters/Health/Afflictions/AfflictionHusk.cs
@@ -231,7 +231,7 @@ namespace Barotrauma
 
             if (character.Inventory != null && husk.Inventory != null)
             {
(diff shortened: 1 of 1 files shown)


Regalis11:
I decided to get rid of the error message altogether as mentioned in https://github.com/regalis11/barotrauma/issues/4856

PR #6542, opened 2021-09-02
Add "saveitems" and "containeridentifier" attribute to "removecharacter" effect

Emanthealmighty:
This PR implements two new attributes for the "removecharacter" effect which allows it to store the items of the removed character in a duffel bag (or a custom container of choice) prior to deleting the character.

Commit message

```
"saveitems", if set to true, will spawn a container right before removing the character that contains all of the character's items.
"containeridentifier", if specified, allows to spawn a custom container when "saveitems" is set to true.
```

This allows the "removecharacter" effect to be more forgiving for whatever character gets deleted by letting their inventory be saved, if so desired.

### DIFF
--- a/Barotrauma/BarotraumaShared/SharedSource/StatusEffects/StatusEffect.cs
+++ b/Barotrauma/BarotraumaShared/SharedSource/StatusEffects/StatusEffect.cs
@@ -256,6 +256,8 @@ namespace Barotrauma
 
         private readonly bool removeItem, removeCharacter, breakLimb, hideLimb;
         private readonly float hideLimbTimer;
+        private readonly bool removeCharacterAndSaveItems;
+        private readonly string saveItemsContainerIdentifier;
 
         public readonly ActionType type = ActionType.OnActive;
 
@@ -518,6 +520,8 @@ namespace Barotrauma
                         break;
                     case "removecharacter":
                         removeCharacter = true;
+                        removeCharacterAndSaveItems = subElement.GetAttributeBool("saveitems", false);
+                        saveItemsContainerIdentifier = subElement.GetAttributeString("containeridentifier", null);
                         break;
                     case "breaklimb":
                         breakLimb = true;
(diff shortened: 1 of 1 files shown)


Barotrauma/BarotraumaShared/SharedSource/StatusEffects/StatusEffect.cs
```diff
@@ -577,6 +577,8 @@ public AITrigger(XElement element)
 
         private readonly bool removeItem, dropContainedItems, dropItem, removeCharacter, breakLimb, hideLimb;
         private readonly float hideLimbTimer;
+        private readonly bool removeCharacterAndSaveItems;
+        private readonly string saveItemsContainerIdentifier;
```
Regalis11, line 581: In my opinion it's not necessary to have two attributes for controlling this: we could just allow the user to define the container and determine whether to save the items based on whether it's defined or not.

The identifier should also be of the type `Identifier`, not a string.

Barotrauma/BarotraumaShared/SharedSource/StatusEffects/StatusEffect.cs
```diff
@@ -2281,15 +2285,74 @@ void OnItemSpawned(Item newItem, ItemSpawnInfo itemSpawnInfo)
...
+                            SpawnDespawnContainer(character);
+                        }
                         Entity.Spawner?.AddEntityToRemoveQueue(character);
                     }
                     else if (target is Limb limb)
                     {
+                        if (removeCharacterAndSaveItems)
+                        {
+                            SpawnDespawnContainer(limb.character);
+                        }
```
Regalis11, line 2355: There's some code duplication here, could be simplified by storing the target character in a variable instead of implementing the logic separately for characters and limbs.

Barotrauma/BarotraumaShared/SharedSource/StatusEffects/StatusEffect.cs
```diff
@@ -2281,15 +2285,74 @@ void OnItemSpawned(Item newItem, ItemSpawnInfo itemSpawnInfo)
...
+                        foreach (Item inventoryItem in inventoryItems)
+                        {
+                            if (!itemContainer.Inventory.TryPutItem(inventoryItem, user: null, createNetworkEvent: true))
+                            {
+                                //if the item couldn't be put inside the despawn container, just drop it
+                                inventoryItem.Drop(dropper: character, createNetworkEvent: true);
+                            }
+                        }
+                    }
+                }
```
Regalis11, line 2338: In my opinion this should be moved into a separate method, because this method is already quite massive.

Barotrauma/BarotraumaShared/SharedSource/StatusEffects/StatusEffect.cs
```diff
@@ -2281,15 +2285,74 @@ void OnItemSpawned(Item newItem, ItemSpawnInfo itemSpawnInfo)
...
+                    else
+                    {
+                        containerPrefab = ItemPrefab.Prefabs.Find(me => me.Identifier == saveItemsContainerIdentifier);
+                    }
+
+                    if (containerPrefab == null)
+                    {
+                        if (saveItemsContainerIdentifier == null)
+                        {
+                            DebugConsole.NewMessage("Could not spawn a container for a removed character's items. No item with the tag \"despawncontainer\" or the identifier \"metalcrate\" found.", Color.Red);
```
Regalis11, line 2306: In my opinion this should use DebugConsole.ThrowError, it's a severe enough issue that it should pop up the debug console

Regalis11 (changes requested):

PR #6577, opened 2021-09-04
Burst Fire Carbine

Tsunder:
Not really sure if this is the right way to go about it since there's no /content/ folder for xml only contributions.

This is the content for a mod for a "Burst fire" style Dead-eye carbine that I'd like to submit as a modification to the base carbine.

The purpose is to make the carbine more comparable and "feel" good to use.

It makes the carbine a bit more accurate, but moreover, it turns it into a burst-fire style weapon. It also improves its overall firerate by a 2x.

It fires (up to) 4 shots over 0.3s, with an overall fire rate of 4 shots/s.

I also tried 3 shots over 0.3s but ended up using 4 shots because it uses the magazine without remainder.

Sample Video

https://user-images.githubusercontent.com/5103517/132077565-a1c151ef-8c6a-4c99-aacb-7c2946467b98.mp4


cshingle:
@Tsunder  I know this is an older PR. How does this balance against the SMG and Shotguns? I think more weapon options are good, but I feel like every weapon should have a purpose. Would this be an "upgrade" to the SMG like the shotgun has better and worse versions?

Tsunder:
in terms of raw DPS it's lower than both SMG and shotgun and dual wield revolvers

However, it's more accurate than smg and shotgun, and has enough impact for low health targets that it could have lower TTK than the SMG.

itchyOwl:
Added in the game with minor adjustments in Regalis11/Barotrauma-development@6339447. GJ!

itchyOwl:
Re-opened and moved to QA.

Rokvach:
Tested, working correctly. Closing.

Issue #4970, opened 2021-02-03
[Modding] NearbyCharacters target for status effects ignores conditionals for individual characters.


Emanthealmighty:
- [x] I have searched the issue tracker to check if the issue has already been reported.

**Description**
When using a "neabycharacters" tag, you may want to add a conditional to the status effect it belongs to, so for example if we want to cause a small AoE effect that will deal an affliction to only a certain species, we would add "conditional speciesname=[speciesname]", and any other conditional we want.

The problem is, however, that once the status effect recognizes one single character that meets these requirements, every other single character in range will have the status effect applied to them

**Steps To Reproduce**
1) Create a status effect with "nearbycharacters" as its target. Add a few conditionals to the status effect.
2) Run the status effect in game with a character nearby that meets the requirements we set earlier with the conditionals.
3) Notice how the character is affected by the status effect.
4) Spawn another character next to the previous one that does not meet the same requirements above.
5) Run the status effect again, notice how both of the characters get affected, while the expected outcome would have been that only the first character gets affected.

**Version**
Version: 0.11.0.10 - Public Branch. OS: Windows 10.

hubismal:
FYI there are a lot of workarounds for this. You can use nearbycharacters to afflict afflictions in an area and have the afflictions do whatever if a speciesname checks out.

Devs: please don't change the core functionality or else it'll break a lot of mods

Regalis11:
We're doing some cleaning on the repository and closing low-prio tickets we most likely won't be addressing at least for the game's full release.

As mentioned in the PR, currently the conditionals determine whether the status effect can execute, not what it can target, so in a sense this isn't a bug. As there are ways around this and since changing it would break a lot of mods, we're not planning to address this at least for the time being.

PR #6590, opened 2021-09-05
Nearbycharacters status effects now checks conditionals for each target character

Emanthealmighty:
This PR adds a new if statement to the code that applies status effects which allows "nearbycharacters" status effects to work as expected by checking if each target actually meets the required conditionals, rather than applying the effect to every character in range as long as there is a single character that meets the requirements.

Commit Message

```
Before applying a status effect, check if the singular target meets the required conditionals.

This prevents "nearbycharacters" status effects from affecting every single character in range instead of only those that meet the specified conditions if there are any nearby, which is a more expected result.

Doesn't affect the functionality of "nearbycharacters" status effects that don't have any conditionals.
```

Fixes https://github.com/Regalis11/Barotrauma/issues/4970.

### DIFF
--- a/Barotrauma/BarotraumaShared/SharedSource/StatusEffects/DelayedEffect.cs
+++ b/Barotrauma/BarotraumaShared/SharedSource/StatusEffects/DelayedEffect.cs
@@ -92,6 +92,7 @@ namespace Barotrauma
             currentTargets.Clear();
             foreach (ISerializableEntity target in targets)
             {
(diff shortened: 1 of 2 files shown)


Regalis11:
This introduces a pretty drastic chance to the way Conditionals work and breaks a lot of existing functionality. Currently the conditionals determine _whether the status effect can execute_, not what it can target, but this change would make it determine both.

I think some other type of approach is needed to address https://github.com/Regalis11/Barotrauma/issues/4970.

PR #6603, opened 2021-09-06
Add custom music types for monsters

TheNinjaScout:
This PR implements the ability to make monsters play certain music types when encountered, as well as a setting for music tracks to mute intensity background music while being played.

Commit Message

```
Characters now accept three new parameters:

- MusicType: Defaults to "monster".  If set to a string which corresponds to a music type found in a sounds xml file, a music that has that music type will begin to play when the monster is encountered by a player/submarine.

- MusicWeight: Defaults to 1. If set to a higher number, it will increase the chances of this monster's music playing over another monsters'.

- MusicRangeMultiplier: Defaults to 1. If set to a higher number, it will increase the minimum distance required before the monster's music will play.

Background music now accepts a new parameter:
- Minimumrequiredtimetoplay: Defaults to 30. Defines the minimum amount of time (which is added to the default 5 it takes to update the current music) that the current monster music must be played for.

Monster encounter music now has priority over "deep" and "flooded" music types, preventing it from stopping to play if the submarine becomes flooded or falls into crush depth.
```

This would allow for boss fight music and/or music specific to certain monster types to be easily implemented, spicing up the music portion of the game.

[Example mod](https://github.com/user-attachments/files/16272281/Musictestmod.zip)
[Mod functionality demonstration video](https://youtu.be/N5a7_F2IcJE)


Barotrauma/BarotraumaClient/ClientSource/Sounds/SoundPlayer.cs
```diff
@@ -655,6 +657,19 @@ private static void UpdateMusic(float deltaTime)
...
+                        for (int i = intensityTrackStartIndex; i < MaxMusicChannels; i++)
+                        {
+                            if (targetMusic[i] != null)
+                            {
+                                targetMusic[i] = null;
+                                break;
+                            }
+                        }
+                        continue;
+                    }
```
Regalis11, line 672: I think this code block is quite confusing.
- Why do we need to do this for each intensity music track inside the `foreach` loop, when it stops all of the intensity tracks?
- Why does it break from the for loop as soon as one non-null target music is found (leaving the rest playing?)
- Why continue, as opposed to break?
- What's the point of first finding the suitable intensity music, and then later muting all of the intensity layers if it's not intended to play?

To me it'd seem much clearer if the method first checked if the intensity music is supposed to play, and if not, mute all the intensity tracks and not run the rest of the intensity music logic.

Barotrauma/BarotraumaClient/ClientSource/Sounds/SoundPlayer.cs
```diff
@@ -26,6 +26,7 @@ static class SoundPlayer
         private static BackgroundMusic previousDefaultMusic;
 
         private static float updateMusicTimer;
+        private static bool increaseUpdateMusicTimer;
```
Regalis11, line 29: I feel this variable needs a more descriptive name and/or a comment: it's not obvious what it does without reading through the code. `IncreaseUpdateMusicInterval` would seem clearer to me (since what it does is increasing the interval at which the music is updated), and I would instead make this a float - now it's not clear how much it increases the interval, but if it was a float, it'd be clear that it means the amount of time the interval is increased by.

Barotrauma/BarotraumaClient/ClientSource/Sounds/SoundPlayer.cs
```diff
@@ -655,6 +657,19 @@ private static void UpdateMusic(float deltaTime)
 
                 foreach (BackgroundMusic intensityMusic in suitableIntensityMusic)
                 {
+                    //if the current maintrack is meant to mute all intensity music, loop through all the intensity channels and set them to null, then continue
+                    if ((targetMusic[mainTrackIndex] != null) && (targetMusic[mainTrackIndex].MuteIntensityMusic))
```
Regalis11, line 661: Unnecessary parentheses

Barotrauma/BarotraumaClient/ClientSource/Sounds/SoundPlayer.cs
```diff
@@ -670,6 +685,10 @@ private static void UpdateMusic(float deltaTime)
 
                 LogCurrentMusic();
                 updateMusicTimer = UpdateMusicInterval;
+                if (increaseUpdateMusicTimer)
+                {
+                    updateMusicTimer += (targetMusic[mainTrackIndex].MinimumRequiredTimeToPlay);
```
Regalis11, line 690: Unnecessary parentheses

Barotrauma/BarotraumaShared/SharedSource/Characters/Params/CharacterParams.cs
```diff
@@ -29,6 +29,15 @@ class CharacterParams : EditableParams
         [Serialize("", IsPropertySaveable.Yes, description: "If defined, different species of the same group consider each other friendly and do not attack each other."), Editable]
         public Identifier Group { get; private set; }
 
+        [Serialize("monster", IsPropertySaveable.Yes, description: "If changed, this character will try to play a custom music track when encountered."), Editable]
+        public string MusicType { get; private set; }
```
Regalis11, line 33: This should be an `Identifier` instead of the `string`, because it's converted to an Identifier in all of the situations it's used in.

Regalis11 (changes requested):
Finally got around to reviewing this, sorry for taking so long (I think this is a new record 😅). Overall this looks like a good addition and the code seems functional, but I think there are some code quality issues here that need to be addressed: some of the new logic in SoundPlayer in particular seems pretty confusing to me.

Barotrauma/BarotraumaClient/ClientSource/Sounds/SoundPlayer.cs
Emanthealmighty: I ended up removing this altogether in favour of just adding the current maintrack's value of "MuteIntensityMusic"  to "updateMusicTimer" regardless of whether or not it's monster music (which is what this bool was being used for)

Barotrauma/BarotraumaClient/ClientSource/Sounds/SoundPlayer.cs
Emanthealmighty: I ended up removing my implementation of "MuteIntensityMusic" altogether because it seems to have already been added to the game at some point during this PR's existence in the form of "MuteIntensityTracks" which does exactly the same thing and is already coded better, so I guess problem solved

Regalis11 (approved):
There are still some minor stylistic issues here (local variables being in PascalCase or all lower case instead of camelCase), but I went ahead and addressed those myself. Moved to a PR in our private repo: https://github.com/FakeFishGames/Barotrauma-development/pull/5752.

Thank you for your contribution!

Issue #4135, opened 2020-10-25
[Feature Request] - Add ability to specify a particular Wreck for Wreck Salvage missions


EricPfau:
Normally a wreck salvage mission selects a random wreck to be generated in the level. Adding the ability to specify a particular wreck to be loaded would allow modders to create "themed" missions. A specific example would be creating a wreck with traps or otherwise difficult to access areas that would hinder a player from salvaging them. This wreck could have a specific mission associated with it giving a greater reward for the increased difficulty.

Regalis11:
This has been implemented since this ticket was created, closing

PR #6639, opened 2021-09-12
Allow to specify which wreck to use in a mission that spawns a wreck

Emanthealmighty:
This PR implements the ability for wreck salvage missions (or any mission in general) to spawn a specific type of wreck rather than spawning a random one.

This PR also implements a new flag for submarines, "Don't spawn wreck randomly". If the submarine in question is a wreck and this flag is set to true, this wreck will never be encountered through random level generation, however it will still be possible to use it within missions.

Commit message

```
Missions now accept a new parameter called "requiredwrecktype".
This new parameter allows to specify which wrecked submarine to use within that mission.

Submarines now have a new flag, "DontSpawnWreckRandomly". If the submarine is a wreck and this flag was ticked before saving the submarine, the wreck will never be spawned randomly through the level generator, making it possible to instead make them only ever spawn through certain missions.
```

Example mission XML using the new parameter
```xml
  <SalvageMission
  name="Salvaging a wrecked dugong"
  identifier="salvagedugong"
  tags="salvage,salvagewreck"
  type="Salvage"
  description="Go check out this wrecked dugong. The payoff is huge!"
  sonarlabel="Wrecked dugong"
  spawntype="wreck"
  itemidentifier="clownmaskunique"
  containertag="locker"
  requirewreck="true"
  requiredwrecktype="Dugong_Wrecked"> <!-- Name is case sensitive -->
  </SalvageMission>
```

Resolves https://github.com/Regalis11/Barotrauma/issues/4135.


Emanthealmighty:
Closing this for the time being as I'll figure out how to make this work again at a later date.

PR #6757, opened 2021-09-26
Allow players to be spawned as non-human characters depending on their job

Emanthealmighty:
This PR implements some new parameters for jobs that allow players with said jobs to spawn in as non-human characters when not playing on the campaign gamemode.

Commit message

```
Job prefabs now accept a new element: "charactertype".
By setting up the subelements of this new element: "identifier" and "spawnoutside", it is now possible to make it so certain jobs will make the client spawn and respawn as a character other than a human (inside or outside the submarine, depending on the "spawnoutside" boolean) in any gamemode other than campaign.

Multiple character types can be set up in order to make the client spawn as different characters each time they are spawned / respawned.

If the character has the same inventory slots as a human, it will automatically equip the items that the human would have gotten with that job.

Removed the "hiddenjob" attribute in favor of "availability", which accepts three flags: "player", "npc", "both".

In multiplayer, bots will no longer have access to jobs with the parameter "availability" set to "player" or jobs that should turn the character into non-human character.
```

Example job XMLs using the new parameters
```xml
  <Job identifier="passivemudraptor"
     minnumber="0" maxnumber="1">
    <Charactertype identifier="Mudraptor_passive" spawnoutside="false"/>
  </Job>

  <Job identifier="playerpet"
     minnumber="0" maxnumber="2">
    <Charactertype identifier="Orangeboy" spawnoutside="false"/>
    <Charactertype identifier="Psilotoad" spawnoutside="false"/>
    <Charactertype identifier="Peanut" spawnoutside="false"/>
  </Job>
```
These job allow some of the players to play as friendly monsters, meaning that at the start of the round they will spawn in as one of the defined characters for that job, and if they die and respawn later, they'll respawn in the respawn shuttle / submarine as one of the defined characters once more.
It is also possible to set custom spawnpoints within the submarine for this job, like all other jobs, allowing mod makers to get creative with custom scenarios / gamemodes.

TODO: Disallow players from picking jobs that have any character types altogether when playing in the campaign gamemode.


Barotrauma/BarotraumaShared/SharedSource/Characters/Character.cs
```diff
@@ -1358,6 +1358,41 @@ public void GiveJobItems(WayPoint spawnPoint = null)
             info.Job.GiveJobItems(this, spawnPoint);
         }
 
+        // Replace this character with another one at the start of the round or after the respawn shuttle is spawned.
+        public IEnumerable<object> ReplaceWithMonsterJob(Client client, Dictionary<string, bool> characters)
+        {
+#if SERVER
+            string characteridentifier = characters.Keys.GetRandom(Rand.RandSync.Server);
```
juanjp600, line 1365: Perhaps unintuitively, `RandSync.Server` means "synchronized between the server and the clients", which means that if you're going to use this random number generator, the clients also have to do it in the same order and in the same amount. If this works just fine as server-only code, use `RandSync.Unsynced` instead.

Barotrauma/BarotraumaShared/SharedSource/Characters/Jobs/JobPrefab.cs
```diff
@@ -192,6 +193,14 @@ public bool HiddenJob
             private set;
         }
 
+        //whether the job should be only available to players
+        [Serialize(false, false)]
+        public bool PlayerOnly
```
juanjp600, line 198: Perhaps this could be merged with `HiddenJob` using an enum?
```cs
[Flags]
public enum Availability {
    Player = 0x1,
    Npc = 0x2,
    Both = 0x3
}

[Serialize(Availability.Both, false)]
public Availability AvailableTo;```

USMP-lancer:
i like this idea

DeltaNedas:
spawning as orange boy high quality pr

Emanthealmighty:
With https://github.com/Regalis11/Barotrauma/pull/5618 making its way to being included in the next public Barotrauma update, most of what was being proposed in this PR can essentially already be achieved through the changes in that one.

The only thing that wouldn't really be possible to without this PR is making bots unable to pick certain jobs exclusive to players, and making the spawned character spawn somewhere outside the submarine instead of inside.
I'll try to tackle the abovementioned in separate PRs at a later date, while closing this one, as I don't think it's really needed anymore.

PR #7103, opened 2021-10-31
Add regex to sub editor

james10424:
Adding regex support to the sub editor. You can still search just like before, but you can use regex expression.

### DIFF
--- a/Barotrauma/BarotraumaClient/ClientSource/Screens/SubEditorScreen.cs
+++ b/Barotrauma/BarotraumaClient/ClientSource/Screens/SubEditorScreen.cs
@@ -2978,7 +2978,7 @@ namespace Barotrauma
             Regex r;
             try
             {
-                r = new Regex(filter);
+                r = new Regex(filter, RegexOptions.IgnoreCase);
             }
             catch (ArgumentException)
             {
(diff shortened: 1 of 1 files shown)


Jlobblet:
I think the regex should probably run with the `i` (case insenstive) flag - most users don't really need or care about case sensitive search, and if necessary it can be enabled in the xpression.

james10424:
> I think the regex should probably run with the `i` (case insenstive) flag - most users don't really need or care about case sensitive search, and if necessary it can be enabled in the xpression.

Great suggestion, I've added the flag

Barotrauma/BarotraumaClient/ClientSource/Screens/SubEditorScreen.cs
```diff
@@ -2973,13 +2974,25 @@ private void FilterEntities(string filter)
...
+                r = new Regex(filter, RegexOptions.IgnoreCase);
+            }
+            catch (ArgumentException)
+            {
+                r = null;
+            }
+
             foreach (GUIComponent child in allEntityList.Content.Children)
             {
+                bool shouldShow = r?.IsMatch(((MapEntityPrefab)child.UserData).Name) ?? false;
```
Jlobblet, line 2990: As an additional thought, given that there is no feedback to the user if the regex fails to compile, it may be worth falling back to the string comparison in the case where `r` is `null` here.

james10424: Was thinking about that. Added the old comparison back

PR #7269, opened 2021-11-11
Allow relateditem to check for items in a specific character inventory slot

Emanthealmighty:
This PR implements the ability for requireditem conditionals to check if a certain item is found within a character's inventory in a specific type of slot. This allows for more precise conditions than were previously allowed by using "picked" and "equipped" types, as the first checks the character's entire inventory and the second only checks its hand slots.

```
The relateditem conditional now accepts a new parameter:
"characterinventoryslottype": a (case sensitive) string which should be set to one of the inventory slot types, which is the slot that will be targeted by the conditional. If not set, defaults to "None", which means every slot will be targeted.
```

Example:
```xml
<Requireditem item="abyssdivingsuit" type="picked" characterinventoryslottype="OuterClothes" />
```
This conditional, if put in a status effect which targets a character, will only be true if the "abyssdivingsuit" item is found in the character's inventory on the "OuterClothes" slot. This means that if you were to pick up the diving suit with your hands, you wouldn't get the status effect, as the item would be found in both of your hand slots.


Regalis11:
I don't quite understand the purpose of the `checkcharacterinventoryslot` attribute - why not always check the slot if the slottype attribute is given?

Emanthealmighty:
> I don't quite understand the purpose of the `checkcharacterinventoryslot` attribute - why not always check the slot if the slottype attribute is given?

I removed that attribute as you suggested: The inventory slot type to check now defaults to "none" if not set, and in that case, all inventory slots will be checked like normal (if the requireditem conditional type is "picked"). However if there is a custom slot type defined, then only that will be checked.

Regalis11:
Tested and merged our private repo in https://github.com/Regalis11/Barotrauma-development/commit/83217473bfd867636bb43ece807727842a3973d9, with 2 small modifications:
- Removed the unnecessary parentheses from `if ((item.ParentInventory?.Owner is Character character) && (CharacterInventorySlotType != InvSlotType.None))`
- Used `XElement.GetAttributeEnum` instead of doing the enum parsing manually.

PR #7281, opened 2021-11-12
Removed weird handling of hallway horizontal/vertical flags

dakkhuza:
When outposts generate you have no control over what hallways they will use, which can cause visual miss-matches to happen (e.g. a custom outpost using vanilla hallways). 

You can theoretically get around this by tagging all modules with a custom tag and allowing them only to connect to the custom tag, however [these lines of code](https://github.com/Regalis11/Barotrauma/blob/10e5fd5f3e5439f446f2ba42c41a1f094814678a/Barotrauma/BarotraumaShared/SharedSource/Map/Outposts/OutpostModuleInfo.cs#L89) prevent this from happening by exiting early before the other flags get the chance to be added. 

This prevents adding multiple tags to hallway modules both in the submarine editor and when loaded into the game, effectively killing any modded outpost that doesn't want to use vanilla hallways, for seemingly no reason.

### DIFF
--- a/Barotrauma/BarotraumaShared/SharedSource/Map/Outposts/OutpostModuleInfo.cs
+++ b/Barotrauma/BarotraumaShared/SharedSource/Map/Outposts/OutpostModuleInfo.cs
@@ -86,18 +86,6 @@ namespace Barotrauma
         public void SetFlags(IEnumerable<string> newFlags)
         {
             moduleFlags.Clear();
-            if (newFlags.Contains("hallwayhorizontal"))
-            {
-                moduleFlags.Add("hallwayhorizontal");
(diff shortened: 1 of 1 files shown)


Regalis11:
Finally got around to testing this. Doesn't seem to cause any issues, so I merged this to our private repo https://github.com/Regalis11/Barotrauma-development/commit/20391a81496fc98ad601e2cdaff0769709bb8687

> Removed weird handling of hallway horizontal/vertical flags (from PR #7281). I'm not entirely sure what the purpose of these returns was, possibly to prevent a designer from misusing the hallway tags in a way that causes hallway modules to be used as normal ones or vice versa? That's not a problem with any of the vanilla modules though, and it prevents using custom hallway modules in a custom outpost/ruin, so let's just remove them.

PR #7303, opened 2021-11-14
Add logic to allow junction boxes to send power and load signals

TBenz4Sportsball:
A very simple change to allow Junction Boxes to send their current power and load values out as signals. This does require xml changes as well. Adding:
`<output name="power_value_out" displayname="connection.powervalueout" />
<output name="load_value_out" displayname="connection.loadvalueout" />` (copied directly from reactor.xml)
to the ConnectionPanel section of the junctionbox item in poweritems.xml

This addition will allow more complex circuitry that checks the power load and available power in specific parts of the sub grid without relying on the "hidden reactor" trick occasionally employed by sub builders.

### DIFF
--- a/Barotrauma/BarotraumaShared/SharedSource/Items/Components/Power/PowerTransfer.cs
+++ b/Barotrauma/BarotraumaShared/SharedSource/Items/Components/Power/PowerTransfer.cs
@@ -172,6 +172,9 @@ namespace Barotrauma.Items.Components
             //if the item can't be fixed, don't allow it to break
             if (!item.Repairables.Any() || !CanBeOverloaded) { return; }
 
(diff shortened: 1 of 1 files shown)


ignomen:
This would be fantastic to have.

StruggleDuck:
+1 this would be epic

Regalis11:
Thank you for your contribution! Merged to our private repo in https://github.com/Regalis11/Barotrauma-development/commit/5477b52.

I did a couple of changes to the logic though: the values are rounded to the nearest integer to get them to match the values displayed in the junction boxes interface, and the output signal string is only instantiated when the value changes to reduce the amount of garbage the method generates.

PR #7516, opened 2021-11-27
Make round and ceil component return 0 instead of -0

### LINKED ISSUE #7515: Round component can output -0 (negative 0)

**Description**
Round component outputs -0 when fed with a number between (exclusive) -0 and -1

expectation is that it outputs 0, i think is reasonable?

**Steps To Reproduce**
feed a round component -0.2

**Version**
0.15.13

**Additional information**

Tsunder:
fixes #7515


Regalis11:

PR #7527, opened 2021-11-28
Modified ESC input interaction with the bug reporter message box

RadinQue:
Made changes to the ESC input behavior such that pressing it while the bug reporter message box is visible, the ESC key will close that instead of the pause menu.


VHZMC:
Yes please

Regalis11:

PR #7778, opened 2021-12-15
make sub turrets start round at mid-rotation

### LINKED ISSUE #708: Add more sources of noise detectable by passive sonar
Passive mode seems pretty useless in terms of vision when you are going in silent mode. None of the creatures or other things producing sound seem to be detectable by the sonars. Because of that you are completely blind when you are in passive mode and not making any noise come from your submarine.

It would be interesting to have some other sources of sonar noise. Exploding railgun shells, hydrothermal vents, creatures making noise. All that should be detectable on the sonar in passive mode.

The problem i see in that idea is that creatures may decide to follow each other or stay near any constant noise source indefinitely(like they do with the nasonov artifact) because their AI seems to be really attracted to noise. Perhaps you might find some way to work around that. 
But i don't see any problems in making anything, that only makes noise once and for a short time(like explosions), detectable by sonars. Creatures might get attracted to them, but they wouldn't get stuck in that position.

After more testing i found out that creatures don't seem to care at all if you shoot at them, they only seem to care about noise being produced by your submarine. This might be exploitable. You can just stay silent and shoot at creatures without expecting them to attack you. Railguns should make noise when they shoot.

### LINKED ISSUE #7083: Sub turrets always start round at angle 0

**Description**
Sub turrets  (coilguns/searchlights) start rounds at angle 0. Expectation is for them to start at halfway between min/max angles, like they appear in editor. OR start at the angle they were at from a previous round.

**Steps To Reproduce**
1. Start a round in campaign/mission mode
2. turrets are at angle 0 (or try to be)

**Version**
0.15.12 

**Additional information**

Tsunder:
targetrotation wasn't initialized (instantiated?) (apparently that means it's a 0!) so we explicitly set it to = rotation

fixes #7083

https://user-images.githubusercontent.com/5103517/146224707-2751523f-db0d-4b2b-b8c5-91b220f2cf66.mp4

3

### DIFF
--- a/Barotrauma/BarotraumaShared/SharedSource/Items/Components/Turret.cs
+++ b/Barotrauma/BarotraumaShared/SharedSource/Items/Components/Turret.cs
@@ -328,6 +328,7 @@ namespace Barotrauma.Items.Components
             FindLightComponent();
             if (loadedRotationLimits.HasValue) { RotationLimits = loadedRotationLimits.Value; }
             if (loadedBaseRotation.HasValue) { BaseRotation = loadedBaseRotation.Value; }
+            rotation = (minRotation + maxRotation) / 2;
             UpdateTransformedBarrelPos();
         }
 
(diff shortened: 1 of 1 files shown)


Tsunder:
doesn't work w/ flipped subs 
image

Tsunder:
fixed

Regalis11:
Thank you for the PR! Merged to our private repo in https://github.com/Regalis11/Barotrauma-development/commit/0d2d30b

PR #7810, opened 2021-12-16
add set output to greater than component, equals component

Tsunder:
requires adding the corresponding set_output pins on both components in signalcomponent.xml

      <input name="set_output" displayname="connection.setoutput" />


Regalis11:
Thank you for the PR! Merged to our private repo in https://github.com/Regalis11/Barotrauma-development/commit/3fe3056

PR #7869, opened 2021-12-20
Makes recipe hash generation more robust to reduce mod desync issues

ninjanomnom:
With large amount of mods even with the hash sorting there's still the issue of client and server recipe lists being desynced and blocking clients from crafting some recipes, or crafting the wrong one. If you have two recipes that make the same item they'll have the same hash, which in turn means that the mod order matters for those particular recipes. While changing load order can fix this to some extent it's good to reduce the downsides when load order is inevitably desynced.

This isn't really a great long term solution as recipes really need some form of load order independent id which the client could then communicate to the server, but it deals with most of the problem for now as long as no recipes share the same hash created by
`$"{Amount}|{outputId}|{RequiredTime}|{requiredItems}|{requiredSkills}"`

You may want to use something close to this to generate the hash in the end anyway. In the future only `GenerateHash()` should need to be modified.

### DIFF
--- a/Barotrauma/BarotraumaShared/SharedSource/Items/ItemPrefab.cs
+++ b/Barotrauma/BarotraumaShared/SharedSource/Items/ItemPrefab.cs
@@ -783,7 +783,16 @@ namespace Barotrauma
                 itemPrefab.FabricationRecipes.Clear();
                 foreach (XElement fabricationRecipe in itemPrefab.fabricationRecipeElements)
                 {
-                    itemPrefab.FabricationRecipes.Add(new FabricationRecipe(fabricationRecipe, itemPrefab));
+                    var newRecipe = new FabricationRecipe(fabricationRecipe, itemPrefab);
+                    var duplicateHashRecipe = itemPrefab.FabricationRecipes.FirstOrDefault(r => r.RecipeHash == newRecipe.RecipeHash);
+                    if(duplicateHashRecipe != null)
+                    {
+                        DebugConsole.AddWarning(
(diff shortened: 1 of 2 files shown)


Barotrauma/BarotraumaShared/SharedSource/Items/ItemPrefab.cs
```diff
@@ -185,6 +186,21 @@ public FabricationRecipe(XElement element, ItemPrefab itemPrefab)
...
+        {
+            var outputId = TargetItem.UIntIdentifier;
+
+            var requiredItems = string.Join(':', RequiredItems
+                .Select(i => i.ItemPrefabs.Select(p => p.UIntIdentifier))
+                .Select(i => string.Join(':', i)));
+
+            var requiredSkills = string.Join(':', RequiredSkills.Select(s => $"{s.Identifier}:{s.Level}"));
+
+            return $"{Amount}|{outputId}|{RequiredTime}|{requiredItems}|{requiredSkills}".GetHashCode();
```
juanjp600, line 203: This should be `ToolBox.StringToUInt32Hash` instead of `string.GetHashCode`. This is because GetHashCode's implementation is not guaranteed to be stable; in our case, since we use .NET Core 3.1, that means that GetHashCode is not even stable across executions, which means that the chances of all clients generating the same hashes for the recipes here is next to none.

ninjanomnom:
I added a check/warning of hashes as well so that you can know if you get overlap from currently installed mods. In the future I'd recommend storing a mod identifier in the recipe and including it in the hash as well in case two mods want to add identical recipes.

juanjp600:
This has been merged into our private repo. Thank you for your contribution!

PR #7872, opened 2021-12-20
Clears cause of death after revival and skill punishment

### LINKED ISSUE #7755: [0.15.20.0] Skill Levels being reset on death and after paying off reaper's tax
- [ ] I have searched the issue tracker to check if the issue has already been reported.

**Description**
1) So playing a multiplayer campaign after about an hour in, whenever a player die, they lose all their gained skill point. This may be due to the command "respawnnow" being used but I am not sure.

2) When you pay off reaper's tax, you lose all your gained skill points while the debuff is on the character. That or it reset the character.

**Version**
0.15.20.0

### LINKED ISSUE #7791: Skills resetting on multiplayer campaign round transition

**Description**
Skills are resetting for some, but not all, of my players on campaign round end. Just in case this helps, here's who's affected and who's not for my specific game.

| Player Job | Affected by bug? |
| --- | --- |
| Engineer, Host | Affected |
| Security Officer | Not Affected |
| Medic | Not Affected |
| Captain | Not Affected |
| Engineer | Affected |
| Mechanic | Affected |

**Steps To Reproduce**
- Level up a skill
- Dock at a station
- The round ends
- Observe that your skills are all reset to the baseline

**Version**
Windows v0.15.21.0

**Additional information**
This is on a custom sub with a no-delay docking port, just in case that matters.

### LINKED ISSUE #7843: Character skills and trait reset when starting new round in multiplayer campaign (and maybe docking as well)

**Description**
When starting a new round in a multiplayer campaign (and maybe docking as well), my character skills and trait reset. 
This started happening after playing the campaign for multiple hours and during several sessions.

**Steps To Reproduce**
We had just docked to an outpost. Then we bought a new submarine (Orca 2). We tried switching in the dockyard terminal but nothing happened. Then we read you might need to undock and dock to get it. So we did that (just undocking and docking to the same station without selecting any missions). After that we still couldn't switch. And at that point I looked at my skills and noticed they had all reset. After that I read some books to regain some skills but when we undocked for a new mission my skills reset once again.

**Version**
v0.15.21.0 using Windows 10

**Additional information**
[Saves.zip](https://github.com/Regalis11/Barotrauma/files/7738150/Saves.zip)

I guess I can just say that switching submarines is very confusing and I'm not sure if the switching issue I described is also a bug or if it's intended. I should also stress that it is only MY character out of 3 players whos skills reset. I was the host of the game. I attached my save files in case they are useful to you.

ninjanomnom:
I tested this and it seemed to be fine with nothing unexpected but it feels like I'm missing something that uses cause of death after this point.

This fixes skills repeatedly getting deducted after someone revives once mid round.

fixes #7755
fixes #7843
fixes #7791

### DIFF
--- a/Barotrauma/BarotraumaServer/ServerSource/GameSession/GameModes/MultiPlayerCampaign.cs
+++ b/Barotrauma/BarotraumaServer/ServerSource/GameSession/GameModes/MultiPlayerCampaign.cs
@@ -220,6 +220,7 @@ namespace Barotrauma
                 {
                     RespawnManager.ReduceCharacterSkills(characterInfo);
                     characterInfo.RemoveSavedStatValuesOnDeath();
+                    characterInfo.CauseOfDeath = null;
                 }
                 c.CharacterInfo = characterInfo;
                 characterData.RemoveAll(cd => cd.MatchesClient(c));
--- a/Barotrauma/BarotraumaServer/ServerSource/Networking/RespawnManager.cs
+++ b/Barotrauma/BarotraumaServer/ServerSource/Networking/RespawnManager.cs
@@ -399,6 +399,7 @@ namespace Barotrauma.Networking
                         {
                             ReduceCharacterSkills(characterInfos[i]);
                             characterInfos[i].RemoveSavedStatValuesOnDeath();
+                            characterInfos[i].CauseOfDeath = null;
                         }
                     }
                 }


chrisxlite:
Thanks for your contribution! Hope this got fixed soon. Quick question: save files don't seem to contain players' death record, so I presume all these only exist in the server codes, and before the issue is officially fixed, I can avoid it by rehosting the server after someone dies and we dock?

Regalis11:
Thank you for taking the time to look into this! Merged to our private repo in https://github.com/Regalis11/Barotrauma-development/commit/90d1fba58be02cae48e29703acdc1327ed7d9fce

@chrisxlite
You're correct, the cause of death isn't saved anywhere so it should be possible to work around the issue by restarting the server.

PR #7904, opened 2021-12-22
Fix bug (#7873) correctly check if sub is owned before switching

thomasheath1996:
Code added to ServerSource\Networking\Voting.cs (line 132-137) checked to see if players could afford subs before switching to them, even if players had already bought the sub. Added an extra condition for cases where players own the sub.


juanjp600:
We've merged this into our private repository. Thank you for your contribution!

thomasheath1996:
Will I be listed as a contributor when this code goes live?

juanjp600:
It seems like we've forgotten to update our contributors list, sorry about that! I'll go ahead and add your name now, it should go up with our next update.

thomas-heath:
Where can I find the contribution list?

PR #7945, opened 2021-12-27
Copy instead of move submarines to mod folder.

Jade-Harleyy:
There was no (?) reason to move submarines to the mods folder instead of just copying them there, and many people have complained so I have decided to fix it.
Tested and seems to work fine, at least in debug.


Tsunder:
this makes updating/editing subs really bad as people now juggle two copies of the "same" sub.

and will obviously lead to many "updates" of a sub that never work because the editor is editing their local copy instead of mod copy. etc etc.

instead i propose that we install version control on subs, complete with release branches and pull requests. this also goes well with the upcoming co-op sub editing feature.

Jade-Harleyy:
or just have the game automatically append "Local" and "Mod" to the end of the submarine name in the editor

cshingle:
When you hover over the sub in the load screen it shows you what folder it is in. My idea is to add an icon on the left side of the listing and just show one icon for sub folder and a different one for the mod folder to make it more obvious. I have not looked into the level of effort this would require.

Ek-Videogames:
The 'Mod' copy won't show up in your submarine list if the mod isn't enabled though, and since the addition of not needing to be subscribed to your own mod in order to update it - I don't see that as a problem.

juanjp600:
This pull request has been rendered obsolete by https://github.com/Regalis11/Barotrauma/commit/3974067915d2943eb3119300906aa0dcbc3d1517; the way submarines are handled is quite different now and solves the problem that this is meant to address. Moreover, there are extraneous changes in this PR that cannot be merged.

Issue #7864, opened 2021-12-19
Chained Relay Power Provided Fluctuating Wildly


gerakis:
- [x] I have searched the issue tracker to check if the issue has already been reported.

**Description**
If you daisy chain relays together to power any load, the relays from the second relay onward from the junction box have something wrong with their power supply. Power available seems to wildly jump up and down for no apparent reason. This is despite the fact that all relays can handle the loads provided. In fact if it cannot handle the load provided it seems to not do the fluctuations which is even more strange.

This isnt the load jumping around, this is the power. But the power supply is totally stable as measured from the first relay or the junction box. This happens with any load, but only if 2 or more relays are combined. This only started recently. This was in single player on the berilia and also in the editor.

**Steps To Reproduce**
Power a load with at least 3 relays that are daisy chained together. Inspect the second relay from the junction box, or any thereafter. Watch the power jump around. Ive tested it on lamps, diving suit racks, and charging docks.

**Version**
0.15.22.1 Windows 10 x64

**Additional information**
![diving suit wiring](https://user-images.githubusercontent.com/16128947/146685151-44056806-9d92-49a6-9ca9-59dfa9471d5b.gif)

Regalis11:
Duplicate of https://github.com/Regalis11/Barotrauma/issues/6471

gerakis:
I found this on #baro-sub-builder:

Relays are really fantastic for signaling, so don't discredit them, but for power transfer they are only useful in very specific scenarios.

The biggest drawback of relays is that there is a delay between calculating the load on the Power_Out grid, and requesting the power from the Power_In grid.

This can and will cause all sorts of issues if theres any "dynamic" load device, or one that fluctuates at all. What will happen is a ripple effect through your relay(s), where before the request for power can be sent to the generator, the load will have shifted, and the two numbers will never match up. This is what causes the flickering that so many people have seen in lights.
ThreadRelays1 Message ›
This thread is archived
Further, chaining relays with lights on them, as in more than one relay, would in theory be fine. But if anywhere on that grid, including at the beginning even before a relay, is a device that has a dynamic load, it will cause the lights to undervolt, then overvolt to compensate, then the ripple effect can occur and could be impossible to get rid of.


There are a couple best practices for using relays to transfer power.
First, don't mix and match devices. You can do one relay to one type of device, and be relatively fine.
[...]

PR #8149, opened 2022-02-01
Fix order of operations issue causing #7864

cshingle:
This fixes the order of operations issue that causes a long delay for power to reach relay components after the first in a chain. The delay was caused by the value of throttlePowerOutput being much higher than it should be because it was doing the subtraction last rather than first.


Regalis11:
Oh wow, I'm glad you spotted that! Thank you so much for taking the time to debug this. Merged the fix to our private repo in https://github.com/Regalis11/Barotrauma-development/commit/504e141c7d051ba1d1c6ec3cb37a22bdbe26ae7f

Issue #8168, opened 2022-02-02
Supercapacitor POWER_IN pin busted


aeondez:
- [X] I have searched the issue tracker to check if the issue has already been reported.

**Description**
POWER_IN pin on supercapacitors is functioning as generic POWER (two way power transfer). POWER_OUT pin works as expected, though.

**Steps To Reproduce**
Open up any sub, put everything on POWER_IN pin
use periscope to shoot something
dakka dakka

**Version**
Windows 10, Unstable V0.16.2.0

**Additional information**
![bug](https://user-images.githubusercontent.com/21000916/152081897-9b698a94-8a30-4bfe-8354-8abebf036851.png)

Regalis11:
Fixed in https://github.com/Regalis11/Barotrauma-development/commit/2f483297ee798d862ddc660160f9ee3ba2cf4f1d

Rokvach:
Tested, working correctly. Closing.

Issue #6824, opened 2021-10-04
[0.14.9.1] Relay component draws all available power


atik1n:
- [x] I have searched the issue tracker to check if the issue has already been reported.

**Description**
Relay component draws all available power instead of needed power.
We have this config with consume values as displayed on image:
![image](https://user-images.githubusercontent.com/14612180/135780316-8d29c4e1-d9df-41c5-9f78-351fd4ba092c.png)
If there's no relays, junction boxes will draw exact that power, what system needs, because they are connected to another "main" box and they will not fire up.

But. If there is relay between main and secondary relays, actual load on them will be 500kW, because relay draws from `power_in` and sends 500kW to the `power_out`. That will break and fire all junction boxes on image without actual overload from reactor.

Is relay actually must draw all power and transmit it away? Maybe it must work exactly as junction boxes `power` node?

**Steps To Reproduce**
Connect reactor -> junction box -> multiple relays -> corresponding junction boxes -> power consumers

**Version**
v0.14.9.1 (ReleaseWindows, branch release, revision df0099ca33)

**Additional information**
Junction box actually just connects everything connected to the `power` node and maybe that's the reason.
Maybe related issue: #6471

Rokvach:
Tested, working correctly. Closing.

Issue #5901, opened 2021-05-27
Battery power output dividing incorrectly?


BorrowedBullets:
- [ ✓ ] I have searched the issue tracker to check if the issue has already been reported.

**Description**
When hooking the output of a battery into more than one junction box, the power is divided among the boxes equally, instead of by load required.

Noticed this while play-testing the sub I'm building. The oxygen generator was not providing enough O2.
The vents were all linked to it, so it wasn't a output problem, it was a production problem.
Go to the electrical room, and start checking wires and junctions...
The junction box leading to the Oxygen Generator had a draw of 900, but a power 377.
The junction box next to it, hooked to the same battery, had a draw of 155, and a power 377.
The third junction box  hooked onto the battery had a draw of 75, but again, had a power of 377.

Basic math shows, for whatever reason, the battery has decided to take all three power draws, add them, and divide the output equally... 900+155+75=1130 & 1130/3=376.666666
To further confirm this, I tried removing the outputs of a box, but leaving the box itself connected. And sure enough, same deal.
A draw of 900, 0, & 75, results in each box now receiving 325 power.
Removing another box's outputs; Resulting in power draw of 900, 0, & 0, results in each box receiving 300 power.

**Steps To Reproduce**
Place a battery (default settings aside from changing the available charge to 2000).
Connect the battery outputs to more than one junction box. (2/3/4)
Place down a machine (I used an Oxygen Gen), connect its power to one of the boxes, and turn it on.
Swap to Test mode, and check the draw & power of any of the junction boxes.
The power provided to the junction will be equal to load, divided by the amount of boxes hooked up.
EG: A default Oxygen Gen. drawing 1000kw. Two junction boxes: Power 500. Three boxes: 333. Four: 250. Five: 200.

**Version**
"Embrace the Abyss" Patch v0.13.3.11 - Steam Build ID: 6614174 - Windows 10 Home

**Additional information**
Did the very basic troubleshooting of reloading the map, restarting Barotrauma, and loading up a different save.
It was the same result in every instance.
This doesn't _SEEM_ like it's intentional. And the wiki doesn't make mention of it...
But I do notice that none of the pre-made ships have their batteries output hooked to more than one source, so perhaps it is.

![power vs draw](https://user-images.githubusercontent.com/84906831/119816133-31d53300-bea1-11eb-9067-367cd208a9ea.PNG)

Rokvach:
Tested, working correctly. Closing.

Issue #5813, opened 2021-05-16
10 second delay for chained relays to receive power even in trivial wiring setups


JohannesMP:
# Description
In the sub editor I'm using a test scene to experiment with how various components function. After running into some odd behavior while trying to create a more complex system I've narrowed down an interesting repro case.

The setup has a full battery, feeding into a primary relay which forwards the power to a secondary relay. Both relays have a lamp attached to their power output, and all components have `Is On` enabled where applicable.

With default power values for all components, it takes about 10 seconds from when I hit 'Test' to when the lights actually receive power.

Here is a video capture of this in action:

https://user-images.githubusercontent.com/662874/118415075-18401a00-b65d-11eb-89f5-1b669adf30b9.mp4

# Steps To Reproduce
## Setup
![image](https://user-images.githubusercontent.com/662874/118414847-c4810100-b65b-11eb-926e-66b5aca23c9c.png)
- A **battery** connected to a **primary relay** that feeds into a **secondary relay**, both of which have a **lamp** attached to their power output.
- Sub file: [TEST_RelayBug3.sub.zip](https://github.com/Regalis11/Barotrauma/files/6490187/TEST_RelayBug3.sub.zip)

## Component Settings
- Battery is default
![image](https://user-images.githubusercontent.com/662874/118412538-76fe9700-b64f-11eb-82ad-418be1edea29.png)
- Relays are default, except `Is On` is enabled
![image](https://user-images.githubusercontent.com/662874/118412564-8e3d8480-b64f-11eb-945b-32f65c998c1d.png)
- Lamps are default except range and color adjusted, and `Is On` is enabled:
![image](https://user-images.githubusercontent.com/662874/118415216-01e68e00-b65e-11eb-8f1a-be965017c3d4.png)

## Wiring Setup
- The first relay receives `POWER_IN` from the battery, and sends `POWER_OUT` to the `POWER_IN` of the secondary relay.
- Both relays also send their `POWER_OUT` to the `POWER_IN` of a lamp.
![image](https://user-images.githubusercontent.com/662874/118415122-63f2c380-b65d-11eb-9b28-fe9d999c4400.png)

# Expected Behavior
When Test mode starts I would expect
- The Battery begins draining immediately.
-The two lamps light up immediately.
- The primary relay to display a 'Load' value of 10kW (2 lamps * 5kW Power Consumption each) and a 'Power' value of 10kW as well (or very quickly increasing to meet that amount)
[...]

PR #8184, opened 2022-02-04
Power Rework

Torca2001:
### Aim
The goal of this pull request is to rework the overall power system of wire connected devices to primarily fix the erratic behaviour of relays, improve the overall behaviour of the power grid and improve the flexibility of the grid to allow for future development.

### Preview
Youtube video to show the power rework in action
[Alt text](https://youtu.be/Z7PPj4S_VMY)

### Table of contents
- Reasoning
- New grid structure
- Flexibility
- Performance
- Relay Behaviour
- Battery Behaviour
- Reactor Behaviour
- Future Development
- Fixes issues
- Known Issues
- Testing

### Reasoning
The current implementation of relays is unstable and highly unpredictable as it is dependant on update order of devices.

Currently connections are treated as out going or in going, this results for example two output connections of a reactor being treated as two separate grids but the reactor's calculation of power output would account for both grids and thus push the same power to both grids, creating power and causing over-voltage/under-voltage depending on the difference in the two grids.

Since the current implementation of the power infrastructure treats power in an event styled fashion thus a device that adds a load must iterate through all of its connected devices and update their load accordingly, since this update order is arbitrary devices that require this load information and/or power information will make their calculations based off dated information or countlessly repeat calculations that can be done in one pass. 

The design of the grids being event based and having to iterate through all devices to inform them of added load or power means that the grids  resolve in O(n^2) where n is the number of connected devices while the grid can be reasonably resolved in O(n) time.

### New grid structure
To resolve the problems outlined above connections aren't specified as just input or output connections this means that the input and output pins of devices act like the current implementation of junction boxes. 
This of course brings problems that devices like engines could be powered off daisy chained lights but this is necessary to ensure connected devices work cohesively together as one grid, a simple solution would be to limit the number of connections to input pins for devices, this shouldn't conflict with a vast majority of sub designs as people typically only wire one power connection to a power input as it is sufficient to power the device.

All touching power connections regardless of input of it being on a input or output power pin will become part of the same grid. These grids are only formed if there is 2 or more devices so that unnecessary grids aren't created, and they keep track of the power, load, voltage, connections and list of power outputting connections for that update frame.
Connections will have a power priority for when they are outputting power the lowest priority has their power calculated first.
This is to ensure devices don't fight over how much power they are outputting to the grid and allows for the connected devices to react in an appropriate manner.
The order of power priority is:
1. Miscellaneous devices - i.e. junction box is zapped and pushes power to the grid 
2. Reactors - Since the they are limited on the minimum power they can output
3. Relays - Will output what it can to the grid to match its supplying grid voltage
4. Batteries/Capacitors - Will try to supply the difference to the grid to reach an equal power to load ratio of 1:1

Each grid is calculated in this following order by the corresponding functions
1. CurrPowerConsumption - Sums the load of all devices on the grid and identifies any device that will be providing power to the grid and their priority.

**Only executed for devices providing power to the grid and are executed in groups with the same priority**
    2. MinMaxPowerOut - Sums the minimum and maximum possible power out of all power providing devices connected 
        devices so they can work together and evenly distribute power output
    3. ConnPowerOut - Sums the power output for the devices in this priority group which is then carried down to lower priority 
        group calculations

**Once all power priority groups are done then voltage is calculated and this is executed for all devices on the grid**
    4. GridResolved - Perform actions on the device once the connected grid has finished updating i.e. store the voltage for next 
        tick calculations.

These calculations are only performed once per tick per grid giving an O(n) where n is the number of wire connected devices in the grid.
These grid updates are performed in [powered.cs](https://github.com/Torca2001/Barotrauma/blob/master/Barotrauma/BarotraumaShared/SharedSource/Items/Components/Power/Powered.cs) in the UpdatePower function.
Since much of the new relay design resolves itself over a series of updates and the relay has a slight delay, thus the limit to the update interval was removed meaning the UpdatePower function runs at the same rate of update (60 ups) but shouldn't hinder performance as shown later on with the significant performance improvements.

### Flexibility
Since now any power connection can act as an input or output this gives the freedom for devices to have more unique behaviours for example a junction box could be stunned and surge power onto the grid in a spike or the back emf of a pump pushing some power on the grid as the pump speed flucuates.

The Extraload variable on junction boxes were changed to allow for such behaviour so they can provide negative value to push power onto the connected grid.

The change to using functions to indicate load or power out will help in easily implementing new devices and their electrical behaviour as you only need to handle the appropriate functions and not worry about the grid itself.

Junction boxes were given power_value_out pins that give the current power on the grid
and load_value_out pins that give the current load of the grid.

### Performance
Since the current implementation runs at a complexity of O(n^2) where n is the number of devices on the grid, the execution time grew significantly with more devices. While this reworked implementation only runs with a complexity of O(n) helping save performance for other areas.

Testing of the execution time of the updatepower showed keep in mind these tests were done with the debugger version:
Power structure  |  Humpback(sub editor)     |    Profiling Level
Current               | 0.31 ms                              |    1.11 ms           
Rework               | 0.08 ms                              |     0.20 ms
Improvement     |  287.5% Faster                    |    455% Faster

These weren't limited to just these tests, across the board on various workshop subs and other vanilla subs the difference in execution time was similar. A workshop sub that I tested that had a large power grid the [[VH] The Iroh ](https://steamcommunity.com/sharedfiles/filedetails/?id=2602920829) in the sub editor took 1.15 ms similair to the profiling level while the reworked time was only 0.18 ms.

### Relay Behaviour
The relays were mostly fixed but this implementation isn't perfect but does achieve a predictable behaviour and should be good for most use cases in game.

Relays were given an internal buffer that is double the size of their maxpower, this is to allow the input side and output side of the relay to be isolated and to be free to be calculated separated of one another. 
The load of the relay is determined by this internal buffer to fill it, this allows the relay to stabilise and find an equilibrium of load when in loop configurations.

The relay will try to mimic the voltage of its supplying side, this works fairly well but has a delay.
Since the voltage for a grid can't be known before it is solved the relay must use the previously stored voltage of its supplying grid, this means that there will always be at least a one update delay between the supplying and outputting grid. This becomes less noticeable at higher update rates for the power grid hence why I encourage the update rate to be at the max to minimise this propagation delay as it becomes more noticeable in large chains of relays.

This pseudo recreation of the voltage also means that relays have to have a minimum ~1~ 0.001 kw draw at all times, this is to allow a chain of relays to know the downstream voltage and correctly ramp up their output and thus increase their load on the grid.
~While this doesn't stick to the original relays pulling no load while off it was a reasonable trade off to get appropriate behaviour from the relays. ~ The low draw will have negligible impact on existing sub design taking 5.5 real world days to drain a default super capacitor, but may cause confusion as junction boxes could light up while showing 0 load and 0 power when there is actually 0.001 kw, this can be resolved by having a minimum power check for the lights on junction boxes and can be changed if needed.

A lot of edge cases were found in testing and thus a lot of the behaviour of the relay is designed to allow the relay to stabilise and not spiral into an oscillating scenario.
Hence the load of the relay is very slightly smoothed as it helped in loop configurations.
Overload scenarios will burn off some power by using the previous voltage for input calculations otherwise the buffer will become unstable.
During  2 voltage overloading the power in is limited to prevent an uncontrolled buffer oscillation.
Relays will output a slightly lower voltage taking off 0.0005v if its above 1v limiting it to 1v to help prevent ghost overvolting when looped back on a reactor powered grid.

Overall much of the relay behaviour is designed to settle itself while still allowing for dynamic loads to function reasonably well. 
Relays will also work together based on their output capabilities to help power a grid, this mean that if you have 3 equal relays they would each output 1/3 of the grid demand, if one of those relays suffers under voltage and can't output the required power then the other relays will accommodate for this and will adjust their output to reach a 1v grid.

Relays buffers start charged hence low power devices when being powered by relays in a loop can seemingly make power but in reality they relays buffers are depleting and are adjusting to the loop configuration slowly but if there is a large power draw then this isn't noticeable as the buffers are quickly drained by the load.

Relays were given an power_value_out pin that represents how much power the **relay** is outputting not the grid power.
They also have a load_value_out pin that outputs the output side grid load

 
### Battery Behaviour
Batteries input act the same as the current implementation respecting the recharge rate set 

Batteries output is set to be a lower priority than reactors and relays, thus batteries will only supply power so the grid meets a 1:1 power to load ratio.
Batteries will evenly share the load based on their output capabilities, thus if a battery and relay are wired together on the output the capacitor would take a majority of the load as it has a significantly higher output capability compared to the battery.

Batteries also have the power_value_out pin that represents how much power the battery is outputting.
The load_value_out pin is also present showing how much load the outputting grid is.

### Reactor Behaviour
Reactors being the main power provider have the lowest priority besides the default.
This is to ensure that the reactor will output its max power and climb in output to meet the entire grid demand since the other devices like relays and batteries are reactive and thus can quickly change output while a reactor can't.

Reactors will adjust their displayed load when there are multiple reactors on the grid so that all reactors will aim to work at the same turbine output, hence evenly distributing their load based on output capabilities of  all the reactors on the grid.
If a reactor is being manually controlled and outputting more than it should in a multi reactor setup then all other reactors will adjust their load as to not over output onto the grid but they will try to creep their loads to meet the ideal even target.

Also due to the change of resolving the load and power in the same moment logic controlled reactors will have a slight lag as they try to chase a load that just happened, this results in a slight nerf hence all current logic reactors will still work well but have this load chasing problem that means simple 2 component reactors wont be as perfect and to get a reactor to perfectly match the grid the logic will have to predict the load and ramp the reactor before it happens making perfect logic reactors significantly harder to achieve, which may be preferred as 2 component are becoming more prevalent.

### Future Development
The structure of the system is intended to make it easier for changes of device behaviours and the addition of new devices to the game.

Improvements to how grid propagation works, current implementation is fairly quick but for instance docking hatches could instead keep the two grids isolated to save time having to determine the two grids again when disconnected.

Isolation of the grids allows for the possibly to perform grid calculations concurrently with multi threading to further improve calculation times in very large scale grids.

New devices such as surge protectors could be added, if a device had a very low priority then after all devices have outputted power this device could put negative power to the calculation to remove the overload and instead "absorb" the power into itself taking damage.

### Fixes issues
Current issues that this would fix
- #7864 
- #6824
- #5901
- #5318
- #5813 
- #8168 
- #3566
- #3238
- #2831
- #2672
- #2217
- #6824

### Known Issues
Of course this implementation of the power grid isn't perfect and has a few issues.
Relays in a large enough and complex configuration involving many loops will be bound to cause issues.
Relays looped back on themselves and batteries will create what I call phantom power, essential the device is providing power back to itself, this doesn't affect the grid operation is merely just increases the numbers on all the readings but doesn't actually create problems, no power is created out of thing air its just stuck in a loop.

### Testing
All of testing was run on my own computer and in a single player only environment, now I believe the implementation should just work for multi player but this is untested and thus is up in the air.

Many edge cases from testing I found from playing around so its very possible that I had missed some.

If there are any problems with merging or weird scenarios I am happy to help to get the pull request accepted.


cshingle:
I have experimented with your solution and here are the pros and cons in my opinion.
## Pros
* This fixes a lot of issues and odd behaviors
* The performance boost will be really nice on large vessels and games with mods.
* Daisy chaining batteries is really nice.
* You're no longer limited in the number of batteries you can put in your sub.
* I have been doing some experimenting with relays to try to fix the issue where when overloaded their load is taken into consideration by the source grid, but they don't use the power causing overload issues. This solution resolves that.
## Cons
* Relays with a large value can be used as batteries/capacitors.
* Relays have to "charge up" on startup. (I guess this is the punishment for having relays with large values)
* **I don't like that equipment can be daisy chained. I think it makes junction boxes obsolete. In my opinion this is the largest argument against this solution.** 
* I have mixed feelings about the relay input pins working like a junction box. 
* **I don't like that the relay output pins can work like a junction box. You can basically wire two relays to function as 4 junction boxes.**
* **Relays will passively drain batteries even when off. This will break some shuttle designs. I believe you can reduce/fix this by lowering the passive draw for calculating voltage to a very small value such as 0.001.**

The current stable build lets you daisy chain power consuming devices which transmits load back to the source, but does not provide power. I actually think that is worse then not letting the user connect the pins in the first place since they are lead to believe it will work.

I would look into the only allowing power consuming devices "power_in pins" to connect to power producing devices "power_output pins". Obviously junction boxes would need to be classified as both. This would allow a small amount of oddities like being able to daisy chain junction box -> relay in -> junction box, but that is a fair improvement from what we have today.

Torca2001:
> I have experimented with your solution and here are the pros and cons in my opinion.
> 
> ## Pros
> * This fixes a lot of issues and odd behaviors
> * The performance boost will be really nice on large vessels and games with mods.
> * Daisy chaining batteries is really nice.
> * You're no longer limited in the number of batteries you can put in your sub.
> * I have been doing some experimenting with relays to try to fix the issue where when overloaded their load is taken into consideration by the source grid, but they don't use the power causing overload issues. This solution resolves that.
> 
> ## Cons
> * ~Relays with a large value can be used as batteries/capacitors.~ I could not find a way to create a test to show this. I no longer think this is possible.
> * ~Relays have to "charge up" on startup. (I guess this is the punishment for having relays with large values)~
> * **I don't like that equipment can be daisy chained. I think it makes junction boxes obsolete. In my opinion this is the largest argument against this solution.**
> * I have mixed feelings about the relay input pins working like a junction box.
> * **I don't like that the relay output pins can work like a junction box. You can basically wire two relays to function as 4 junction boxes without even turning them on.**
> * **~Relays will passively drain batteries even when off. This will break some shuttle designs.~** Active relays with a downstream grid have a minimum drain of 1kW. I personally feel this is fine as a game mechanic as real relays have a small minimum drain.
> 
> Note: The current stable build lets you daisy chain power consuming devices which transmits load back to the source, but does not provide power. I actually think that is worse then not letting the user connect the pins in the first place since they are lead to believe it will work.
> 
> I would look into the only allowing power consuming devices "power_in pins" to connect to power producing devices "power_output pins". Obviously junction boxes would need to be classified as both. This would allow a small amount of oddities like being able to daisy chain junction box -> relay in -> junction box, but that is a fair improvement from what we have today.

I agree that there will be issues with the relays passive draw, 1kw I thought was a valid mechanic but in reality 0.001 kw works just as well but I thought it may confuse people as they see junction boxes lit up but 0 load and power plus it wouldn't hurt mechanically wise 

I'm not a fan of the daisy chaining from power connections either, limiting connections to power_in to power_out and either power_in or power_out can connect to a junction box would avoid the most blatant abuse but even with these rules a configuration such as two relays outputting to a junction box would mean that both output connections of the relays are the same grid hence giving 8 more connections but at the same time this could be seen as a valid configuration if both the relays are powered providing power to the junction box. But atleast with these rules it would force any kind of grid abuse to atleast have one junction box in the mix so that grid wouldn't be completely immune to over/undervolting.

Maybe the best solution would be to limit relay out connections or some other incentive to using junction boxes.

cshingle:
Take a look at the changes I suggested. I think the first one has no downsides and avoids a potential bug with custom relays. 
 
> I agree that there will be issues with the relays passive draw, 1kw I thought was a valid mechanic but in reality 0.001 kw works just as well but I thought it may confuse people as they see junction boxes lit up but 0 load and power plus it wouldn't hurt mechanically wise

My main concern is the 1kw load when the relay is disabled/not connected to a downstream grid. I don't have an issue with the 1kw draw when it is connected/enabled. The second suggested change does just that.
If we want to completely remove it we could just reduce it to 0.001kw like I initially suggested, then update the code for the junction boxes to not light up unless the load is >= 1kw. I didn't find an obvious way to do this in the code, but I am sure we can figure it out if we want to.

> I'm not a fan of the daisy chaining from power connections either, limiting connections to power_in to power_out and either power_in or power_out can connect to a junction box would avoid the most blatant abuse but even with these rules a configuration such as two relays outputting to a junction box would mean that both output connections of the relays are the same grid hence giving 8 more connections but at the same time this could be seen as a valid configuration if both the relays are powered providing power to the junction box. But atleast with these rules it would force any kind of grid abuse to at least have one junction box in the mix so that grid wouldn't be completely immune to over/undervolting.
> 
> Maybe the best solution would be to limit relay out connections or some other incentive to using junction boxes.

I completely agree with your assessment.

- The best solution I have come up with is allow a few items to be daisy chained such as batteries, lamps and oxygen tank racks. These items are the largest reason for relay spam. Allowing them to be chained would remove most of the incentive for relay spam as well as simplify the power grid calculations on most subs.
- Another idea that I have is to leverage the 1kw power drain mechanic you already created but make it consistent for any relay that has power on the power_out side.

Reducing the relay out connections is an option as well. Is there a way to do it that does not break current subs?

Some feedback from @Regalis11 would be great since it is his opinion that matters.

Torca2001:
@cshingle  Following your suggestion I feel it to be best to lower the passive relay draw as it would eliminate the conflict with the original relay design as this passive draw of 0.001 kw will allow for the wanted relay behaviour to still occur while having negligible impact on battery grids, for reference it would take 5.5 real world days to drain a standard super capacitor.

Regalis11:
Wow, this looks very promising. Thank you so much for putting this much work into fixing the mess that is the current power logic. :D

I will take a more in-depth look at this and do some testing once my hands are a little less full with the next update, but on a cursory look this seems like a much more robust system that'd take care of most of the issues in the current system.

> I don't like that equipment can be daisy chained. I think it makes junction boxes obsolete. In my opinion this is the largest argument against this solution.

I think this is something that probably should be changed though. I'm not entirely sure what would be the best way to deal with daisy chaining relay outputs (limiting the number of wires you can connect to it could perhaps be something to consider), but imo it shouldn't be possible to daisy-chain devices, because as said, it makes junction boxes obsolete. Is there a reason the grid needs to propagate through the devices' power connections? I'm probably overlooking something, but to me a simple way to prevent the daisy-chaining would be to prevent propagating through the power connections of anything else but PowerTransfer components (junction boxes and relays).

Torca2001:
> Wow, this looks very promising. Thank you so much for putting this much work into fixing the mess that is the current power logic. :D
Thanks its been bugging me for months and I had been coming up with all different methods on how to resolve the grid, eventually getting bored one day and decided to spend 3 weeks coding it, unsurprisingly a majority of the time was spent tweaking relays.

> I think this is something that probably should be changed though. I'm not entirely sure what would be the best way to deal with daisy chaining relay outputs (limiting the number of wires you can connect to it could perhaps be something to consider), but imo it shouldn't be possible to daisy-chain devices, because as said, it makes junction boxes obsolete. Is there a reason the grid needs to propagate through the devices' power connections? I'm probably overlooking something, but to me a simple way to prevent the daisy-chaining would be to prevent propagating through the power connections of anything else but PowerTransfer components (junction boxes and relays).
I wish there was a simple way to resolve this but only propagating through connections of PowerTransfer components arise its own issues. A big one for the reason that the power input of a device could only be apart of one grid is how should the device's load be divided among its input connections? The current stable build just applies its load to every grid its connected to this means that both grids will provide the necessary power for the device so the device would essentially receive more power than it requested significantly, this isn't a problem in the current build as the device caps the voltage to 1 so it won't go over but if you wanted a device that was affected by overvolting there would be no accurate way in determining whats overvolting and whats just incorrect power coming in. 

Now I considered that maybe the device could evenly distribute its loads across its connections so if it had 3 connections each get 1/3 but in this configuration if one of those grids couldn't provide power then the device could only at most receive 2/3 of its power demand even if one of those grids could provide the missing power hence it wouldn't work.

Another possibility would be to distribute the device's load based on the max power output of those grids, but for a grid to determine its max power output it must know its load first, this is especially important with relays so they can perform their under volting and over volting behaviour. Thus that results in a catch 22 since we can't know the load without knowing the power and vice versa. 

Not to forget we would most likely need to apply this same behaviour to the power output of devices as you intend the power to push out of the output pin of the device, but then again the question of how much power coming out of the device should go to each grid and again I could only come up with the same solutions that all wouldn't work.

Not to forget that implementing these would most likely require a majority of the code to be rewritten as its a drastic change in how the grid calculations would be done and require that grids be updated in a certain order due to the fact that multiple grids would be connected together and have to be solved together which is likely to cause unwanted behaviour in itself.

> EDIT: A small note, why's the minMaxPower a 3-dimensional vector? What does the 3rd dimension do?
The minMaxPower is really intended to act as a way for same type devices to communicate together and coordinate they power output, thus the first two variables are used for minimum power out and maximum power output.
The third is a spare variable that devices may need sometimes such as in reactors, they will use the third value to store the max power output rating of all reactors and thus this allows for the reactors to adjust their loads accordingly and match all their turbine rates together so no reactor is fighting against each other.
It is easier to implement a third variable in the minMaxPower since I want the third value to be summed just like the minimum and maximum and thus combining it into one data struct is easier. Honestly it could just be replaced with a data struct to make its use clearer.

Torca2001:
> Ah, you're right, that didn't occur to me. I wonder if it'd work though if we just ignored connections from a device's power input to another device's power input (and by device I mean anything else than a PowerTransfer component)? That way the device would still only be part of one grid.

Wouldn't that be no different to preventing the wiring connection to be made in the first place? It would make it less confusing to players why the device isn't working.

This also wouldn't prevent configurations such as two relays output connected to one lamp, and then one of the relay output connections could be connected to a junction box so essentially giving 7 connections.

cshingle:
> Wouldn't that be no different to preventing the wiring connection to be made in the first place? It would make it less confusing to players why the device isn't working.

I am very much a fan of not letting players connect wires in a configuration that is not appropriate. It "teaches" the correct way of doing things as well as improving performance by moving the checks to the wire connect loop (rare) rather than the power update loop (multiple times a second).

> This also wouldn't prevent configurations such as two relays output connected to one lamp, and then one of the relay output connections could be connected to a junction box so essentially giving 7 connections.

What if we only create the relay output grid when the relay input grid is not null (and maybe isn't the same grid as the output grid)? The players could still find some convoluted way around this, but we don't need to make it impossible. We just need to make it worse than chaining relays like they do now.

The bug that caused the extended delay on chained relays has been fixed, so chaining relays will be even more viable for mostly static loads than in the current stable.

It was the bad behavior or relays that drew me in as well.

Regalis11:
> Wouldn't that be no different to preventing the wiring connection to be made in the first place? It would make it less confusing to players why the device isn't working.

That's a good point. Although, I haven't heard of players being confused by that even though that kind of connections have never worked in the game. To me it seems pretty intuitive that a device needs to be connected to a device that can provide power, be it a battery, relay or a junction box.

> This also wouldn't prevent configurations such as two relays output connected to one lamp, and then one of the relay output connections could be connected to a junction box so essentially giving 7 connections.

I mean just ignoring connections between the power inputs of two non-powertransfer items. It wouldn't affect that kind of a circuit, because all of the connections are between powertransfer components or a powertransfer component and a lamp.

> The bug that caused the extended delay on chained relays has been fixed, so chaining relays will be even more viable for mostly static loads than in the current stable.

I actually reverted that change: as I mentioned [here](https://github.com/Regalis11/Barotrauma/issues/7864), the order of operations that you changed was actually intentional. The "throttling" was supposed to decrease at a speed relative to the relay's maximum output, but after the change it dropped to 0 instantly when there's no need for the throttling, which caused other kinds of issues (namely, even worse fluctuations in circuits where the load is much lower than the maximum output of the relay).

Torca2001:
I've changed how grids propagate through connections, they should now only consider connections between power_out to power_in and to junction boxes to be valid. For example daisy chaining the power in pin of lamps won't pass power on anymore, nor will daisy chaining relay outputs work anymore.

cshingle:
> I've changed how grids propagate through connections, they should now only consider connections between power_out to power_in and to junction boxes to be valid. For example daisy chaining the power in pin of lamps won't pass power on anymore, nor will daisy chaining relay outputs work anymore.

Nice work. The electrical behaves close enough to the current release that most people are not likely notice the difference other than a bunch of issues being resolved.

Regalis11:
Awesome! I think I'll merge this to a feature branch in our private repo now and start doing some more internal testing on this.

Rokvach:
I haven't noticed any major issues with this, closing in favor of potential smaller issues, closing.

PR #8195, opened 2022-02-05
Add a low oxygen pin

cshingle:

**Description**
This is a feature request
Oxygen generators only put out a percentage of oxygen in the current hull. This makes creating circuits that toggle on and off the oxygen generator a complete abomination that likely impacts performance. This seems to be getting more popular (maybe I am just noticing people do it more).
See;
https://steamcommunity.com/sharedfiles/filedetails/?id=2732232801
https://steamcommunity.com/sharedfiles/filedetails/?id=2732422820
https://steamcommunity.com/sharedfiles/filedetails/?id=2419010949

This is easily resolved by adding a low_oxygen pin to the oxygen generator similar to how the Water Detector has a high_pressure pin. Adding this pin would also make creating an alarm for low oxygen work in a similar manor to how fire and flood alarms work in vanilla subs.

**Steps To Reproduce**
Attempt to create a circuit that will toggle on and off the oxygen generator if the oxygen is low in any hull without using a ridiculous number of components.

**Version**
v0.15.23.0 (Windows/Linux).
I play on Linux, but develop/debug on windows because I use Visual Studio.

**Additional information**

Add an low_oxygen pin to the Oxygen Detector. 
This change requires xml changes as well. I can't add those to the PR because they are not in this repo.

This needs to be added to the oxygendetector ConnectionPanel in the signalitems.xml
`<output name="low_oxygen" displayname="connection.lowoxygenout" />`

This needs to be added to the EnglishVanilla.xml. I don't know other languages so I am no help with translation.
`<connection.lowoxygenout>low_oxygen</connection.lowoxygenout>`

### DIFF
--- a/Barotrauma/BarotraumaShared/SharedSource/Items/Components/Signal/OxygenDetector.cs
+++ b/Barotrauma/BarotraumaShared/SharedSource/Items/Components/Signal/OxygenDetector.cs
@@ -4,6 +4,8 @@ namespace Barotrauma.Items.Components
 {
     class OxygenDetector : ItemComponent
     {
+        public const int LowOxygenPercentage = 35;
+
         public OxygenDetector(Item item, XElement element)
             : base (item, element)
         {
@@ -14,7 +16,9 @@ namespace Barotrauma.Items.Components
         {
             if (item.CurrentHull == null) return;
 
-            item.SendSignal(((int)item.CurrentHull.OxygenPercentage).ToString(), "signal_out");            
+            int oxygenPercentage = (int)item.CurrentHull.OxygenPercentage;
+            item.SendSignal((oxygenPercentage).ToString(), "signal_out");
+            item.SendSignal((oxygenPercentage <= LowOxygenPercentage ? "1" : "0"), "low_oxygen");
         }
 
     }


Regalis11:
Thank you for the PR! We'll wait for the release build for the next update to be ready before merging this, because the new text won't have time to make it into the next update.

JoneKone:
Does it actually impact perf? .. I doubt it.

cshingle:
> Does it actually impact perf? .. I doubt it.

@JoneKone This is a very small change. It should have less performance impact than adding a single logic component.

JoneKone:
@cshingle I think you missed the context on witch my sentence was based on. (this change will not save any perf)

In principal to respect ppl who actually take time to make complex systems.... And to be honest then being wiped under the carpet for conveniences sake... Some ship builders are not happy about reducing complexity and adding more automation... Not talking on behalf of my self... But why add all components that can build complex in game logistics only to then replace them with internal game logistics.

I'm aware this is a broader subject... I personally believe if you make a challenge don't let other ppl influence by complaining about it.

Regalis11:
@JoneKone In general I agree that we shouldn't streamline things too much to leave room from creativity and complexity when it comes to custom circuits, but I don't think adding this specific output is something that'd hurt. Checking whether the value is above 35 or whatever with a "greater than" component doesn't take much creativity nor does it add much complexity to the circuits, and this change would just allow you to skip that check. It won't make it any less trivial to build the kind of sophisticated automatic oxygen generator systems mentioned in the PR.

But I'd say performance is not an argument for adding this output: the performance impact of skipping that one "greater than" component is negligible at best, and might even be negated by the (also negligible) performance hit caused by having all the oxygen detectors send 2 outputs instead of just 1.

Regalis11:
Merged to our private repo in https://github.com/Regalis11/Barotrauma-development/commit/1b5f1f766bc4bcea35461d2748549b747c2ee478

Tsunder:
35% o2 for low o2 warning won't/shouldn't be used in any of the circuits above.

The only use for this is a shittier fire alarm or sky artifact, as by the time o2 hits 35% people will be asphyxiating/unconscious.

as for those circuits, they're for tons of o2 detectors, which means tons of rooms and vents, large subs. and also optional and provide little tangible benefit. It's much easier and practical to leave o2 on all the time. Making a circuit that controls it is a nice little experiment, but not something I would generally recommend implementing into a sub as an sub editor nor as a player

PR #8231, opened 2022-02-11
option for remove turrets hitbox outside of hull.

war3scv:
so enemy damage wall behind them properly and we don't get bunch of corpses stuck around.


war3scv:
first time my bad:)

Issue #8388, opened 2022-02-26
Reactor shutdown incorrect load


Torca2001:
- [x] I have searched the issue tracker to check if the issue has already been reported.

**Description**
If a reactor is outputting power to a grid and is shut down, the load reading of the reactor jumps down to 0 and slowly comes back to the true load.

**Steps To Reproduce**
Power up a reactor and wait for it to stabilise to the grid's load.
Shut down the reactor and observe the sudden load drop.

**Version**
V0.17.0.0 Unstable, Windows

**Additional information**
It is most likely related to the multi reactor load behaviour being incorrect as during shutdown the calculations accidently used the multi reactor calculations for the load.
It was previously working in my PR #8184

Regalis11:
Fixed in https://github.com/Regalis11/Barotrauma-development/commit/5f65b9e6fde47c41cc135c487fde44db370c2712

Rokvach:
Tested, working correctly. Closing.

Issue #8387, opened 2022-02-26
Multi Reactor incorrect load


Torca2001:
- [x] I have searched the issue tracker to check if the issue has already been reported.

**Description**
Reactor will display the wrong load during shutdown spiking down to zero. In multi reactor configurations each reactor would report significantly lower load than the true load and the sum of all the reactor loads was closer to half the true load.

**Steps To Reproduce**
Loading in the humpback and attaching a second reactor to the main grid and turn both reactors on and put them on auto.

**Version**
V0.17.0.0 Unstable, Windows

**Additional information**
It was previously working in my PR #8184

Regalis11:
Fixed in Fixed in https://github.com/Regalis11/Barotrauma-development/commit/5f65b9e6fde47c41cc135c487fde44db370c2712

Rokvach:
Tested, working correctly. Closing.

Issue #8386, opened 2022-02-26
No output JB Load & Power value out


Torca2001:
- [x] I have searched the issue tracker to check if the issue has already been reported.

**Description**
Junction boxes power_value_out and load_value_out only output 0 regardless of the load and power on the grid shown by clicking on the junction box.

**Steps To Reproduce**
Load up the humpback, attach a text display each for the power_value_out and load_value_out pin of a junction box next to the nav terminal.
Observe the 0's

**Version**
V0.17.0.0 Unstable, Windows

Rokvach:
Tested, working correctly. Closing.

PR #8392, opened 2022-02-26
Unstable Power Rework Fixes

Torca2001:
Fixes bugs relating to the power rework caused by refactoring. Most of the issues were very minor.
Also fixed batteries causing flickering which was actually from the original pr.

### Fixed
- Multi-reactor equations giving incorrect load
- Battery's equations outputting incorrectly and flickering if output power exceeded input power
- Broken devices being repaired not updating grid connections
- Junction box power_value_out and load_value_out outputting incorrect values
- Battery's power_value_out being negative
- Battery's dividing by zero causing nan errors
- Powerpriority of connections being incorrectly assigned

### Related issues
- #8386 
- #8387 
- #8388

### DIFF
--- a/Barotrauma/BarotraumaShared/SharedSource/Items/Components/Machines/Reactor.cs
+++ b/Barotrauma/BarotraumaShared/SharedSource/Items/Components/Machines/Reactor.cs
@@ -418,7 +418,7 @@ namespace Barotrauma.Items.Components
             {
                 float idealLoad = MaxPowerOutput / minMaxPower.ReactorMaxOutput * loadLeft;
                 float loadAdjust = MathHelper.Clamp((ratio - 0.5f) * 25 + idealLoad - (turbineOutput / 100 * MaxPowerOutput), -MaxPowerOutput / 100, MaxPowerOutput / 100);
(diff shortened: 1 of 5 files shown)


Regalis11:
Thank you for the contribution again! That was fast, I didn't even have time to look into these issues yet! :)

A couple of questions about these changes:

- Where exactly were batteries dividing by zero, and which of these changes are meant to address that?

- Why's it necessary to flag the changed connections when a broken item is repaired, when that connection is not part of any grid?

Torca2001:
> - Where exactly were batteries dividing by zero, and which of these changes are meant to address that?

During testing the power test sub started to throw an assertion error since during startup as I made changes to batteries to use previous charge for calculations. To prevent a race condition and improve pass through behaviour as it caused flickering.
This is on line 268 of powercontainer. It just checks to make sure the max power is not 0, otherwise it will perform a divide by zero error in the following lines of the if check.


> - Why's it necessary to flag the changed connections when a broken item is repaired, when that connection is not part of any grid?

Grids are only formed for devices which can pass power to each other. So a junction box breaking can split a grid, thus the current grid is invalid and is removed. So the junction box then isn't apart of any grid since its broken.

Regalis11:
Ah, makes sense, thank you for the clarification! Merged these changes to our private repo now, closing

PR #8443, opened 2022-03-03
remove waitforkeyhit check

Tsunder:
i'm sure this will have no unforeseen consequences

### DIFF
--- a/Barotrauma/BarotraumaClient/ClientSource/GameMain.cs
+++ b/Barotrauma/BarotraumaClient/ClientSource/GameMain.cs
@@ -774,8 +774,7 @@ namespace Barotrauma
                         SoundPlayer.Update((float)Timing.Step);
                     }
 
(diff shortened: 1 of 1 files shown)


cshingle:
It would be nice to not have to push the anykey to go to the main menu every time the game starts.

Regalis11:
What's the reasoning behind this change? I think it's important to wait for the input so a round doesn't start while your afk if you go and do something else while it loads. Even though it isn't that important in the main menu, I think it's good that all loading screens behave consistently.

And what's the purpose of the waitKeyHit parameter in the ShowLoading menu and the waitForKeyHit field in GameMain now if they're not used for anything?

Jade-Harleyy:
I think a better way to implement this would be to make it toggleable in the settings menu, so players can choose whether they want to have to press a key to close loading screens or not.

Tsunder:
The primary reasoning was just to remove the initial game load key wait (that goes to main screen)

I forgot that wait for key is used elsewhere (i.e., single player mode)

PR #8541, opened 2022-03-17
Unstable Power fixes

Torca2001:
### Fixes

- Power incorrectly displayed in reactor when unwired
- Power being incorrectly stored in their class (Cause of above issue)
- Some docking connections being considered invalid (Relay output to docking port)
- Removed Voltage property from setting voltage of the grid. Caused confusion and in many cases is not needed or wanted.

I think I got all the power related bug, hopefully this is the last of it.


Regalis11:
> Power incorrectly displayed in reactor when unwired

I don't think this works correctly: now devices that are powered by batteries, but that can output power, would never power down.

Torca2001:
> > Power incorrectly displayed in reactor when unwired
> 
> I don't think this works correctly: now devices that are powered by batteries, but that can output power, would never power down.

What devices can output power?

Regalis11:
Merged to out private repo with some minor adjustments

Issue #6000, opened 2021-06-08
LOS shader gives inconsistent and unsightly results 


Working-Joe:
- [x] I have searched the issue tracker to check if the issue has already been reported.

**Description**
To put it blunt, the current LOS shader fails in most circumstances where two colliders overlap. 

To give an example, take this back of a submarine I made:
<img src="https://user-images.githubusercontent.com/50582661/121258652-743e3d00-c8af-11eb-8419-22b7e4d87d0c.png" width="500">
This looks fine in the editor. But once you open it in game, it looks like this: 
<img src="https://user-images.githubusercontent.com/50582661/121258915-c67f5e00-c8af-11eb-9296-8fafd8e86b8b.png" width="500">
Which looks rather unappealing. To make things worse all the triangles move with the character just to drag that extra bit of attention to it. 

It is only very rarely possible to mitigate these problems by very precisely aligning the colliders, but with most shell combinations it is outright impossible. Removing these artefacts usually requires redesigning parts of a sub over and over until you're lucky enough to have somewhat acceptable spikes.

Note that I assume this issue is known given how long it has been in the game. I'm reporting it anyways because I could find no open issues about it while it is one of the biggest eye-sores in sub building. This is also one of these topics that keeps coming up again and again in the sub-building channel on discord. 

As daunting of a task it may be to rewrite the LOS shader or even some part of the rendering code to fix this, I think it would be worth it given that it affects all submarines including the vanilla subs. It may also affect gameplay somewhat since there will be less spots where you can incorrectly see outside of the submarine, which currently kind of breaks the immersion of being inside somewhat (at least for me :).

**Steps To Reproduce**
I've attached the specific sub to which the images belong [here](https://github.com/Regalis11/Barotrauma/files/6619639/sub.concept.A.zip).

Working-Joe:
Because I didn't just want to file a complaint I've been looking a bit into ways in which the shader could be improved, and I think the following approach may achieve results very similar to the existing LOS shader but without the weird spikes that the current shader has:

![image](https://user-images.githubusercontent.com/50582661/121345524-3ed43680-c925-11eb-8feb-89c9b2eb6ad1.png)

The end result will then look something like this (note that I did this mockup in paint.netso some things, especially the zoom blur, are not ideal)

![image](https://user-images.githubusercontent.com/50582661/121346091-f49f8500-c925-11eb-8964-94042ba19876.png)

[...]

PR #8570, opened 2022-03-22
Added Raycast LoS setting

Working-Joe:
# Added Raycast LoS setting 
This PR aims to address some problems with the current LoS rendering techniques. (See also #6000)

This image shows a comparison of the old and new technique: 

scene_comp

The main differences are the weird corners on top of geometry in the old version, and other quirks like how the top of the door does not cast a LoS shadow. By drawing the actual sprites to an occlusion map rather than using the colliders, a more accurate LoS is drawn. 

While the shader works and the settings are integrated, a few things still need to be done to implement it properly (see "Work left"). I hope it can be of use regardless. Let me know if you have any questions.

## Working
To get the new effect, I've implemented a series of shaders and RenderTargets. The working can be summed up with the following figure:

shader

These shaders essentially do a raycast from the viewtarget outwards, sampling an occlusion map to calculate the distance at which a hit occurs. These distances are written to a RenderTarget which is then used as a LUT to test if a pixel is behind this occluders. You can find a shadertoy of the shader here as well: https://www.shadertoy.com/view/NdlfD8

The penumbra LUT is precalculated and saved as PNG. It's included in the attached zip. You can copy it from the shadertoy by uncommenting line 74 of the Image shader and pressing right-click -> save as on the image. The image will have to be flipped vertically to work with the shader in Baro.

## Performance
I have done some profiling using RenderDoc to determine the load of these LoS shaders. For debugging I've used the following scene:

scene

The results in RenderDoc are as follows:

benchmark

Take these with a grain of salt as drawing times vary between frames. 

It's not entirely free as can be seen, compared to the previous LoS it's a few times slower. However, it's still a fraction of the total lighting drawing time, and it scales well with the amount of sprites (as only drawing the occlusion map is slowed by that). 

You can also improve performance a good bit with the settings, especially the LoS texture size helps. 

## Additional changes
I also made the "castshadow" property of structures editable and serialised. This is useful as some front shell parts have it disabled as it looked bad with the normal LoS settings. It's also nice to have in general, as it gives sub builders more control over what casts shadows which can be convenient for decorations.
On top of that, the ViewTarget is now set to the head of the character if it's available. I only did this for the Raycast LoS code as not to interfere with the existing system, though it should be easy to make it used for both.

## Work left
Though this is mostly complete, there are some things left to do. However, I'm kind of done experimenting with this so this would be up to you. :)
- The shaders have to be rewritten for OpenGL still, as I couldn't easily test this on OpenGL. I don't know if you usually convert the shaders automatically, but this might be possible.
- I did not yet make the new LoS mode work with the server settings. Somehow the server should have a setting that can require the clients to use the LoS ones of their choice.
- The settings text still needs to be localised.
- The settings might have to be moved around a bit with the changes on Unstable.
- Though I think I've chosen sensible values, you may want to change the default raycastsettings.

## known problems
There's a few minor trade-offs in how things work, which I've listed here
- _Some of the shaders use a different compile target, namely ps_4_0_level_9_3 rather than ps_4_0_level_9_1. This might cause some compatibility problems_
- There is some jitter in the shadows which has been hard to pin down. Likely has something to do with the movement of the character and the submarine. Aside from increasing the number of rays in the settings, it may be possible to mitigate this in two ways. Firstly, the viewtarget could be smoothed. Secondly, it may be possible to do some temporal blurring by adding the new occlusion to a faded out old occlusion each frame. This would require a swapbuffer (similar to how I implemented it for raycasts) for the whole screen and an extra shader pass.
- Because the amount of rays you can cast is limited, if you look along the edge of geometry there may be a slither of unoccluded parts. Partially covered up by penumbra
- Because the amount of ray steps is limited, very thin geometry does not occlude. This is rather rare
- Because of how the InDist is calculated, in rare cases it is possible to see other occluders behind thin occluders if they are still within the InDist. Parts inbetween are still occluded properly.

## Extra info
The zip added to this commit also contains a .sub file that I've used to test the LoS.

[LoSFix.zip](https://github.com/Regalis11/Barotrauma/files/8324295/LoSFix.zip)


JoneKone:
Wau

Regalis11:
Finally got around to taking a closer look at this PR. This is very impressive, thank you for taking the time to implement all this!

The rest of the remaining work and issues seem relatively easy to address, but the jitter problem is a little concerning. In my opinion it's so noticeable and distracting that we need to do something about it before this technique can be used.

I did some debugging to try and figure out the cause to the jitter, but unfortunately didn't make much progress with that. I initially suspected that it had something to do with the movement of the sub and/or the character too (or perhaps getting the interpolated draw positions and the "actual" positions mixed up somewhere), but on a closer look that doesn't seem to explain it: the jitter happens even if the sub and the character are fixed in place whenever the camera moves. I tried applying some smoothing to the camera movement, but that didn't have any kind of effect.

I also added some quick and dirty temporal blurring to it, but it didn't work out that well either: the jitter is just so heavy that it's hard to hide it with just blur.

https://user-images.githubusercontent.com/14351070/168280100-bcd58ed1-300e-4ad1-ab76-02b7b1069ff4.mp4

I'm afraid the jitter might actually be an inherent problem with the technique, which happens for the same reason as very thin geometry not occluding. Since the number of raycasts and raycast steps is limited, the shape of the shadows can have some (essentially random) variation depending on how the raycasts happen to hit the corners of the occluders.

So far I haven't been able to come up with a good way to solve this, but I'll have a think about it and see if I can figure something out, because otherwise this effect looks so much more better than the vanilla one that I'd really love to be able to take advantage of. In the meantime, if you have any ideas on potential ways to address that, I'd love to hear them!

Working-Joe:
I agree that the jitter is something that should be sorted out. I may have learned to live with it at the time, but looking at it now it's definitely noticable. 

### Parts inherent to the technique
It is in part inherent to the technique like you said. 

I have updated the shader there to use a rectangle for a better comparison with baro's walls. I have also set the options for the step count to the minimum of 64, so problems are extra noticable. 

https://www.shadertoy.com/view/NdlfD8

The jitter happens because at a slightly different position, a sample point may be moved on the outer corner of the. With this implementation the "detection points" of geometry lie on equally spaced concentric rings.

I think this is most easily seen if you move the mouse towards a corner from e.g. diagonally above. You'll notice the shadow slides back and forth as the radial "detection rings" pass the geometry.

In this way discontinuous features of the geometry are sort of reflected in the shadow. There's no way to mitigate this part, aside from more rays and more steps. i've thought about spacing the points differently (by placing them on a spiral, for instance), but then you'll just see the discontinuities in other axes, while adding some overhead in the calculations.

### Possible extra causes in baro
However, it seems to me that the jitter is still a good bit more pronounced in baro for some reason, even with the low settings on the shadertoy. Especially if you move the position smoothly, you don't see much jitter, or not in the same way at least. If the camera is stationary, there should be no jitter.

The only reason I can think of is that the camera is never truly stationary relative to the occluders. If you move around the mouse in a small circle in the shadertoy you do get jitter more similar to what's seen in baro. I also initially thought there could be some precision errors relating to the sub movement. But with your tests relating to the camera pos not changing things I'm not sure that could still be the case either. 

### Possible Solutions
One thing that might work is to make the temporal blur a bit more aggressive, but make it very additive. That is, rather than just adding a bit each frame, the shadow is added each frame at full opacity, added to a slightly decreased shadow of the previous frame. 

In the vid you sent when you move around the camera a lot you can see through walls in some places, which suggests to me you are adding a fraction of the shadow rather than the whole shadow for the temporal blur. By instead adding the full shadow to the previous shadow with some fraction you would not be able to see through walls. That may make it possible to get away with a bit slower of a blur, something like `.95*old_shadow+new_shadow`. 

It might be a little ugly though as the blurred shadow moves with the camera, rather than with the geometry. Perhaps the blur fac could decrease with the camera movement speed?

A little heavier an option might be to use a larger texture than the screen (in world space, not in pixels) for the shadow, and position that on the screen relative to the geometry. Then the blur could be even stronger, at a sacrifice of quality changes at different zoom levels.  

This is what comes to mind now, if I end up thinking of anything else that could cause or solve things I'll follow up

Working-Joe:
Note that there's also still the option to try different ways of making the shadow map. Since in this PR I just used the simplest raycast approach, so that it could be done in one shader pass. 

The reduction based approach here
http://www.catalinzima.com/2010/07/my-technique-for-the-shader-based-dynamic-2d-shadows/
which I also mentioned in #6000 might be possible alternatives too. Though you'd need a larger number of shader passes (though without the large loops like I have in this PR) and more intermediate rendertextures, it should find the smaller geometry better as all pixels in the occlusion map are considered.

Though I think some sort of raycast to find the distance to the geometry is required if you want the LoS to be based on the sprites themselves rather than the simplified geometry, which is what causes the problems in the current LoS. Otherwise you'd have to go the alternate route of adding geometry to the occluders.

Issue #8947, opened 2022-04-24
Unkillable ballast flora after docking with a station


shoffing:
- [yes ] I have searched the issue tracker to check if the issue has already been reported.

**Description**
There is a ballast flora in our ship. We've completely drained the ballast tank and have been shooting it with plasma cutters for over five minutes. Water level is also not synced, whoever manages the electrics on the pump sees the water go down, but everyone else still sees a little bit of water. (This seems like a separate bug?)

Even with the host removing the water, the ballast flora will not die. There's no health bar. The pump does not catch on fire.

Here's us trying to kill it, after undocking from the station, holding welders on it for over 2 minutes consecutively. Still alive. We also tried all our grenades and explosives. We even overloaded the reactor, and noticed it's still not dead.
![image](https://user-images.githubusercontent.com/2903742/164955177-4913e571-ea48-4561-a40c-2ea72f08d55d.png)

**Steps To Reproduce**
- We got ballast flora
- We killed it with welding torches
- Later, we got another ballast flora.
- We _almost_ killed it with welding torches. Its health bar was _almost_ empty.
- We docked with a station. (I think this is where the bug originates)
- When we transitioned to station, the ballast flora became invincible, no health bar, etc.

**Version**
v0.17.12.0

**Additional information**

Here's our save file, I hope it helps you track this down! 
[unkillable_ballast_flora_save.zip](https://github.com/Regalis11/Barotrauma/files/8548980/unkillable_ballast_flora_save.zip)

You can kill the non-root pieces, but you can't kill the root. It just wiggles without ever dying.

Is there some console command I can use to recover our save? We'd really like to keep playing it, but it's impossible to continue with this unkillable plant person.

shoffing:
UPDATE:

I think `enablecheats`, `killmonsters` killed it. Can't get achievements now but at least we can continue our save...

cshingle:
Same exact issue. using the killmonsters command didn't make it disappeared, but it did stop growing and the pump worked. All I really have to add is the exception when we first docked. 
![20220424010010_1](https://user-images.githubusercontent.com/2231937/164959496-7a93aa14-9f9e-4000-9b5b-1e6443f9f9a8.jpg)
![20220424011337_1](https://user-images.githubusercontent.com/2231937/164959498-a878c7c0-a53a-4aff-a638-dc841ca19ede.jpg)
[...]

PR #8956, opened 2022-04-24
Fix unkillable ballast flora bug

shoffing:
This fixes https://github.com/Regalis11/Barotrauma/issues/8947

After making this change and building locally, I was able to kill the ballast flora after loading my save: 
image


shoffing:
Note that this doesn't fix the root cause, where `branchElement.GetAttributeInt("parentbranch", -1);` ends up being out of range of the Branches List. However it does prevent the user-facing bug of unkillable ballast flora.

JoneKone:
Screenshot from 2022-04-23 19-15-37


Do you mean this issue? I spend half a tank in this pump, the progress bar was empty during the whole half bottle.

Regalis11:
Thank you for the PR! As mentioned in the ticket, we've fixed this at our end: this PR helped me find the place where the issue is, but the way I addressed it was a little different. The problem was that the game was trying to assign a branch that hadn't been loaded yet as the current branches parent, and I fixed that by assigning the parent branches after all the branches have been loaded.

Issue #9022, opened 2022-04-29
Native Apple Silicon build


Ichicoro:
I tried building Barotrauma on M1 natively, and the performance (well, the FPS) almost doubled (went from an average of around 110 to an average of 200). To build it, I just had to replace the SDL, freetype and openal dylibs with arm64 ones in the solution and then I had to change the `TargetFramework` from `netcoreapp3.1` to `net6.0` and the `PlatformTarget` from `x64` to `AnyCPU`. I think this is worth a look, since it means higher framerates/lower power consuption on Apple Silicon Macs.

This is with the Intel build of the game (current Steam release)
<img width="1440" alt="image" src="https://user-images.githubusercontent.com/21175939/166072928-b83c6ba3-5281-4d4e-9e38-b8f4cff3bbbe.png">

This is with the ARM64 native build of the game (built on my M1 Air)
<img width="1440" alt="image" src="https://user-images.githubusercontent.com/21175939/166073070-17618082-1c28-410d-8188-8f4f2d38e06a.png">

evilfactory:
I would like to note that while testing the game on net6.0, i noticed that anything related to compression was pretty much broken, all old saves were unable to be loaded, joining any server would cause you to get errors, etc.

cshingle:
I too built the game on net6.0 on Linux and found a significant performance improvement. I was able to load a saved campaign and load a custom sub in the sub editor without any issues.
```
quickstart Azimuth
```
Here is the steam build
![Screenshot from 2022-05-02 16-21-30](https://user-images.githubusercontent.com/2231937/166331995-82a9f0e1-4ae6-470b-b3a6-f5f51fe18c19.png)
vs my release build on .Net 6.0
![Screenshot from 2022-05-02 16-26-33](https://user-images.githubusercontent.com/2231937/166332057-d45fa478-3f4a-4890-9e5b-026791ef3063.png)

cshingle:
I too built the game on net6.0 on Linux and found a significant performance improvement. I was able to load a saved campaign and load a custom sub in the sub editor without any issues.
```
quickstart Azimuth
```
Here is the steam build
![Screenshot from 2022-05-02 16-21-30](https://user-images.githubusercontent.com/2231937/166331995-82a9f0e1-4ae6-470b-b3a6-f5f51fe18c19.png)
vs my release build on .Net 6.0
[...]

PR #9049, opened 2022-05-02
Fix save decompression not working on .NET 6 with other platforms

evilfactory:
Simply removes the LINUX preprocessor check, because the problem also happens on other platforms.

### DIFF
--- a/Barotrauma/BarotraumaShared/SharedSource/Utils/SaveUtil.cs
+++ b/Barotrauma/BarotraumaShared/SharedSource/Utils/SaveUtil.cs
@@ -480,7 +480,7 @@ namespace Barotrauma
             int read = 0;
 
             // FIXME workaround for .NET6 causing save decompression to fail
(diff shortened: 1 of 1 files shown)


cshingle:
This is in reference to #9022

3e849f2e5c:
Thank you for the contribution.

The reason why I set it as Linux only was because I changed the DebugLinux project to use .NET 6.0 as seen here https://github.com/Regalis11/Barotrauma/blob/master/Barotrauma/BarotraumaClient/LinuxClient.csproj#L18-L31 since Edit and Continue that .NET Core 3.1 has does not work on Linux but Hot Reload that was introduced in .NET 6.0 does.

I don't see any reason to not remove it if you guys have an use for it so I've merged it in our private repo in https://github.com/Regalis11/Barotrauma-development/commit/940413854d072d8db8fa24c44d9c582658cf8607

Issue #6305, opened 2021-07-13
Mouse4 and Mouse5 do not work on Linux


LastTalon:
- [x] I have searched the issue tracker to check if the issue has already been reported.

**Description**
Attempting to bind Mouse4 and Mouse5 on Linux doesn't seem to work. 

**Steps To Reproduce**
In the settings in either the Controls of Voice Chat menus selecting an input field and attempting to press Mouse4 or Mouse5 does not detect any input (inputs work fine in other games/applications) and leaves the field waiting for input.

**Version**
v0.14.6.0 Linux.

**Additional information**
This is possibly related to closed issue #915. I checked there, but didn't see any discussion.

emmyposs:
The problem's with SDL here - SDL_GetGlobalMouseState doesn't check for the extra buttons on X11. As a workaround, you can patch MonoGame to get button state from GetMouseState instead (Input/Mouse.SDL.cs, line 28 - add `state = Sdl.Mouse.GetState(out _, out _);`)

Rokvach:
Tested, I was unable to bind my mouse 4th and 5th buttons on Linux but I was able to bind them just fine on Windows, moving to Review.

Regalis11:
@Rokvach Works fine at my end. Tested on Ubuntu 18.04.2, with a Logitech M720.
![Screenshot from 2022-09-07 20-49-06](https://user-images.githubusercontent.com/14351070/188945492-2e71d212-d507-4c24-809b-9e5a4e84bc45.png)

Which Linux distro/version and mouse are you using? Are you sure the 4th and 5th mouse buttons work elsewhere on your system, and it's not for example a driver issue of some sort?

Regalis11:
But do the buttons work outside the game? If not, I would suspect it's a driver issue.

Rokvach:
I had to configure my VM to support those buttons, tested now again and they are working correctly. Closing.

PR #9055, opened 2022-05-03
Make 4th 5th mouse buttons work on Linux

cshingle:
I followed emmyposs's advice here #6305 to make the 4th and 5th mouse buttons work.

I tested it on my system (Ubuntu 20.04) and it works. I am able to assign the fourth and fifth mouse buttons to actions and they work in game. They also get labeled correctly in game which is nice as well. I didn't do extensive testing but I did not notice any negative side affects.


mygamingaccount:
Working great, no adverse effects!

PR #9099, opened 2022-05-10
Allow characters to switch attacks without snapping ropes from previous attacks

Emanthealmighty:
This PR adds a way to "get around" one of the most recent changes introduced in the "Urban Expanses" update:
` Ropes attached to limbs now automatically snap when another attack is chosen.`
The abovementioned is pretty a pretty useful QoL improvement which makes it very easy to make a character drag another character towards itself and immediately release it after attacking. However it makes it _impossible_ to make a character drag another character towards itself and NOT immediately release it without taking more complicated steps.

```
Adds a new parameter to attacks: "SnapRopeOnNewAttack", set to true by default.
If true, when a new attack is picked to be used next by a character, any rope created by a previous attack will automatically snap.
If false, when a new attack is picked to be used next by a character, ropes will be left intact.
```

This would allow, for example, to create a monster which drags a crewmember towards itself and periodically attacks it while holding onto it and also swimming away.

### DIFF
--- a/Barotrauma/BarotraumaShared/SharedSource/Characters/AI/EnemyAIController.cs
+++ b/Barotrauma/BarotraumaShared/SharedSource/Characters/AI/EnemyAIController.cs
@@ -87,11 +87,11 @@ namespace Barotrauma
                 if (_attackLimb != value)
                 {
                     _previousAttackLimb = _attackLimb;
-                    _previousAttackLimb?.AttachedRope?.Snap();
+                    if (_previousAttackLimb != null && _previousAttackLimb.attack.SnapRopeOnNewAttack) { _previousAttackLimb?.AttachedRope?.Snap(); }
                 }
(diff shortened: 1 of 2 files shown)


itchyOwl (commented):
Approved and merged to dev in Regalis11/Barotrauma-development@02cff1c.

NilanthAnimosus:
Tested against dev commit https://github.com/Regalis11/Barotrauma-development/commit/6c89a275a94f5c3410996a8f8ebdad8227021421, no issues found, closing.

Notes: Tested on latchers changing their attack frequency and general AI, tested what occurs after they die if the tongue can be removed or removed itself instead of just if it detaches automatically on new attacks. all seems fine (Actually seems the rope they attached may be more related to the item, but either way its possible for modders to have no issues there so I consider it fine without changing that).

Latcher was able to have multiple tongues at once connected to a submarine, so the intended behavior is fine.

PR #9145, opened 2022-05-15
Fallback ConversationAction text to the tag text

evilfactory:
This allows modded server builds to send conversation events with custom text without having to define localization tags with XML


Regalis11 (approved):
Can't see any issues here, I think we can merge this after we've tested it internally. Moving to QA.

Rokvach:
Tested, didn't find any issues, can be merged to dev.

PR #9191, opened 2022-05-22
Alter AttackResult ctor

Cintique:
Make Afflictions an empty list rather than null. Reason: Causes a potential NRE with HumanAIController, rather than altering that logic and leaving other avenues for potential NREs I think it's better to just make AttackResult.Afflictions not null. This generally shouldn't cause an issue because it's just an empty enumerable.


Regalis11:
I don't think this is a good idea, because AttackResult structs are created so frequently that making them always instantiate a new object would generate a ton of garbage.

In my opinion a much better alternative would be to turn Afflictions into a property that returns an empty enumerable when the backing field is null.

Issue #9237, opened 2022-06-01
It is currently impossible to mod an item without copying its xml in its entirety


cshingle:
### Disclaimers

- [X] I have searched the issue tracker to check if the issue has already been reported.
- [X] My issue happened while using mods.

### What happened?

It is currently impossible to mod an item without copying its xml in its entirety. 
This is a problem because;
1. It means your mod will become obsolete if the item is changed in any way. For example a new balance release that modifies the price and fabrication requirements.
2. Creating a mod that modifies something on someone else's mod such as the fabrication cost requires you to basically plagiarize their xml to change a single attribute on their items.
3. Simple little tweak mods are impossible so you may as well just copy pasta the tools.xml and abandon it after the next release.
4. LuaCs hooks are overkill for item mods.

```
<?xml version="1.0" encoding="utf-8"?>
<Items>
  <Override>
    <!-- Inherit and override to make harpoons orange -->
    <Item name="Spear Override" identifier="spear" inherit="spear" spritecolor="255,165,0,255" InventoryIconColor="255,165,0,255">
    </Item>
  </Override>
</Items>
```

This example fails with the error below. It actually inherits the attributes from the vanilla spear correctly before adding the prefab to the PrefabCollection, but then fails validation on this line `IdToNode.Values.ForEach(v => FindCycles(v, uncheckedNodes));` because the parent entity has been replaced with it self causing a circular reference.

![image](https://user-images.githubusercontent.com/2231937/171324945-d8665cad-f299-47f7-8c90-1b37b929a0a9.png)

Override and inherit/instanceof are the most intuitive way to accomplish this, but all we really need is a way to achieve the same result.

### Reproduction steps

- Create a local mod like the xml above that tries to override and inherit the same item.
- Enable the mod.

### Bug prevalence

Happens regularly

### Version

0.17.15.0

### Which operating system did you encounter this bug on?

Linux

PR #9294, opened 2022-06-09
Add the ability to inherit AND override a prefab

cshingle:
This modifies the prefab loading code to allow inheriting from a prefab while overriding it. This is useful for "patching" existing prefabs without having to duplicate them in their entirety. It also allows patching chains to exist where ModA modified an item from Vanilla and ModB can then modify it further without undoing what ModA did (assuming they are modifying different properties).

With this change the following is now a valid item that will override only the name, spritecolor and InventoryIconColor. All other properties are inherited from the parent version that is being overridden. It is now possible to create mods that tweak prefabs from other mods as long as you load them in the correct order.

```
<?xml version="1.0" encoding="utf-8"?>
<Items>
  <Override>
    <!-- Inherit and override to make harpoons orange -->
    <Item name="Spear Override" identifier="spear" inherit="spear" spritecolor="255,165,0,255" InventoryIconColor="255,165,0,255">
    </Item>
  </Override>
</Items>
```

I tested with the mods included in the zip as well as a number of my mods and mods I got from the workshop. You can use this order to chain patches over a few mods.

mods

I also included `inherit override self`, `inherit without override` and `inherit without override 2` that are all doing unsupported things that should prevent the mods from being activated with an error message.

[test-mods.zip](https://github.com/Regalis11/Barotrauma/files/8867148/test-mods.zip)

This relates to item #9237

### DIFF
--- a/Barotrauma/BarotraumaShared/SharedSource/Prefabs/PrefabCollection.cs
+++ b/Barotrauma/BarotraumaShared/SharedSource/Prefabs/PrefabCollection.cs
@@ -133,6 +133,8 @@ namespace Barotrauma
                 
                 prefabCollection
                     .Cast<IImplementsVariants<T>>()
+                    // don't return the prefab instance that triggered the query
+                    .Where(p => p != prefab)
                     .Where(p => p.VariantOf == id)
                     .Cast<T>()
                     .ForEach(p =>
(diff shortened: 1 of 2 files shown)


Barotrauma/BarotraumaShared/SharedSource/Prefabs/PrefabCollection.cs
```diff
@@ -174,12 +176,34 @@ public void AddNodesAndInheritors(IEnumerable<Identifier> ids)
             public void InvokeCallbacks()
             {
                 HashSet<Node> uncheckedNodes = IdToNode.Values.ToHashSet();
+                // We no longer need to worry about parent self references, but maybe this will still catch other issues???
```
juanjp600, line 179: Cycles are still a problem in cases like:
- `A` inherits from `B`
- `B` inherits from `C`
- `C` inherits from `A`

juanjp600 (changes requested):
Left a couple comments. I would also like to suggest using a different attribute for extending the behavior of an existing prefab. Self-inheritance is not a good representation of what you're trying to achieve here, and implies some potentially complex logic (inheritance trees where the same identifier shows up several times, but refers to different prefabs) that as far as I'm aware, would be impractical to implement correctly.

Regalis11:
Closing this at least for now, as the changes aren't usable in their current form.

Issue #6412, opened 2021-08-09
Multiple Dedicated servers (2) from same ipv4 seem to mess up master servers listing's


JoneKone:
- [x] I have searched the issue tracker to check if the issue has already been reported.

**Description**
Tried to spin up multiple dedicated servers different names (different ports) same ip (first server loses ping value on serverlist and second is not visible)

**Steps To Reproduce**
spin up 2 dedicated server with own files from same ip.

**Version**
v0.14.6.0

'''' additional info ''
You can direct connect to the unlisted fine, the first one is visible on server list and you can connect to that 2.

Regalis11:
Fixed in https://github.com/Regalis11/Barotrauma-development/pull/3605/commits/d0ef7bc6d9e560aa6c6003bc641ca69f013838d4

NilanthAnimosus:
Tested against serverlist-ping-fix commit https://github.com/Regalis11/Barotrauma-development/commit/95c040146060302b57b5c01dda76c63a2fd4ee31, no issues found

Notes:
Before the fix on the same machine even (using both different port and query port per server instance):
![image](https://user-images.githubusercontent.com/29177976/194882905-cadec805-d5bd-41ed-9f77-568a51ac0ecb.png)

Same test with fix:
![image](https://user-images.githubusercontent.com/29177976/194883089-82f056d8-6416-495c-bb58-f264213be180.png)

I think this can be safely merged. but I've only tested this on my own machine (which still seems to be an improvement overall for the server list for both public servers).

Rokvach:
Moving to waiting as this is waiting to be merged

PR #9479, opened 2022-06-28
Fix multiple dedicated servers running on same IP Address not working properly

evilfactory:
This PR fixes two bugs:

1) Multiple servers on same IP wouldn't show up on the server browser, this is fixed by adding the port value check that was left out here:
https://github.com/Regalis11/Barotrauma/blob/master/Barotrauma/BarotraumaClient/ClientSource/Screens/ServerListScreen.cs#L1766-L1767

2) Multiple servers on same IP would have "?" as ping, this is fixed by changing activePings so it uses a ServerInfo object instead of a string IP Address

Related to #6412

### DIFF
--- a/Barotrauma/BarotraumaClient/ClientSource/Screens/ServerListScreen.cs
+++ b/Barotrauma/BarotraumaClient/ClientSource/Screens/ServerListScreen.cs
@@ -81,7 +81,7 @@ namespace Barotrauma
         private List<ServerInfo> favoriteServers;
         private List<ServerInfo> recentServers;
 
-        private readonly Dictionary<string, int> activePings = new Dictionary<string, int>();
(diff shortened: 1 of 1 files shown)


MassCraxx:
Please add this soon. It would help a lot.

mygamingaccount:
@MassCraxx you can build the forked branch https://github.com/evilfactory/LuaCsForBarotrauma/tree/multiple-dedicated-servers-fix if you need this

evilfactory:
> @MassCraxx you can build the forked branch https://github.com/evilfactory/LuaCsForBarotrauma/tree/multiple-dedicated-servers-fix if you need this
> 
> Here's how: https://github.com/Regalis11/Barotrauma/blob/master/CONTRIBUTING.md#building-the-game

This won't work, every single player would need to use that custom build of the game, and that's not doable.

mygamingaccount:
I see the change is client side, never mind.

PR #9480, opened 2022-06-28
[Feature Request] Improved PowerContainer recharge rate on autocontrol

Oukido:
This PR changes the behaviour of powercontainers when autocontrolled.

- Increases the accuracy of rechargerate adjustments from `maxRechargeSpeed * 0.1f` to `maxRechargeSpeed * 0.01f`
- Adds a LockTimer variable and isAutocontrolled Variable
- Locks rechargeSpeedSlider.Enabled when the rechargespeed is controlled by signals.

The change allows for smaller increments when circuits control the component, while the manual interaction remains the same for increments, but is locked when signals control the chargerate. This also prevents accidental firing of network events when the slider is manually pulled but since it is autocontrolled it is instantly reset to the signals value. Locking the slider prevents that.
Network event resolution remains the same with 10% steps on manual interaction.
IsAutocontrolled variable allows for addition of a gui-element to signal the autocontrol state if desired.

Discussion:
https://github.com/Regalis11/Barotrauma/discussions/9420

Note:
Manual interaction with the slider in its current vanilla implementation can lead to high frequency network events since every OnMoved event also triggers a ClientEvent.
This is also true for Pumps when not autocontrolled.

### DIFF
--- a/Barotrauma/BarotraumaShared/SharedSource/Items/Components/Power/PowerContainer.cs
+++ b/Barotrauma/BarotraumaShared/SharedSource/Items/Components/Power/PowerContainer.cs
@@ -111,14 +111,7 @@ namespace Barotrauma.Items.Components
             {
                 if (!MathUtils.IsValid(value)) return;
                 rechargeSpeed = MathHelper.Clamp(value, 0.0f, maxRechargeSpeed);
-                if (IsAutoControlled)
-                {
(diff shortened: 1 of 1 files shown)


Regalis11 (changes requested):
This PR doesn't seem to be in a mergeable state: the changes to the recharge speed adjustment don't work properly in multiplayer because the value is synced as an integer between 0-10.

PR #9607, opened 2022-07-19
[Feature Request] - Factor in submarine depth to the crush depth damage interval

Jade-Harleyy:
Currently, a team of about ~3-5 players (depending on sub size) could weld walls faster than crush depth damaged them, making crush depth more of a nuisance than something to avoid.

With this change, crush depth would become more of a danger to the crew and something to avoid reaching (unless under some certain situations).


Jade-Harleyy:
@Regalis11 I've went ahead and implemented some optimizations in the latest commit, was able to run at ~240 fps on max settings at the speediest damage interval. (albiet with little else on the map)
If it still manages to be extra laggy on older computers, tweaking the minimum interval might be able to cull some of the lag.

Issue #9638, opened 2022-07-25
AI does not properly open two nearby located doors


pakvg85:
### Disclaimers

- [X] I have searched the issue tracker to check if the issue has already been reported.
- [ ] My issue happened while using mods.

### What happened?

Situation: there are two doors that are placed nearby. A bot opens first door, then almost immediately tries to open second door. The problem is - when the bot is opening the second door, he has not yet fully passed through the first door. This could cause incorrect behavior if the second door's logic results in closing the first door.
Example: https://www.youtube.com/watch?v=WpQyH1VJI00
Test submarine: [Test Sub.zip](https://github.com/Regalis11/Barotrauma/files/9179288/Test.Sub.zip)

What I expect: the bot should not try to press second door's buttons until he fully passes through the first door.

I decided to separate this issue from #9600 just to make it easier. I will create a pull request for this a bit later.

### Reproduction steps

1. In sub editor place 2 doors near each other.
2. Place buttons and logic components so there will be a button that in the same time opens second door and closes first door.
3. Place 2 bots so they will be separated by 2 doors.
4. Command a bot to follow another bot.

### Bug prevalence

Happens every time I play

### Version

0.18.15.0

### Which operating system did you encounter this bug on?

Windows

MaaiG:
I have a similar problem. A custom submarine.
The bot opens the door but not goes forward as the next door is closed.
The next door should open when the first closes.
But the bot opens the first door and just stands.
The bot must approach the locked door but it does not! Apparently 
 the bot expects the second door to open, at a great distance from it.
It seems to me the problem is at what distance the bots open the door or at what distance they expect the path to open.

pakvg85:
Could you attach your custom sub here in a zip file?

MaaiG:
> Could you attach your custom sub here in a zip file?

You want to test the door!? The idea is simple: the ballast door opens in front of the character automatically if there is no water (If water - manual mode) and is a primitive gateway (as simple as possible for AI). But 90% of the time the bots don’t know how to use it 😔.
So, that is a partial copy of the circuit (there is nothing superfluous): [_Test-doors.zip](https://github.com/Regalis11/Barotrauma/files/9214021/_Test-doors.zip)

itchyOwl:
Fixed in Regalis11/Barotrauma-development@7320dbdad51d0426b1b4857c68d523f556835d66

Regalis11:
[...]

Issue #9600, opened 2022-07-18
AI stucks in airlock-like compartment (it clicks on wrong button)


pakvg85:
### Disclaimers

- [X] I have searched the issue tracker to check if the issue has already been reported.
- [ ] My issue happened while using mods.

### What happened?

Situation: there is an airlock-like compartment with doors and hatches that are wired to corresponding buttons via some logical elements. When a bot approaches to this compartment, it chooses to click on button that is not directly related to that door/hatch and thefore stucks eternally clicking on wrong button.
What I expect to happen: bot chooses the correct button.

Below are few examples and a test submarine file.
https://drive.google.com/file/d/12r56BJ-YfCdBNleBv6jaGUcwp_BNZC3-/view?usp=sharing

Example # 1 (https://youtu.be/buuy4-cyvrk)
   Compartement does not have logical components and the bot successfully opens doors/hatches. Everything is fine.

Example # 2 (https://youtu.be/sglqQcNqjAs)
   In this example buttons are wired to doors with some logical elements. The goal of logical scheme is: when one door (or hatch) is being opened, the other two neighbour doors should be closed.
   What happens here: bot comes to first door, clicks on button that relates to that door, and then immediately tries to open the hatch (because that hatch is located pretty close to the door and is accesible from bot's position).
   But instead of clicking hatch's button, the bot for some reason clicks on the first door's button, and it immediately being closed.
   I believe it happens because of this: when the bot's current waypoint is a door waypoint, bot tries to find the nearest button that is wired to the corresponding door/hatch, no matter what logic is behind that wiring - any button will fit. In this example it leads to incorrect button.

Example # 3 (https://youtu.be/eUtrQpU0NtI)
   This example is similair to previous example (# 2) - buttons are wired to doors with some logical elements, and the goal of logical scheme is the same (when one door/hatch is being opened, the other two neighbour doors should be closed).
   The difference is in starting position of the bot - now it is located to the left of the compartment.
   What happens here: bot comes to first door, clicks on button that relates to that door. The door opens normally (unlike exmaple # 2), because the hatch's button is located quite far from the first door and could not be clicked immediately.
   After first door is opened, bot comes to the hatch and tries to open it, but because of the same issue, he mistakenly defines that suitable button for that hatch is the nearest button to his center of mass - the first door's button.

### Reproduction steps

1. Open test submarine from "Test Sub.zip" file
2. Place bots at the positions that are shown in examples above
3. Command a bot to follow other bot
[...]

PR #9639, opened 2022-07-25
[Bugfix] AI does not properly open two nearby located doors (#9638)

pakvg85:
PR for issue #9638

After the bugfix: https://youtu.be/oIkn7GN3qV0
Now AI is allowed to open the hutch only after the bot is fully passed through the first door.
Notice that the hutch is not yet properly opened because issue https://github.com/Regalis11/Barotrauma/issues/9600 is not yet resolved.

### DIFF
--- a/Barotrauma/BarotraumaShared/SharedSource/Characters/AI/IndoorsSteeringManager.cs
+++ b/Barotrauma/BarotraumaShared/SharedSource/Characters/AI/IndoorsSteeringManager.cs
@@ -86,6 +86,7 @@ namespace Barotrauma
             this.CanBreakDoors = canBreakDoors;
 
             character = (host as AIController).Character;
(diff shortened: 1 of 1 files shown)


pakvg85:
After the bugfix: https://youtu.be/oIkn7GN3qV0
Now AI is allowed to open the hutch only after the bot is fully passed through the first door.
Notice that the hutch is not yet properly opened because issue #9600 is not yet resolved.

Regalis11:
Thank you for your contribution!

However, I don't think this fix works correctly in all situations, and might actually lead to more issues than it solves. There may be situations where they need to open the next door even before they're through the current one, for instance when climbing up a hatch with another hatch above it. It might also cause problems in situations where the waypoints are set up in a way that prevents the bot from moving through the door/hatch to the next waypoint before that next door/hatch is open.

PR #9679, opened 2022-08-04
Increase MaxValueInt of LevelObjectAmount in CaveGenerationParams.cs

Jade-Harleyy:
Increases the maximum amount of objects in a cave from 1000 to 10000.

Useful for modders who want to make more detailed or less empty caves.

Only performace drawbacks are increased loading times while using mods that make use of this change.


Regalis11:
Thank you for the PR! Merged to our private repo in https://github.com/Regalis11/Barotrauma-development/commit/e7ff17ddcb425d045e2f9556686a0dc29fe675cd

PR #9776, opened 2022-08-27
[BugFix] JB ExtraLoad negative power

Torca2001:
Extraload when set to negative would push negative power onto the grid instead of adding power to the grid.
Simply a +- mistake during implementation.

### DIFF
--- a/Barotrauma/BarotraumaShared/SharedSource/Items/Components/Power/PowerTransfer.cs
+++ b/Barotrauma/BarotraumaShared/SharedSource/Items/Components/Power/PowerTransfer.cs
@@ -273,7 +273,7 @@ namespace Barotrauma.Items.Components
 
         public override float GetConnectionPowerOut(Connection conn, float power, PowerRange minMaxPower, float load)
         {
-            return conn == powerOut ? PowerConsumption + ExtraLoad : 0;
+            return conn == powerOut ? MathHelper.Max(-(PowerConsumption + ExtraLoad), 0) : 0;
         }
 
         public override bool Pick(Character picker)


ghostsheet:
you should add in extravoltage then

Torca2001:
Could you elaborate?

Do you mean for it to add power based on current load? So that some of the grid demand could be met? Or to just increase a device's voltage?

Regalis11:
I think ghostsheet may have misunderstood what this change does. I can't see any problems with the fix, but one thing I'm not sure about, is this method actually even used in any situation? To me it seem Powered.UpdatePower updates PowerTransfers separately and never calls this method for those. I also tested putting a breakpoint in the method, and couldn't get it to trigger.

Torca2001:
Currently there is nothing that utilises a negative ExtraLoad due to previous implementation doing nothing with negative numbers. This fix is more to stay consistent with the way the power system was implemented and also gives modders a possible way to directly interact with a junction box's grid like how Ballast Flora currently does but allowing for Power to be added instead of just increasing Load.

Regalis11:
I mean, to me it seems like currently there is nothing that calls this method. So the method could just as well return nothing, and it'd make no difference. It also doesn't seem like this affects modders in any way. Am I missing something here, is there some situation where this method is called?

Torca2001:
I'm probably explaining this poorly, but currnetly there are only 2 use cases of ExtraLoad.
Ballast flora which is implemented in code where it changes the ExtraLoad variable on Junction boxes and
Piezo Crystals which use it through the statusEffect system in minerals.xml Line 1072
`<StatusEffect type="OnActive" target="NearbyItems" targettags="junctionbox" range="3000.0" setvalue="true" stackable="false" delay="0.95" extraload="500" />`

The behaviour of a Piezo crystal zap being that it will target all the nearby junction boxes and set their ExtraLoad value to 500,
this results in each junction box adding 500 load.
Now what this change is meant to allow for example lets change the Piezo Crystal ExtraLoad to -500
`<StatusEffect type="OnActive" target="NearbyItems" targettags="junctionbox" range="3000.0" setvalue="true" stackable="false" delay="0.95" extraload="-500" />`

Now the behaviour (with this fix) should be that each Junction Box will add 500 power to their connected grid.
This behaviour seems reasonable as actually taking away load from a grid doesn't make much sense and will mess with most of the power calculations.

This change being targeted for use in the statusEffect system which is why I was saying it gives modders a more direct way to interact with a power grid. An example use case is lets say we want to implement that an artefact that will zap junction boxes and slightly overvolt your grid which would be achieved by simply adding power to the grid. This fix would allow for a simple statusEffect that sets the extraload value on junction boxes.
Without this change the only class that would allow for specific power output to the grid regardless of the state of the grid is reactors. Which implementing through xml would need for it to have no UI, a fuel item, controlling of fission rate and turbine rate, and since it is a reactor class all "reactors" will try to balance with each other and changes the main sub reactor behaviour.

The reason why this function in your testing is never fired is because it is only intended to fire if the device is producing power. Of course currently there is no scenario in Barotrauma where a junction box produces power hence why the function currently won't fire. So you can only get it to fire if you modify the xml of a junction box to have a negative extraLoad.

All this fix does it at the moment if we set a negative ExtraLoad of -100 for a grid that has
Power: 40
Load: 200
Current bug is that we get
Power: -60
Load: 200
What it should be and what the fix does is
Power: 140
Load: 200
And of course if we use a extraLoad of 100 we get
Power: 40
Load: 300

Regalis11:
Thank you for the clarification!

> The reason why this function in your testing is never fired is because it is only intended to fire if the device is producing power.

This was what I was confused about, now that I tested this with a modified junction box and took a closer look at how the method is used, this makes sense.

Regalis11:
Merged to our private repo in https://github.com/Regalis11/Barotrauma-development/commit/609bd5db4cb19fd34db50908c388c58468fe2ebc

Issue #9848, opened 2022-09-08
Battery, Supercapacitor  - charge indicator does not rotate when the battery is rotated


MaaiG:
### Disclaimers

- [X] I have searched the issue tracker to check if the issue has already been reported.
- [ ] My issue happened while using mods.

### What happened?

Charge indicator does not rotate when the battery is rotated!

### Reproduction steps

![Image 001_22-28](https://user-images.githubusercontent.com/93348499/189160500-986dbbe7-e51c-45fe-8c3c-9c281f1ddc41.png)
![Image 002_18-21](https://user-images.githubusercontent.com/93348499/189161761-42e30ec2-605c-41d4-a7fd-aea313035067.png)

### Bug prevalence

Happens every time I play

### Version

0.18.15.1 / 0.18.15.2 on macOS

### Which operating system did you encounter this bug on?

Windows

Regalis11:
Fixed in https://github.com/Regalis11/Barotrauma-development/commit/59f2fe767592d2868f5f621e89e468b121a62c1c

someone972:
So github will link these: This should have also fixed issue #9106

NilanthAnimosus:
Tested against dev commit https://github.com/Regalis11/Barotrauma-development/commit/3f874d71a8cdc82d1d16314abdd6ae91f7e42098, no issues found, closing.

PR #9978, opened 2022-10-04
Fix power indicator not rotating with battery.

someone972:
PR for issue #9848

Drawing the charge indicator did not take into account the rotation of the parent container. This fixes this by adding two functions for drawing rotated rectangles, one for an outline and one for filled. The function for drawing the outlined rectangle has been written to accept the desired position for the outline for increased ease of use.

Behavior maintained from previous implementation:
- Mirroring the parent X or Y does not mirror the indicator.

Behavior changed/added:
- The outline is now drawn on the inside of the given coords, so that the overall drawn size matches the size provided.
- The charge indicator accounts for the thickness of the outline so that very small charge levels are still visible and not covered by the outline.
- Added a 'FlipIndicator' property to the power container so that it is possible to truely mirror the object if desired.

Before changes:
BeforeChange

After changes:
AfterChange

Test sub:
[PowerIndicatorTests.zip](https://github.com/Regalis11/Barotrauma/files/9710214/PowerIndicatorTests.zip)

### DIFF
--- a/Barotrauma/BarotraumaClient/ClientSource/GUI/GUI.cs
+++ b/Barotrauma/BarotraumaClient/ClientSource/GUI/GUI.cs
@@ -3,7 +3,6 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using Barotrauma.IO;
 using System.Linq;
-using System.Xml.Linq;
 using Barotrauma.CharacterEditor;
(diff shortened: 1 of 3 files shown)


Regalis11:
Thank you for the contribution! Merged to our private repo.

PR #10031, opened 2022-10-09
Quality of life: filter item interactions when pressing a particular key

Darker:
I created a simple quality-of-life improvement. I am not sure if this is the best way to do it, but I'd at least like to suggest something along these lines. I was originally intending to also allow switching what filter is applied during the mouse wheel, but I decided it might not be needed. I am keeping that door open though, by leaving the filter as a lambda instead of just passing a bool argument to `FindItemAtPosition`.

Other than the implementation itself, I have also overridden the `IsAttached` getter for doors. I am not sure why they implement `Pickable`, but decided it is safe to assume they are attached since they cannot be ever dropped as an item on the ground.

I did change the `FindItemAtPosition` signature as the original filtering by array was unused - the function has only one refference. I also added a performance change to it - if the item is way outside the search range, ignore it straight away.

This needs to be added to translations:

```
<inputtype.filteritemfocus>Select pickable items only</inputtype.filteritemfocus>
```

### DIFF
--- a/Barotrauma/BarotraumaClient/ClientSource/Characters/Character.cs
+++ b/Barotrauma/BarotraumaClient/ClientSource/Characters/Character.cs
@@ -503,9 +503,8 @@ namespace Barotrauma
         /// </summary>
         /// <param name="character">The Character who is looking for the interactable item, only items that are close enough to this character are returned</param>
         /// <param name="simPosition">The item at the simPosition, with the lowest depth, is returned</param>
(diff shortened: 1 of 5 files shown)


Darker:
Yes, that was the intent. Without this, it was often really difficult to pick spears for the spear gun and fruits, since the character would instead keep interacting with objects on the submarine.

I created this PR over a year ago though, so it might need to be rebased.

Regalis11:
This has been implemented in a slightly different way in the latest update in the form of the "interaction labels", closing

PR #10133, opened 2022-10-21
Drop Item Hotkey

Petethegoat:
Add a drop item hotkey.

First checks for open containers and tries to place the item, starting with left hand, and then right if there's nothing in the left hand.
Failing that, it drops the item.

Works with two handed items, one or both two handed items, and doesn't allow you to free yourself from handcuffs.

I haven't tested it with creatures since I'm not sure how.

It's unbound by default, and there's no included translation, so as is, this doesn't show up correctly in the keybind menu. I assume there's a different content repository to add that to, but if it's something I can provide, please let me know.

There are some videos on my discussion at #10122.


Petethegoat:
neat, thanks!

PR #10150, opened 2022-10-25
Add a simple disallowcargo tag for cargo containers.

Petethegoat:
Simple addition. Any item container tagged as a "cargocontainer" (such as the crate shelf or unit load device) may also be tagged with "disallowcargo" to exclude it from consideration for cargo missions.

I have seen several sub editors asking if this is possible, usually because they have a dedicated cargo hold on their ship and want shelves for manual player storage elsewhere.

Tested in the campaign and seems to function as expected, mission assignment and payout on completion all reflects the lower cargo capacity.

Only downside (in my opinion) is the discoverability of this addition is bad, but if the modding guides and wiki can be updated that should be okay.


Regalis11:
Finally took a look at this. Can't see any issue with the changes, merged to our private repo in https://github.com/FakeFishGames/Barotrauma-development/commit/22bb6885576c73cdacd15480a63c6edb258a81bd. Thank you for the contribution!

Petethegoat:
Thanks!

PR #10154, opened 2022-10-25
Upgrade UI supports upgrade categories with only swappables

Petethegoat:
Related to #10152, though I've basically resolved all the issues I had.

This lets the upgrade UI show upgrade categories that have no upgrade modules, but which do have swappable items that are on the submarine.

The UI works as you would expect- the upgrade indicators are hidden, and the customize/upgrade buttons are hidden.
image

Categories with both (weapons) still show both correctly. as they did before.

The only unintended change is that in order to make upgradeless categories show up when first opening the UI, I call `SelectTab(UpgradeTab.Upgrade)` which does work, but also has the side effect of resetting your selected upgrade category, etc.
Given this only happens when you talk to the NPC to open the upgrade UI, I don't consider this very problematic, but it probably could be improved.

Probably needs more thorough testing, but in my limited testing I haven't found any other issues, and it provides the functionality I wanted to add.

Here is my mod (and a sub with the right swappables for convenience) that can be used for testing: [upgradeUIrepro.zip](https://github.com/Regalis11/Barotrauma/files/9863916/upgradeUIrepro.zip)

Please let me know if there's anything I can add or need to change.

#6180 and #8632 are basically the same as my mod plans and could also could benefit from this enhancement.


Petethegoat:
Not strictly related to this pull request (same behaviour in main branch right now), but with my example mod active: Not entirely sure why the loaders show up here. Possibly they're a special case that were ignored until the fabricator gave this category a real swappable to care about.
image

I think the loaders should not be shown at all, and so I added a check that swappables must be `canbebought="true"` to show in the customization list.

Besides that, a way to let categories be more discriminating about which swappables they want to show would be good. Stuff like the mechanical and electrical categories are too general to show what could potentially be a wide range of swappables.

Petethegoat:
> let categories be more discriminating about which swappables they want to show

I think I've got an adequate solution for this, with the only downside that an XML update would be needed. It's as simple as a new property in the UpgradeCategory for swappable tags, which it can check instead of the general item ones.

So my mod would look like so:
```xml
<?xml version="1.0" encoding="utf-8"?>
<UpgradeModules>
  <UpgradeCategory name="Outfitting" identifier="max_outfitting" swappables="max_outfitting"/>
  
  <Override>
    <UpgradeCategory name="" identifier="weapons" items="turret" swappables="turret"/>
  </Override>
</UpgradeModules>
```

And with that, only my new outfitting category, and the weapons category would have the customize tab shown.
Despite my mod adding SwappableItem to the fabricator, it won't be shown as customisable there because the fabricator category doesn't define any tags for swappables.

I think this is a good solution, but I'll hold off on committing it for now, pending feedback. Code-wise it's basically duplication of the tags hashset, and checking the new one when handling swappables.

Regalis11:
Thank you for the contribution! We finally got around to taking a look at this. It seemed to require some more work still: it seems the item previews on the submarine schematic didn't appear if the category didn't include any upgrades. Addressed that myself and moved this to a PR in our private repo (https://github.com/FakeFishGames/Barotrauma-development/pull/6170), so this will most likely be included in the next or one of the future updates.

Petethegoat:
Nice, thanks.

PR #10201, opened 2022-10-31
Let quest items be found recursively (fix #10193)

### LINKED ISSUE #10193: Quest Items not counted if in Toolbelt
### Disclaimers

- [X] I have searched the issue tracker to check if the issue has already been reported.
- [ ] My issue happened while using mods.

### What happened?

Encountered an opiate overdose on Gope as medic. I had Naloxone in a toolbelt on me, and the quest told me that I didn't have any on me, so there was nothing I could do.

### Reproduction steps

_No response_

### Bug prevalence

Just once

### Version

0.19.14.0

### -

_No response_

### Which operating system did you encounter this bug on?

Windows

### Relevant error messages and crash reports

_No response_

Petethegoat:
Let quest items be found recursively (inside toolbelts or held containers)

Tested and confirmed it fixes the issue described at #10193, but it may cause unintended behaviour elsewhere, so maybe it should be optional.
That said, I don't think it's an issue with the existing vanilla events that use CheckItemAction.

### DIFF
--- a/Barotrauma/BarotraumaShared/SharedSource/Events/EventActions/CheckItemAction.cs
+++ b/Barotrauma/BarotraumaShared/SharedSource/Events/EventActions/CheckItemAction.cs
@@ -97,7 +97,7 @@ namespace Barotrauma
         {
             if (inventory == null) { return false; }
             int count = 0;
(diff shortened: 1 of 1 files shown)


Regalis11:
Thank you for the contribution! We've now fixed this in our private repo with a similar approach (+ with the option to choose whether CheckitemAction should do the check recursively or not).

PR #10288, opened 2022-11-08
Pirate Submarine/Mission Changes

Jade-Harleyy:
Similar to the ideas in #5945, this implements saving pirate submarines in the sub editor with alternate reward and preferred difficulty, instead of needing to define them in the mission file.

Useful for modding as now, similarly to beacon stations, _all_ enabled pirate submarines are used for the mission instead of only the ones defined in the mission file. This would prevent certain mods from overriding each other, and would let the game use both of the mod's submarines.

Additionally, a tagging system has been implemented, where pirate missions will only use submarines with the same tag as the mission.

**Edited Content Files:** [ChangedContent.zip](https://github.com/Regalis11/Barotrauma/files/10034630/ChangedContent.zip)


Regalis11:
This does not really work with the upcoming faction changes I'm afraid, since the submarines are faction-specific and we need to choose a submarine belonging to an appropriate faction based on the mission.

Jade-Harleyy:
@Regalis11 
That would still work with the tagging system that I've implemented, no?

E.g. the "Coalition" mission would have the "coalition" tag, and the submarines specific to that mission would also have the "coalition" tag on them as well.

And if there was a submarine available for multiple types of missions, it would have the tags of all the missions it would appear in.

Regalis11:
Ah, yes, that'd work well with the faction-specific missions too! Should've read this a bit more carefully. This seems like a really welcome change, although I think it's unfortunately going to have to wait for some time, because we've already sent the texts for the next 2 updates into localization and might not be able to squeeze these text changes in anymore.

KrusteDE:
that would be very cool, i have the same issue rn

PR #10311, opened 2022-11-11
Allow the speed that wires and auto-control are able to manage the reactor to be editable in the subeditor. 

DavidHearn:
Good reactor controllers usually require a decent amount of effort in research and design to actually implement in a submarine. With the current hard cap in reactor responsiveness when controlling with wires, the gain in performance is not completely worth those efforts. This change would add two fields to the reactor component in the subeditor that would allow sub creators to fine tune reactor responsiveness to both the default autotemp system and to wires, as well. It would also create the possibility of mods for improving reactor responsiveness to autotemp and wired controllers. An example of one such mod is this repository ([https://github.com/DavidHearn/ReactorAutocontrolUpgrade](https://github.com/DavidHearn/ReactorAutocontrolUpgrade)). This mod adds a new upgrade that can be bought to improve reactor response time. 

PS. By "good" reactor controller, I do not mean bang-bang controllers. Those controllers are dumb but in my personal opinion better made obsolete by changes to overall reactor behavior.

### DIFF
--- a/Barotrauma/BarotraumaShared/SharedSource/Items/Components/Machines/Reactor.cs
+++ b/Barotrauma/BarotraumaShared/SharedSource/Items/Components/Machines/Reactor.cs
@@ -51,6 +51,9 @@ namespace Barotrauma.Items.Components
         private float? signalControlledTargetFissionRate, signalControlledTargetTurbineOutput;
         private double lastReceivedFissionRateSignalTime, lastReceivedTurbineOutputSignalTime;
 
+        private float wiredControlSpeed;
+        private float autoTempControlSpeed;
+    
         private float temperatureBoost;
 
         private bool _powerOn;
@@ -65,6 +68,7 @@ namespace Barotrauma.Items.Components
 #if CLIENT
                 UpdateUIElementStates();
 #endif
+
             }
         }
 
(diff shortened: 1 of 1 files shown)


mygamingaccount:
Fun fact: set the `autoTempControlSpeed` to float.Max, and the reactor automatic mode will controlled via a bang-bang process

VaelophisNyx:
sadly last I recall the developers loathed the concept of ARCs entirely, which kinda sucks given that they were directly nerf-batting people's creativity as a result of disliking that creativity's output

mygamingaccount:
Because creativity is copying hieroglyphs from youtube videos without a clue of what is going on.

That change of slowing down the signal input slew rate resulted in an explosion of new ideas and solutions that people invented since the launch of the update.

Going back would be a downgrade.

DavidHearn:
> Because creativity is copying hieroglyphs from youtube videos without a clue of what is going on.
> 
> That change of slowing down the signal input slew rate resulted in an explosion of new ideas and solutions that people invented since the launch of the update.
> 
> Going back would be a downgrade.

From my memory, the way it's implemented it actually slows down the rate of change of the reactors settings, not how often the input can be changed. But regardless, in my own experience with trying to balance the changes I've experimented with in this PR has been a nightmare. The whole ARC reactor problem is much more complicated than can be fixed by adding a new slider in the subeditor. For that reason I'm thinking about closing the PR.

Anyway, I'd love to see some of the solutions that people have come up with in response to the changes to the reactor. Could you link me some good ones?

VaelophisNyx:
idk I just don't think it was ever a real issue, as it should have just come down to players having the capability **if they so choose**

why remove player choice

mygamingaccount:
> Anyway, I'd love to see some of the solutions that people have come up with in response to the changes to the reactor. Could you link me some good ones?

Plenty of discussion going on in the sub builders forum
You might want to take a look at the School of Illegal Wiring, the overall message volume is lower as that thread is more focused on research.
https://discord.com/channels/103209684680323072/903478852259573760

The methods that tend to work manage the load instead of the reactor

mygamingaccount:
> idk I just don't think it was ever a real issue, as it should have just come down to players having the capability **if they so choose**
> 
> why remove player choice

Trust the devs.
Choice is bloat.

VaelophisNyx:
> Trust the devs. Choice is bloat.

Choice is choice, bloat is bloat. Choice is not always bloat.
The devs were wrong on this one.

DuckY-Y:
I was wondering why the properties of the auto-reactor weren't changeable, turns out there are people here fucking things over. Now I'm stuck with a submarine that fries its junction boxes due to the slowness of the auto-reactor.

mygamingaccount:
skill_issue

DuckY-Y:
Yeah ok I had that one coming, but it would just make it a lot easier if I didn't have to nerf the power usage of endgame subs into the ground or attach a buffer in the grid.

mygamingaccount:
Haven't really thought about this since I mostly delve in solutions related to vanilla subs

You want to add the ability to edit the reactor's external control speed in the submarine editor
So you can fix an unstable, overvolting grid
A grid that you have made unstable by adding engines that have an excessive power value

...

Why don't you Just edit the power value of the engines so you no longer have an unstable grid?

DuckY-Y:
The Camel is a good example of this, I'd like to use it as an endgame sub but how the engine load just fries the junction boxes with a good captain is quite annoying, just being able to change the auto reactor speed in the editor is also fine for me.

VaelophisNyx:
Still think the ARC change should be reverted; it solved nothing in the end and just caused massive upset among the playerbase

PR #10544, opened 2022-11-25
Limb enhancements for status effects

Emanthealmighty:
This PR implements the following:

- Ability to check if a character or limb is flipped.
- Ability to check the speciesname/speciesgroup of a character within status effects for limbs
- Ability to trigger status effects of type "always"/"onactive" for limbs

```
Added "isflipped" conditional for characters. Allows to check if the character is flipped.
Status effects inside of limbs now allow "speciesname"/"speciesgroup" conditionals, as well as "always"/"onactive" status effect types.
```

With these changes, the following could be done:

- Create character variants that don't have the same amount of limbs as the original character (This could allow to fully turn the mudraptor_unarmored and mudraptor_hatchling characters into variants of the mudraptor, thus truly making each variant of the mudraptor actually depend on the mudraptor character).
- Create asymmetrical characters that can visually look different through conditional sprites when flipped, as well as physically look different by hiding and showing certain limbs when the character is flipped.
- Create a character that can lose and regain certain limbs on demand through status effects with whatever conditionals we want.


Regalis11:
Thank you for the contribution! Finally got around to reviewing and testing this. Merged to our private repo in https://github.com/Regalis11/Barotrauma-development/commit/36611b3bc3b632b97937a6e2100f055d3aba2a38

PR #10845, opened 2022-12-31
Make the engine sound level scale with the output

mygamingaccount:
Previously the level would depend on the minimum voltage configured in the editor, only altering the volume if the `prevVoltage` was less than the `MinVoltage`. But this would cause force to be 0 due to no power. Only if minimum voltage was configured higher than 1.0 would affect the equation.

### DIFF
--- a/Barotrauma/BarotraumaShared/SharedSource/Items/Components/Machines/Engine.cs
+++ b/Barotrauma/BarotraumaShared/SharedSource/Items/Components/Machines/Engine.cs
@@ -58,7 +58,7 @@ namespace Barotrauma.Items.Components
 
         public float CurrentVolume
         {
-            get { return Math.Abs((force / 100.0f) * (MinVoltage <= 0.0f ? 1.0f : Math.Min(prevVoltage / MinVoltage, 1.0f))); }
+            get { return Math.Abs(force * Math.Min(prevVoltage,MaxOverVoltageFactor) / 100.0f) ; }
         }
 
         public float CurrentBrokenVolume


mygamingaccount:
Minor change after testing (Math.Max -> Math.Min)
Now the loudness scales just like it does for the AI target

mygamingaccount:
I was considering further increasing the level as `condition` went down so it matches the noise level heard by the AI

For now I'm enjoying the difference this makes. No longer does the engine sound as if it was revving at max power at startup when there's barely any voltage

mygamingaccount:
Also considering raising/lowering the pitch of the fabricators as the voltage changes

Regalis11:
I'm not entirely sure I understand the changes here. Previously the volume was maxed when the engine was at minimum voltage, now minimum voltage is ignored completely.

> only altering the volume if the prevVoltage was less than the MinVoltage. But this would cause force to be 0 due to no power. Only if minimum voltage was configured higher than 1.0 would affect the equation.

This is not quite correct. The volume would scale linearly from 0 to 1 when the voltage approaches MinVoltage, and be 1.0 after that. Now the minimum voltage would have no effect on the volume. I also don't quite understand the purpose of using MaxOverVoltageFactor here - that'd mean the volume can get arbitrarily high, which does not seem desireable?

mygamingaccount:
The Minimum Voltage parameter is set in the editor, defaults to 0.5
Below that there is no sound as the engine is not operating, since `force` is 0.0

At barely on, Voltage = 0.51 you have 0.51/0.5 = 1.02 which is greater than 1.0 so `min(1.02, 1.0)` will use 1.0
At overvoltage Voltage = 1.5 you have 1.5/0.5 = 3.0, `min(3.0, 1.0)` will use 1.0

Only if you set Minimum Voltage above 1.0 will the second half of the equation matter, as that can bring the prevVoltage/MinVoltage quotient to less than 1

MinVoltage is nowhere part of the calculation for the AI range so why should it be considered for the engine sound?

`force` is is not the final force applied to the sub, it is to be corrected by prevVoltage, but not more than the maximum overvoltage factor as that does not increase the total thrust.

Regalis11:
Yeah, that does make some sense, although we can't completely ignore the MinVoltage. That'd mean that if the engine is e.g. configured to work with very little to no power, it'd make barely any sound even if it was running at 100%.

Regalis11:
Fixed in https://github.com/Regalis11/Barotrauma-development/commit/6d12596e02bd560888be8e924193f210fc5214ba by removing the division with MinVoltage. We cannot remove MinVoltage altogether, because then engines configured to work without power wouldn't work.

mygamingaccount:
It would make as much sound as the amount of thrust it outputs. That seems correct to me.

I have built a prototype sub to test this: second reactor only going only to the engine, reactor throttled by the nav terminal output, engine fixed at + or - 100

On my branch the sound is what I would expect.

PR #10962, opened 2023-01-18
Add support for ItemContainer signal outputs (Example Included)

Jade-Harleyy:
<h3>Features:</h3>

This commit makes available 3 connections for items with an ItemContainer component:
- `contained_conditions` - Outputs the sum of the contained items' condition values.
- `contained_condition_percentages` - Outputs the sum of the contained item's condition percents.
- `contained_items` - Outputs the total number of unbroken (condition != 0) contained items.

<h3>Example Content:</h3>

[Content.zip](https://github.com/Regalis11/Barotrauma/files/10448507/Content.zip)
- Adds signal outputs to loaders.
- Changes the reactors' `fuel_percentage_left` connections to use `contained_condition_percentages` instead.
- Additionally, ammunition boxes have been changed to work similar to other consumables, in that the condition change on use is always -1, and the number of shots is determined by the max condition instead.

[Submarine.zip](https://github.com/Regalis11/Barotrauma/files/10449407/Submarine.zip)
- A testing area for the above changes.


Barotrauma/BarotraumaShared/SharedSource/Items/Components/ItemContainer.cs
```diff
@@ -396,7 +394,13 @@ public void OnItemRemoved(Item containedItem)
...
             drawableContainedItems.RemoveAll(i => i.Item == containedItem);
             //deactivate if the inventory is empty
-            IsActive = activeContainedItems.Count > 0 || Inventory.AllItems.Any(it => it.body != null);
+            IsActive = Inventory.AllItems.Any();
+            if (!IsActive)
+            {
+                item.SendSignal("0", "contained_conditions");
+                item.SendSignal("0", "contained_condition_percentages");
+                item.SendSignal("0", "contained_items");
+            }
```
Regalis11, line 403: I think this should be optimized by not forcing the container to stay active if it doesn't have any of these connections.

Barotrauma/BarotraumaShared/SharedSource/Items/Components/ItemContainer.cs
```diff
@@ -497,8 +517,11 @@ public override void Update(float deltaTime, Camera cam)
             {
                 SetContainedItemPositions();
             }
-            else if (activeContainedItems.Count == 0)
+            else if (!Inventory.AllItems.Any())
             {
+                item.SendSignal("0", "contained_conditions");
+                item.SendSignal("0", "contained_conditions_percentage");
+                item.SendSignal("0", "contained_items");
```
Regalis11, line 524: Duplicate code, this and the SendSignal calls done in OnItemRemoved could be consolidated into one method.

Barotrauma/BarotraumaShared/SharedSource/Items/Components/ItemContainer.cs
```diff
@@ -458,6 +462,22 @@ public override void Update(float deltaTime, Camera cam)
...
+                if (item.Condition != 0)
+                {
+                    totalConditionValue += item.Condition;
+                    totalConditionPercentage += item.ConditionPercentage;
+                    totalItems++;
+                }
+            }
+            item.SendSignal(totalConditionValue.ToString(), "contained_conditions");
+            item.SendSignal(totalConditionPercentage.ToString(), "contained_condition_percentages");
+            item.SendSignal(totalItems.ToString(), "contained_items");
```
Regalis11, line 479: The ToString calls need to use invariant culture, otherwise the values will be different depending on the culture.

Barotrauma/BarotraumaShared/SharedSource/Items/Components/Machines/Reactor.cs
```diff
@@ -376,7 +376,6 @@ static float adjustValueWithoutOverShooting(float current, float target, float s
             item.SendSignal(((int)-CurrPowerConsumption).ToString(), "power_value_out");
             item.SendSignal(((int)Load).ToString(), "load_value_out");
             item.SendSignal(((int)AvailableFuel).ToString(), "fuel_out");
-            item.SendSignal(((int)fuelLeft).ToString(), "fuel_percentage_left");
```
Regalis11, line 379: This should not be removed, because it breaks backwards compatibility with items that still use this connection.

Regalis11 (changes requested):
Found some issues, see comments

Barotrauma/BarotraumaShared/SharedSource/Items/Components/Machines/Reactor.cs
Jade-Harleyy: Re-added.

Barotrauma/BarotraumaShared/SharedSource/Items/Components/ItemContainer.cs
Jade-Harleyy: Fixed.

Barotrauma/BarotraumaShared/SharedSource/Items/Components/ItemContainer.cs
Jade-Harleyy: Made some changes while working on optimization, now the three SendSignal() calls are only made once.

Barotrauma/BarotraumaShared/SharedSource/Items/Components/ItemContainer.cs
Jade-Harleyy: Changed, now the component only stays active if it has at least one of the three connections, or is containing active items or items with a physics body (same as before).

Regalis11:
There are still some more issues here:
- Unused using directives (no need to use Steamworks.Ugc here)
- Lots of duplicate code (the signal check should be consolidated into one place so we don't need to remember to update every single place it's defined in if we e.g. change something about the names of the connections).
- There is no need to check whether the item has the connections constantly, it could be done once after loading the item.
- The signal logic (calculating the conditions and such) runs even if the item doesn't have any of the outputs.
- `item.Condition != 0.0f` - floating point comparisons should not be done using `=` or `!=`. Due to floating point inaccuracies, the value can be practically 0 but not exactly bit-wise equal.

Regalis11 (changes requested):
There are still some issues here: I think the biggest one is that blank loaders have both the "contained_conditions" output and the "contained_items" output. Swapping one of those to a loader that only has one of those would presumably disconnect the wiring.

Barotrauma/BarotraumaShared/SharedSource/Items/Components/ItemContainer.cs
```diff
@@ -3,10 +3,14 @@
...
 using System.Linq;
 using System.Xml.Linq;
+using System.Globalization;
 using Barotrauma.Extensions;
 using FarseerPhysics;
 using System.Collections.Immutable;
 using Barotrauma.Abilities;
+using System.Data.Common;
+using FarseerPhysics.Common.PhysicsLogic;
+using Steamworks.Ugc;
```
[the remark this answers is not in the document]
Jade-Harleyy: Fixed.

Barotrauma/BarotraumaShared/SharedSource/Items/Components/ItemContainer.cs
```diff
@@ -375,8 +379,8 @@ public void OnItemContained(Item containedItem)
                 GameAnalyticsManager.AddDesignEvent("MicroInteraction:" + (GameMain.GameSession?.GameMode?.Preset.Identifier.Value ?? "null") + ":GardeningPlanted:" + containedItem.Prefab.Identifier);
             }
 
-            //no need to Update() if this item has no statuseffects and no physics body
-            IsActive = activeContainedItems.Count > 0 || Inventory.AllItems.Any(it => it.body != null);
+            // No need to Update() if the new item has no statuseffects or a physics body, and if there are no signal connections.
+            IsActive = (item.Connections?.Any(c => c.Name is "contained_conditions" or "contained_conditions_percentage" or "contained_items") ?? false) || activeContainedItems.Count > 0 || Inventory.AllItems.Any(it => it.body != null);
```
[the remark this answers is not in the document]
Jade-Harleyy: Changed, now we check the connections once in OnItemLoaded and read the saved value everywhere else.

Barotrauma/BarotraumaShared/SharedSource/Items/Components/ItemContainer.cs
```diff
@@ -461,6 +465,22 @@ public override void Update(float deltaTime, Camera cam)
                 alwaysContainedItemsSpawned = true;
             }
 
+            float totalConditionValue = 0;
+            float totalConditionPercentage = 0;
+            int totalItems = 0;
+            foreach (var item in Inventory.AllItems)
+            {
+                if (item.Condition != 0)
```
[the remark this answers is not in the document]
Jade-Harleyy: Fixed with the above comment, all values are now integers.

Barotrauma/BarotraumaShared/SharedSource/Items/Components/ItemContainer.cs
```diff
@@ -461,6 +465,22 @@ public override void Update(float deltaTime, Camera cam)
...
+                {
+                    totalConditionValue += item.Condition;
+                    totalConditionPercentage += item.ConditionPercentage;
+                    totalItems++;
+                }
+            }
+            item.SendSignal(totalConditionValue.ToString(CultureInfo.InvariantCulture), "contained_conditions");
+            item.SendSignal(totalConditionPercentage.ToString(CultureInfo.InvariantCulture), "contained_conditions_percentage");
+            item.SendSignal(totalItems.ToString(CultureInfo.InvariantCulture), "contained_items");
```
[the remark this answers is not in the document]
Jade-Harleyy: Fixed; now we first check if the current and previous values are different, and only then update the strings.

Jade-Harleyy:
> There are still some issues here: I think the biggest one is that blank loaders have both the "contained_conditions" output and the "contained_items" output. Swapping one of those to a loader that only has one of those would presumably disconnect the wiring.

I have tested this in-game and it seems to preform how we want it to:
- Wires stay on the same connection _name_ when switching _to_ a blank loader.
- Wires stay on the same connection _pin_ when switching _from_ a blank loader.

mygamingaccount:
Would love to have this on detonators!

Barotrauma/BarotraumaShared/SharedSource/Items/Components/ItemContainer.cs
```diff
@@ -461,6 +467,42 @@ public override void Update(float deltaTime, Camera cam)
                 alwaysContainedItemsSpawned = true;
             }
 
+            if (hasSignalConnections)
+            {
+                int totalConditionValue = 0, totalConditionPercentage = 0, totalItems = 0;
+                foreach (var item in Inventory.AllItems)
+                {
+                    if (item.Condition != 0)
```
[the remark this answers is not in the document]
Jade-Harleyy: Fixed, all floating point comparisons should be using `MathUtils.NearlyEqual` now.

Barotrauma/BarotraumaShared/SharedSource/Items/Components/ItemContainer.cs
```diff
@@ -461,6 +467,42 @@ public override void Update(float deltaTime, Camera cam)
...
+                    {
+                        totalConditionValue += (int)item.Condition;
+                        totalConditionPercentage += (int)item.ConditionPercentage;
```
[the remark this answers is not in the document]
Jade-Harleyy: Fixed. Condition values now stay as floats until they are converted to strings.

Regalis11:
Finally found some time to take another look at this! Can't see any further issues with the changes, moved to a PR in our private repo https://github.com/FakeFishGames/Barotrauma-development/pull/5251. Thank you for the contribution! :)

PR #11084, opened 2023-01-29
Unhide hidden items/assemblies in the Submarine Editor with DebugDraw

Jade-Harleyy:
This commit makes visible hidden items and item assemblies in the Submarine Editor while debugdraw is enabled.

This allows submarine builders to be more creative, while still protecting newer players from veiwing spoilers (see branch: faction-test).

The only drawback is that the items are not immediately visible; the Entity Menu must be "reloaded" by selecting any category.


Regalis11:
Thank you for the contribution! Merged to our private repo in https://github.com/Regalis11/Barotrauma-development/commit/0676015dc1368b3712a94ccca01ee90a8346ea3e

Issue #11136, opened 2023-02-02
[Factions] Hostile faction vessels are either incredibly rare or don't spawn


Wheatley1001:
### Disclaimers

- [X] I have searched the issue tracker to check if the issue has already been reported.
- [ ] My issue happened while using mods.

### What happened?

Throughout the entirety of Season 1 and 2 of the testing sessions, we never experienced a single hostile vessel attacking us. For the vast majority of gameplay, we were at -100 reputation for the Separatists (season 1) and the Coalition (season 2). We were never randomly attacked by either side as the code suggests we should've. 

![image](https://user-images.githubusercontent.com/105495756/216460995-81d50dff-7c78-4452-95c8-4f1cee2086ff.png)
_Figure 1. Reference in the code of these mission types existing._

### Bug prevalence

Just once

### Version

Faction/endgame test branch

### Which operating system did you encounter this bug on?

Windows

Jade-Harleyy:
Upon further testing, I was able to get an ambush missions to spawn by following these steps:
1. Create a new campaign with the 'hard' preset.
2. Enter the following commands: `enablecheats` `setreputation coalition -100` `setreputation separatists -100`.
3. Open the map and select any destination, then undock from the starting outpost.
4. Use the `freecam` command to check if there is an enemy submarine in the level.
5. If not, redock to the starting outpost and repeat steps 3 through 5.

Image of spawned Coalition submarine:
![image](https://user-images.githubusercontent.com/67431770/216552890-f07c82f1-d1c5-41cd-945b-799fb27c3593.png)

Image of the mission text and Coalition submarine on sonar:
![image](https://user-images.githubusercontent.com/67431770/216554092-d97258e8-ac48-40c9-80ca-87aa435e24ad.png)

JeroonBosch:
Temporarily set minprobability="0.5" and tested it. Spawned a hostile Separatist sub on the 2nd route I tried at a 2-passageway distance, so seems to work as intended. The route I took was between two Coalition stations though, so players may find that odd.

Regalis11:
Addressed in https://github.com/Regalis11/Barotrauma-development/commit/efccf55fb854d03fd86a55a3fd53b3ae1e1eac39: now the attacks don't happen in connections between outposts that belong to another faction (i.e. you're safe from separatist attacks when traveling from a coalition outpost to another, or vice versa).

JeroonBosch:
[...]

PR #11140, opened 2023-02-02
Fix Issue #11136 - "[Factions] Hostile faction vessels are either incredibly rare or don't spawn"

Jade-Harleyy:
This PR fixes issue #11136 by removing an unnecessary(?) check for mission prefabs of type "Pirate" in the AddExtraMissions method.


Regalis11:
This check is not unnecessary. It prevents adding an extra pirate mission to the level if there's already a pirate mission active. Having multiple ones active at a time will lead to various issues.

PR #11264, opened 2023-02-16
Implement better semi-automatic weapons.

Jade-Harleyy:
This commit makes use of the `holdtrigger` attribute of the `RangedWeapon` item component to force players to release their "Shoot" binding to be able to fire the weapon again if it is set to `true`.

Additionally, `PlayerInput.cs` has been edited to use different mouse input methods, as before the weapon would only fire once mouse buttons were released, instead of when they were first pressed.


pantherpusher:
this'd be really nice to have, as currently everything is automatic.

Regalis11:
Even though this feels like a somewhat useful addition, I think there's a major issue here: This changes how mouse inputs are handled everywhere in the game (basically making a mouse click mean pressing the key down, as opposed to pressing it and then releasing). I don't think it's a good idea, nor safe, to change such a core part of input handling for such a minor feature.

PR #11303, opened 2023-02-21
GreaterComponent support for multiple signals within a timeframe

mygamingaccount:
This change would allow us to make simple circuits that compare multiple signals to see if any of them is greater or less than one or many signals.

It is a 100% backwards compatible way of allowing players to use Min(a,b) and Max(a,b) when comparing values.

The description could be modified to better reflect this change, although this is not necessary since the change does not alter existing behavior.
_Original description:_
> Sends a signal if the value the signal_in1 input is larger than the signal_in2 input.

_New description_
> Sends a signal if signal_in1 receives a value that is larger than values sent to signal_in2 within the timeframe.

**Example**
the circuit which compares conditions to show if any of the devices need servicing.
**Before**, using makeshift min(a,b):
image

**After:**
image

### DIFF
--- a/Barotrauma/BarotraumaShared/SharedSource/Items/Components/Signal/GreaterComponent.cs
+++ b/Barotrauma/BarotraumaShared/SharedSource/Items/Components/Signal/GreaterComponent.cs
@@ -5,7 +5,8 @@ namespace Barotrauma.Items.Components
 {
     class GreaterComponent : EqualsComponent
     {
-        private float val1, val2;
+        private float val1 = float.NegativeInfinity;
+        private float val2 = float.PositiveInfinity;
 
         public GreaterComponent(Item item, ContentXElement element)
             : base(item, element)
@@ -25,6 +26,8 @@ namespace Barotrauma.Items.Components
             if (sendOutput)
             {
                 string signalOut = val1 > val2 ? output : falseOutput;
+                val1 = float.NegativeInfinity;
+                val2 = float.PositiveInfinity;
                 if (string.IsNullOrEmpty(signalOut)) return;
 
                 item.SendSignal(signalOut, "signal_out");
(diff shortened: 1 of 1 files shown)


Saevon:
This would nicely let "greater than" support both MIN and MAX. Now if only we had a good guide that reminds you how to make common simple logic/math circuits using this system (in-game)

Barotrauma/BarotraumaShared/SharedSource/Items/Components/Signal/GreaterComponent.cs
```diff
@@ -5,7 +5,8 @@ namespace Barotrauma.Items.Components
 {
     class GreaterComponent : EqualsComponent
     {
-        private float val1, val2;
+        private float val1 = float.NegativeInfinity;
+        private float val2 = float.PositiveInfinity;
```
Regalis11, line 9: In my opinion the use of these sentinel values isn't the clearest way to go about this. I think it'd make the intention clearer if these were for example nullable floats, where null represents "no value received".

Barotrauma/BarotraumaShared/SharedSource/Items/Components/Signal/GreaterComponent.cs
```diff
@@ -25,6 +26,8 @@ public override void Update(float deltaTime, Camera cam)
             if (sendOutput)
             {
                 string signalOut = val1 > val2 ? output : falseOutput;
+                val1 = float.NegativeInfinity;
+                val2 = float.PositiveInfinity;
```
Regalis11, line 30: As far as I can see, this breaks backwards compatibility. It means the component will not keep sending the same output, but reverts back to false immediately after it's sent one output.

Barotrauma/BarotraumaShared/SharedSource/Items/Components/Signal/GreaterComponent.cs
```diff
@@ -33,15 +36,18 @@ public override void Update(float deltaTime, Camera cam)
...
+                    float signal1 = float.NegativeInfinity;
+                    float.TryParse(signal.value, NumberStyles.Float, CultureInfo.InvariantCulture, out signal1);
+                    val1 = signal1 > val1 ? signal1 : val1;
                     timeSinceReceived[0] = 0.0f;
                     break;
                 case "signal_in2":
-                    float.TryParse(signal.value, NumberStyles.Float, CultureInfo.InvariantCulture, out val2);
+                    float signal2 = float.PositiveInfinity;
+                    float.TryParse(signal.value, NumberStyles.Float, CultureInfo.InvariantCulture, out signal2);
+                    val2 = signal2 < val2 ? signal2 : val2;
```
Regalis11, line 50: I believe this breaks backwards compatibility too. Previously the last received value was used, now the min/max value is used.

Regalis11 (changes requested):
Thank you for the PR, and sorry for taking so long to take a look at this! I think this would be an useful change, but to me it seems like there's a handful of issues here - most importantly, I think this'd cause backwards compatibility issues because it changes the behavior of greater components in existing circuits too.

Regalis11:
Does not seem to be in a mergeable state, closing at least for now

PR #11309, opened 2023-02-22
Sonar Refactor

Jade-Harleyy:
This PR allows modders to change the interval between active sonar pings.


mygamingaccount:
The interval should instead depend on the selected distance, no? You can change the strength of the ping, but not the speed at which it travels through the water.

Regalis11:
> The interval should instead depend on the selected distance, no? You can change the strength of the ping, but not the speed at which it travels through the water.

The interval just means the interval at which a ping is emitted, not the speed at which the emitted ping travels. There should be no limit to the frequency, nor should it depend on the distance.

Regalis11 (changes requested):
The speed at which the pings travel is currently relative to the frequency:

`activePings[pingIndex].State += deltaTime * PingFrequency;`

This needs to be changed if the frequency is made editable.

mygamingaccount:
> The interval just means the interval at which a ping is emitted, not the speed at which the emitted ping travels. There should be no limit to the frequency, nor should it depend on the distance.

Right, but if you reduce the interval between transmissions, that will reduce the range, since when you have two SONAR pings spreading out, then you won't know which one you're getting an echo from on the receiver.

So the range of the active sonar with relation to the interval is `R = 1/2 * c/Δt ` where c is the speed of sound, and Δt is the ping interval

Regalis11:
> Right, but if you reduce the interval between transmissions, that will reduce the range, since when you have two SONAR pings spreading out, then you won't know which one you're getting an echo from on the receiver.

Wouldn't this be trivial to work around by emitting pings of slightly different frequencies, so you could deduce which ping it was based on the frequency of the echo?

mygamingaccount:
> Wouldn't this be trivial to work around by emitting pings of slightly different frequencies, so you could deduce which ping it was based on the frequency of the echo?

Perhaps with wildly different frequencies. The narrower the pulse, the wider its spectrum, so it might interfere with the other ping and its reflections if they are similar.
The book I have on sonars elegantly avoids this question of multiple simultaneous transmissions.

Jade-Harleyy:
oops

Barotrauma/BarotraumaShared/SharedSource/Items/Components/Machines/Sonar.cs
```diff
@@ -94,6 +82,27 @@ public float Range
...
+            get { return pingInterval; }
+            set { pingInterval = MathHelper.Max(value, 0.01f); }
+        }
+
+        [Serialize(5000f, IsPropertySaveable.No, description: "The speed at which sonar pings travel.")]
+        public float PingSpeed
+        {
+            get { return pingSpeed; }
+            set { pingSpeed = MathHelper.Max(value, 0.01f); }
+        }
```
Regalis11, line 97: Imo it doesn't make sense for this to be editable, the pings should travel at a fixed speed (realistically should be the speed of sound, even though it's actually much slower in-game).

Barotrauma/BarotraumaShared/SharedSource/Items/Components/Machines/Sonar.cs
```diff
@@ -183,73 +192,72 @@ public override void Update(float deltaTime, Camera cam)
...
+                    }
+
+                    if (item.AiTarget != null)
+                    {
+                        item.AiTarget.SoundRange = Math.Max(item.AiTarget.SoundRange, MathHelper.Lerp(item.AiTarget.MinSoundRange, item.AiTarget.MaxSoundRange, activePing.Radius / Range));
                     }
                 }
-                else
+
+                // TODO: Allow multiple seperated sectors & iterate over each active ping to disallow pinging a sector then moving the ping direction to gain information without aggro'ing enemies.
```
Regalis11, line 250: Unaddressed todo note. Now the sector seems to be based on the latest ping, which could potentially be abused.

Regalis11 (changes requested):
Found some more issues (the speed being editable isn't really a major problem, but I think the way the AITarget's sector behaves is). All in all I'm starting to get the feeling this is turning to be such an extensive rewrite, for a feature that seems quite niche, that I'm starting to question whether it's worthwhile to do. If it was only possible to apply a delay between pulses, not freely edit the frequency (which means there can be multiple pings active at a time), I think it'd significantly simplify the logic and only take a handful of changes.

Regalis11:
We feel like this is a too extensive rewrite (which would be very time-consuming to test and could cause hard-to spot side-effect) to address a very niche issue, so I don't think we'll be accepting the PR in this form. Thank you for your contribution regardless!

Issue #11320, opened 2023-02-23
ExperiencePoints server/client 


plaksy:
### Disclaimers

- [X] I have searched the issue tracker to check if the issue has already been reported.
- [ ] My issue happened while using mods.

### What happened?

My friend when get more than 17 talent point, when ExperiencePoints  goes over 65536, so he cannot add new talents and get new levels from start.
We think, the problem is with short int (int16) while server sends client info, so, maybe if we change it to int32. it will be good.
In file **CharacterInfo.cs**:
Change **msg.WriteUInt16((ushort)ExperiencePoints);** to **msg.WriteUInt32((uint)ExperiencePoints);**
I hope it will help.

### Reproduction steps

1. Start multiplayer game and get more than 65536 total ExperiencePoints.
2. After your character cannot get talents.

### Bug prevalence

Just once

### Version

0.21.6.0

### Which operating system did you encounter this bug on?

Windows

Wheatley1001:
Several other people have also reported this issue in the #baro-questions-and-help channel on the discord. Two reports were regarding the security class specifically, though it could relate to all the classes.

IhorBand:
> Several other people have also reported this issue in the #baro-questions-and-help channel on the discord. Two reports were regarding the security class specifically, though it could relate to all the classes.

I fixed that issue in PR(You can find link below). We already did a small test with friends and looks like everything is OK. 
Thank you for your attention 😊
Have a lovely day.😁

https://github.com/Regalis11/Barotrauma/pull/11330

Regalis11:
Fixed in https://github.com/Regalis11/Barotrauma-development/pull/4022/commits/ea2cb10ffee7eecc7b9fd5ac2c2a418da94d3d1c

PR #11330, opened 2023-02-23
Fix for data transferring between server and client Experience property

IhorBand:
Fix for issue #11320 .
My friend got 72000 experience and now he cannot get more talents for his character anymore. I found this code and looks like 72000 converts to ushort and my friend receives 6464.


plaksy:
Xd👍

Regalis11:
Thank you for the PR! We've fixed this in our private repo, so this is no longer needed. We also went with a slightly different approach: there is no need to use signed integers here because the experience can never be negative.

PR #11361, opened 2023-02-25
[Requires Localization] Add max level objects to graphics settings.

Jade-Harleyy:
This commit adds a setting to the graphics settings menu that allows users to change the max amount of level objects.

Localization is required for the `LevelObjectLimit` and `LevelObjectLimitTooltip` tags.


Regalis11:
I don't think making this an adjustable setting is a good idea. We can't have level objects (some of which are hazards the players need to see and be able to react to) become invisible when the player is near them - the default limit should only kick in in situations where the camera is zoomed out extremely far, not during normal gameplay.

PR #11378, opened 2023-02-26
Center sonar AITarget on sub

GoodDays13:
If there aren't any transducers, center on the sub's position, if there are use the transducers

### DIFF
--- a/Barotrauma/BarotraumaShared/SharedSource/Characters/AI/AITarget.cs
+++ b/Barotrauma/BarotraumaShared/SharedSource/Characters/AI/AITarget.cs
@@ -18,7 +18,8 @@ namespace Barotrauma
                 return entity;
             }
         }
-        
+
+        private Entity entityOveride;
         private float soundRange;
         private float sightRange;
 
@@ -31,6 +32,15 @@ namespace Barotrauma
         public bool StaticSound { get; private set; }
         public bool StaticSight { get; private set; }
         
+        public Entity EntityOveride
+        {
+            get { return entityOveride; }
+            set
+            {
+                entityOveride = value;
+            }
+        }
+
         public float SoundRange
         {
             get { return soundRange; }
@@ -161,6 +171,8 @@ namespace Barotrauma
                     return Vector2.Zero;
                 }
 
+                if (entityOveride != null)
+                    return entityOveride.WorldPosition;
                 return entity.WorldPosition;
             }
         }
@@ -180,6 +192,8 @@ namespace Barotrauma
                     return Vector2.Zero;
                 }
 
+                if (entityOveride != null)
+                    return entityOveride.SimPosition;
                 return entity.SimPosition;
             }
         }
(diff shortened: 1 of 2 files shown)


GoodDays13:
I would like the opinion of someone more experienced on the way I'm doing this so far and if it would screw anything up.

mygamingaccount:
Is this the same thing Regalis said they won't fix?

GoodDays13:
> Is this the same thing Regalis said they won't fix?

Eh. It seemed more to me that they wanted the sonar interface centered on the sub, which I didn't change here.

Regalis11:
> Eh. It seemed more to me that they wanted the sonar interface centered on the sub, which I didn't change here.

Yeah pretty much: although I do consider this a very low-prio issue, because I think it can only be exploited in some very, very specific situations, and if you've specifically built the sub to take advantage of the discrepancy between the sonar origin and the nav terminal's position.

In either case, I don't think the issue can be fixed like this, because it'd mean any AI character targeting the nav terminal's AITarget (let's say, some monster inside the sub that tries to destroy the device) would think the terminal is located at the sub's origin, and end up attacking thin air.

GoodDays13:
> > Eh. It seemed more to me that they wanted the sonar interface centered on the sub, which I didn't change here.
> 
> Yeah pretty much: although I do consider this a very low-prio issue, because I think it can only be exploited in some very, very specific situations, and if you've specifically built the sub to take advantage of the discrepancy between the sonar origin and the nav terminal's position.
> 
> In either case, I don't think the issue can be fixed like this, because it'd mean any AI character targeting the nav terminal's AITarget (let's say, some monster inside the sub that tries to destroy the device) would think the terminal is located at the sub's origin, and end up attacking thin air.

I will test this when I can but do enemies try to attack the nav terminal? Can they damage it?

Regalis11:
To my knowledge, nothing in the vanilla game does, but there could be mods in which they do. And it's worth noting that the changes here don't apply to just the nav terminal, but any item with a Sonar component, which includes handheld sonars (and again, many kinds of modded items). Pretending the item is located at the center of the sub is even more problematic when it comes to that sort of items.

Overriding the position also causes problems with the visibility of the nav terminal: a monster standing right next to it wouldn't be able to see it if. I also don't think it's a good idea to make the position default to the position of the first connected transducer: it should probably be the average position of the transducers instead (and as pointed out by your comment, that's a little tricky to do with this approach).

All in all, I think this isn't a trivial problem to solve without breaking backwards compatibility and side-effects that may be even worse than the original issue. Which is another reason why I've been reluctant to start addressing it, but I appreciate your effort in trying to come up with a workable solution!

Regalis11:
>would it be possible to make the nav terminal have a smaller target just for the audible pings a player or enemy could hear and then have a new separate entity have an aitarget that can only be seen when outside the sub 

Currently no, each entity can only have one AITarget. It could be changed, but most likely requires quite a lot of refactoring.

> is it possible to give the sub's center itself an aitarget?

This isn't possible either atm, only entities like items, structures and hulls (basically the parts a submarine consists of), and characters can have AITargets.

PR #11690, opened 2023-03-28
Improve textbox control navigation

amadare42:
Improving navigations withing all textbox controls.

Specifically:

- <kbd>Ctrl</kbd> + <kbd>Left</kbd> moves caret to beginning of current **word**
- <kbd>Ctrl</kbd> + <kbd>Right</kbd> moves caret to beginning of next **word**
- <kbd>Ctrl</kbd> + <kbd>Delete</kbd> removes characters starting from carret position to start of next **word**
- <kbd>Ctrl</kbd> + <kbd>Backspace</kbd>  removes characters starting from carret position to start of current **word**
- <kbd>Home</kbd> moves caret to start of current line
- <kbd>Ctrl</kbd> + <kbd>Home</kbd> moves caret to start of whole text
- <kbd>Ctrl</kbd> + <kbd>End</kbd> moves caret to the end of whole text
- <kbd>End</kbd> moves carent to end of current line
- Holding <kbd>Shift</kbd> enables selecting mode so selection can be made by keystrokes <kbd>Shift</kbd>+<kbd>Left/Right</kbd> and even by <kbd>Shift</kbd>+<kbd>Ctrl</kbd>+<kbd>Left/Right</kbd> according to rules defined above

**word** is defined as sequence of characters that is surrounded from both sides by one of following
- punctuation characters (`char.IsPunctuation()`)
- whitespace
- newline character(`\n`)
- start of string
- end of string

### DIFF
--- a/Barotrauma/BarotraumaClient/ClientSource/GUI/GUITextBox.cs
+++ b/Barotrauma/BarotraumaClient/ClientSource/GUI/GUITextBox.cs
@@ -420,7 +420,7 @@ namespace Barotrauma
                 }
                 else
                 {
-                    isSelecting = PlayerInput.PrimaryMouseButtonHeld();
+                    isSelecting = PlayerInput.PrimaryMouseButtonHeld() || PlayerInput.IsShiftDown();
                 }
                 if (PlayerInput.DoubleClicked())
                 {
@@ -617,10 +617,15 @@ namespace Barotrauma
                 case '\b' when !Readonly: //backspace
                 {
                     using var _ = new TextPosPreservation(this);
-                    if (PlayerInput.KeyDown(Keys.LeftControl) || PlayerInput.KeyDown(Keys.RightControl))
+                    if (PlayerInput.IsCtrlDown())
                     {
-                        SetText(string.Empty, false);
-                        CaretIndex = Text.Length;
+                        var removeTo = FindStopSymbolLeft(Text, CaretIndex);
+                        var len = CaretIndex - removeTo;
+                        var newText = Text.Remove(removeTo, CaretIndex - removeTo);
+                        SetText(newText, false);
+                        CaretIndex = Math.Max(0, CaretIndex - len);
+                        CalculateCaretPos();
+                        ClearSelection();
                     }
                     else if (selectedCharacters > 0)
                     {
@@ -706,7 +711,16 @@ namespace Barotrauma
                     {
                         InitSelectionStart();
                     }
-                    CaretIndex = Math.Max(CaretIndex - 1, 0);
+
+                    if (PlayerInput.IsCtrlDown())
+                    {
+                        CaretIndex = FindStopSymbolLeft(Text, CaretIndex);
+                    }
+                    else
+                    {
+                        CaretIndex = Math.Max(CaretIndex - 1, 0);
+                    }
+
                     caretTimer = 0;
                     HandleSelection();
                     break;
@@ -715,7 +729,16 @@ namespace Barotrauma
                     {
                         InitSelectionStart();
                     }
-                    CaretIndex = Math.Min(CaretIndex + 1, Text.Length);
+
+                    if (PlayerInput.IsCtrlDown())
+                    {
+                        CaretIndex = FindStopSymbolRight(Text, CaretIndex);
+                    }
+                    else
+                    {
+                        CaretIndex = Math.Min(CaretIndex + 1, Text.Length);
+                    }
+
                     caretTimer = 0;
                     HandleSelection();
                     break;
@@ -756,7 +779,27 @@ namespace Barotrauma
                     HandleSelection();
                     break;
                 case Keys.Delete when !Readonly:
-                    if (selectedCharacters > 0)
+                    if (PlayerInput.IsCtrlDown())
+                    {
+                        var removeTo = FindStopSymbolRight(Text, CaretIndex);
+                        var len = removeTo - CaretIndex;
+                        if (len <= 0)
+                        {
+                            if (selectedCharacters > 0)
+                            {
+                                RemoveSelectedText();
+                            }
+
+                            break;
+                        }
+
+                        var newText = Text.Remove(CaretIndex, len);
+                        SetText(newText, false);
+                        OnTextChanged?.Invoke(this, Text);
+                        CalculateCaretPos();
+                        ClearSelection();
+                    }
+                    else if (selectedCharacters > 0)
                     {
                         RemoveSelectedText();
                     }
@@ -812,6 +855,52 @@ namespace Barotrauma
                         return null;
                     }
                     break;
+
+                case Keys.Home:
+                    if (isSelecting)
+                    {
+                        InitSelectionStart();
+                    }
+
+                    if (PlayerInput.IsCtrlDown())
+                    {
+                        CaretIndex = 0;
+                    }
+                    else
+                    {
+                        CaretIndex = Text.LastIndexOf('\n', CaretIndex - 1) + 1;
+                    }
+
+                    caretTimer = 0;
+                    HandleSelection();
+                    break;
+
+                case Keys.End:
+                    if (isSelecting)
+                    {
+                        InitSelectionStart();
+                    }
+
+                    if (PlayerInput.IsCtrlDown())
+                    {
+                        CaretIndex = Text.Length;
+                    }
+                    else
+                    {
+                        var idx = Text.IndexOf('\n', CaretIndex);
+                        if (idx < 0)
+                        {
+                            CaretIndex = Text.Length;
+                        }
+                        else
+                        {
+                            CaretIndex = idx;
+                        }
+                    }
+
+                    caretTimer = 0;
+                    HandleSelection();
+                    break;
             }
             if (caretPosDirty) { CalculateCaretPos(); }
             OnKeyHit?.Invoke(this, key);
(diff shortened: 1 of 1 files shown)


Barotrauma/BarotraumaClient/ClientSource/GUI/GUITextBox.cs
```diff
@@ -756,7 +779,27 @@ public void ReceiveSpecialInput(Keys key)
...
+                            if (selectedCharacters > 0)
+                            {
+                                RemoveSelectedText();
+                            }
+
+                            break;
+                        }
+
+                        var newText = Text.Remove(CaretIndex, len);
+                        SetText(newText, false);
```
itchyOwl, line 797: Should be `SetText(newText, store: true)`, or undo/redo memento won't register this.

Barotrauma/BarotraumaClient/ClientSource/GUI/GUITextBox.cs
```diff
@@ -756,7 +779,27 @@ public void ReceiveSpecialInput(Keys key)
...
+                        OnTextChanged?.Invoke(this, Text);
+                        CalculateCaretPos();
```
itchyOwl, line 799: Caret position seems to be incorrect after undoing and redoing.

Barotrauma/BarotraumaClient/ClientSource/GUI/GUITextBox.cs
```diff
@@ -617,10 +617,15 @@ public void ReceiveCommandInput(char command)
...
                     using var _ = new TextPosPreservation(this);
-                    if (PlayerInput.KeyDown(Keys.LeftControl) || PlayerInput.KeyDown(Keys.RightControl))
+                    if (PlayerInput.IsCtrlDown())
                     {
-                        SetText(string.Empty, false);
-                        CaretIndex = Text.Length;
+                        var removeTo = FindStopSymbolLeft(Text, CaretIndex);
+                        var len = CaretIndex - removeTo;
+                        var newText = Text.Remove(removeTo, CaretIndex - removeTo);
+                        SetText(newText, false);
```
itchyOwl, line 625: Should be `SetText(newText, store: true)`, or undo/redo memento won't register this.

Barotrauma/BarotraumaClient/ClientSource/GUI/GUITextBox.cs
```diff
@@ -617,10 +617,15 @@ public void ReceiveCommandInput(char command)
...
+                        CaretIndex = Math.Max(0, CaretIndex - len);
+                        CalculateCaretPos();
```
itchyOwl, line 627: Caret position seems to be incorrect after undoing and redoing.

Barotrauma/BarotraumaClient/ClientSource/GUI/GUITextBox.cs
```diff
@@ -838,6 +927,53 @@ public void SelectAll()
...
+            // special case: if cursor at the start of the line, move to the end of previous line
+            if (idx > 0 && text[idx - 1] == '\n')
+            {
+                return idx - 1;
+            }
+            // convert cursor index to character index
+            idx = Math.Max(0, idx - 2);
+            if (idx == 0) return idx;
+
+            for (; idx >= 0; idx--)
```
itchyOwl, line 941: This is a slightly unconventional loop structure. Consider replacing with 
```
while (idx >=0)
{
   idx--;
   // Do stuff
}
```

Barotrauma/BarotraumaClient/ClientSource/GUI/GUITextBox.cs
```diff
@@ -838,6 +927,53 @@ public void SelectAll()
...
+        private static int FindStopSymbolLeft(string text, int idx)
+        {
+            // special case: if cursor at the start of the line, move to the end of previous line
+            if (idx > 0 && text[idx - 1] == '\n')
+            {
+                return idx - 1;
+            }
+            // convert cursor index to character index
+            idx = Math.Max(0, idx - 2);
+            if (idx == 0) return idx;
```
itchyOwl, line 939:
```suggestion
            if (idx == 0) { return idx; }
```

Barotrauma/BarotraumaClient/ClientSource/GUI/GUITextBox.cs
```diff
@@ -838,6 +927,53 @@ public void SelectAll()
             CalculateSelection();
         }
 
+        private static int FindStopSymbolLeft(string text, int idx)
```
itchyOwl, line 930: Don't abbreviate: "index" instead of "idx"

Barotrauma/BarotraumaClient/ClientSource/GUI/GUITextBox.cs
```diff
@@ -617,10 +617,15 @@ public void ReceiveCommandInput(char command)
                 case '\b' when !Readonly: //backspace
                 {
                     using var _ = new TextPosPreservation(this);
-                    if (PlayerInput.KeyDown(Keys.LeftControl) || PlayerInput.KeyDown(Keys.RightControl))
+                    if (PlayerInput.IsCtrlDown())
                     {
-                        SetText(string.Empty, false);
-                        CaretIndex = Text.Length;
+                        var removeTo = FindStopSymbolLeft(Text, CaretIndex);
+                        var len = CaretIndex - removeTo;
```
itchyOwl, line 623: Don't abbreviate: "length" instead of "len".

Barotrauma/BarotraumaClient/ClientSource/GUI/GUITextBox.cs
```diff
@@ -756,7 +779,27 @@ public void ReceiveSpecialInput(Keys key)
                     HandleSelection();
                     break;
                 case Keys.Delete when !Readonly:
-                    if (selectedCharacters > 0)
+                    if (PlayerInput.IsCtrlDown())
+                    {
+                        var removeTo = FindStopSymbolRight(Text, CaretIndex);
+                        var len = removeTo - CaretIndex;
```
itchyOwl, line 785: Don't abbreviate: "length" instead of "len".

Barotrauma/BarotraumaClient/ClientSource/GUI/GUITextBox.cs
```diff
@@ -812,6 +855,52 @@ GUITextBox SelectNextTextBox(GUIListBox listBox)
...
+                        InitSelectionStart();
+                    }
+
+                    if (PlayerInput.IsCtrlDown())
+                    {
+                        CaretIndex = Text.Length;
+                    }
+                    else
+                    {
+                        var idx = Text.IndexOf('\n', CaretIndex);
```
itchyOwl, line 890: Don't abbreviate: "index" instead of "idx"

Barotrauma/BarotraumaClient/ClientSource/GUI/GUITextBox.cs
```diff
@@ -617,10 +617,15 @@ public void ReceiveCommandInput(char command)
                 case '\b' when !Readonly: //backspace
                 {
                     using var _ = new TextPosPreservation(this);
-                    if (PlayerInput.KeyDown(Keys.LeftControl) || PlayerInput.KeyDown(Keys.RightControl))
+                    if (PlayerInput.IsCtrlDown())
                     {
-                        SetText(string.Empty, false);
-                        CaretIndex = Text.Length;
+                        var removeTo = FindStopSymbolLeft(Text, CaretIndex);
```
itchyOwl, line 622: With simple types like `int` and `string,` using `var` just adds mental payload for the reader without any benefits. `var` can improve readability when used with generic types (e.g. `var list = new List<MyType>()`), but shouldn't be used when it's not clear from the context what the type is.

Barotrauma/BarotraumaClient/ClientSource/GUI/GUITextBox.cs
```diff
@@ -838,6 +927,53 @@ public void SelectAll()
...
+
+        private static int FindStopSymbolRight(string text, int idx)
+        {
+            // special case: if cursor is at the end of line, move to the start of next line
+            if (idx < text.Length && text[idx] == '\n')
+            {
+                return idx + 1;
+            }
+
+            for (; idx < text.Length; idx++)
```
itchyOwl, line 961: See the comment above.

Barotrauma/BarotraumaClient/ClientSource/GUI/GUITextBox.cs
```diff
@@ -838,6 +927,53 @@ public void SelectAll()
             CalculateSelection();
         }
 
+        private static int FindStopSymbolLeft(string text, int idx)
+        {
+            // special case: if cursor at the start of the line, move to the end of previous line
+            if (idx > 0 && text[idx - 1] == '\n')
```
itchyOwl, line 933: Accessing the string array without first checking whether the index is within the valid range will throw an index out of range exception with "incorrect" or unexpected parameters, like `FindStopSymbolLeft("Test", 100)`.

Barotrauma/BarotraumaClient/ClientSource/GUI/GUITextBox.cs
```diff
@@ -838,6 +927,53 @@ public void SelectAll()
...
+                    return idx + 1;
+                }
+            }
+            return Math.Clamp(idx, 0, text.Length);
+        }
+
+        private static int FindStopSymbolRight(string text, int idx)
+        {
+            // special case: if cursor is at the end of line, move to the start of next line
+            if (idx < text.Length && text[idx] == '\n')
```
itchyOwl, line 956: Accessing the string array without first checking whether the index is within the valid range will throw an index out of range exception with "incorrect" or unexpected parameters, like `FindStopSymbolRight("Test", 100)`.

itchyOwl (changes requested):
Found some issues, but in general the additions work as described. With some improvements to the code, I think we could merge this in.

Regalis11:
I also wonder whether this works correctly in languages where there are no spaces between words?

Barotrauma/BarotraumaClient/ClientSource/GUI/GUITextBox.cs
itchyOwl: Also note that it might be ok to crash there, if it shouldn't be that case that the index can be outside of the string array, but I think it's better to make that explicit: throw a custom exception or a console error with some debug info about the text and the index. Currently it looks to me that it just happens to work, because the text and the index are in sync, because we use the Text and TextCaret fields. And if we just use them, I'd maybe just remove the parameters from the methods, make them non-static, and use the fields. In that case, there wouldn't be need for any extra error handling.

Barotrauma/BarotraumaClient/ClientSource/GUI/GUITextBox.cs
amadare42: Thanks for review, @itchyOwl . 
Cursor position is always at the end of the input after undo/redo regardless of memento state before or after such action. I would agree that this is somewhat non-traditional behavior, but it will require non-trivial refactoring on how memento is working and focus of the PR was to add navigational shortcuts. I would prefer if this particual fix would be implemented in scope of different PR.

Here is a snippet of code for handling undo/redo that was not modified in PR:
```cs
case (char)0x1A when !Readonly && !SubEditorScreen.IsSubEditor(): // ctrl-z
    text = memento.Undo();
    if (text != Text)
    {
        ClearSelection();
        SetText(text, false);
        CaretIndex = Text.Length;
        OnTextChanged?.Invoke(this, Text);
    }
    break;
case (char)0x12 when !Readonly && !SubEditorScreen.IsSubEditor(): // ctrl-r
    text = memento.Redo();
    if (text != Text)
    {
        ClearSelection();
        SetText(text, false);
        CaretIndex = Text.Length;
        OnTextChanged?.Invoke(this, Text);
    }
    break;
```

Barotrauma/BarotraumaClient/ClientSource/GUI/GUITextBox.cs
itchyOwl: Not addressed.

itchyOwl (changes requested):
Found some issues with the latest implementation:

Writing "Test" and then ctr+backspace doesn't remove the word. Writing "Test this" and doing the same crashes, which I think is because `FindStopSymbolLeft() `now modifies` CaretIndex`.

image

image

Regalis11:
Does not seem to be in a mergeable state, closing

PR #12113, opened 2023-06-03
Dev Date&Time column in server browser

pavel4win:
It is very inconvenient that there is no way to sort the games played on the servers by the date and time of the last game.
I propose to introduce an additional column that will indicate the date and time of the last game on the server.
As well as the ability to sort the list of servers by the date/time column.


Regalis11:
Thank you for the suggestion! I presume you meant to post this in discussions, not create a pull request?

PR #12211, opened 2023-06-22
Reactor fuel percent left fix

Boolean-Buckeye:
Right now the reactor output signal "fuel percentage left" outputs a nonsensical range from 0-400%, it probably should be from 0-100%, so fix this. This makes measuring fuel efficiency of the reactor possible because you can then sample the fuel usage as (% fuel left) * (total fuel). With the current bug you can not do this because there is no way to make sense of how many rods are in the reactor, since for example two fulgarium fuel rods has the same "fuel out" as three thorium fuel rods, so there is no way to measure fuel effieciency, even with very complex circuits to try to calculate how many rods are in the reactor. With this bug fix, you would not even need the number of rods because you can know the actual fuel percentage left and compare it to the total fuel_out which is a unit of heat, in order to calculate what one single percent of fuel means, as opposed to the 0-400% range in which one single percent of fuel could mean anything.

### DIFF
--- a/Barotrauma/BarotraumaShared/SharedSource/Items/Components/Machines/Reactor.cs
+++ b/Barotrauma/BarotraumaShared/SharedSource/Items/Components/Machines/Reactor.cs
@@ -336,6 +336,7 @@ namespace Barotrauma.Items.Components
             var containedItems = item.OwnInventory?.AllItems;
             if (containedItems != null)
             {
+                int num_rods = 0;
                 foreach (Item item in containedItems)
                 {
                     if (!item.HasTag("reactorfuel")) { continue; }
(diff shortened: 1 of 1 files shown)


Regalis11:
Thank you for the PR!

The current behavior is not a bug: the connection intentionally outputs the sum of the conditions of the items, essentially "how many rods worth of fuel do you have left" (400% meaning there's 4 full rods). I would also argue that making it output the average of the conditions instead of the sum would make the name of the connection very misleading.

While I can see how this complicates calculating the fuel efficiency of the reactor, to me that seems like a somewhat rare use case - the most common use for the output is simply to measure how much fuel (in total) you have left, which would become more difficult if the output sent out the average - 50% could mean there's just one half-used rod, or it could mean there's four of them.

And even if it wasn't a bug, we could not change it just like this with no backwards compatibility support without breaking a ton of custom submarines that rely on the old behavior.

PR #12298, opened 2023-07-06
Orca 1 Refresh

WJohn:
The Orca 1 is out of date in many ways, yet most players are pushed towards using it as it's one of the first available early-game Submarines with a Fabricator, but is mired by many longstanding annoyances. This PR aims to address that, while leaving its quirks and general gameplay intact.

**Here are the changes made:**
- Swapped legacy walls for current walls. Hull shape and sizes had to be refitted, but layout and appearance look the same.
- Emergency Sirens are now used for the Reactor Meltdown alarm, instead of whenever water is on the ship.
- Alarm Buzzer by the Navigation Terminal is now used for flood alert.
- Light has been added in the middle of the Gunnery, it had a very large dark spot.
- Removed redundant Supplies Cabinet on the left of the Navigation Terminal, there is already one on the right.
- Single Diving Locker has been added by the lower Docking Port, as only 3 Diving Lockers was a bit sparse, and it took a very long time to get suited up. Could be increased to two, if needed.
- Cleaner wiring across the ship, more and better use of cable holders.
- Linked Fabricator to the Cabinet.
- Removed obsolete Railgun Shell Rack, replaced with Weapon Holder.
- Added Light Component by the Toggle Batteries button, as there was no clear way to tell when backup power was on.
- Lightly tweaked aesthetics such as reduced wiring decor (it was everywhere), different backgrounds for Medbay and Armory, added labels, changed Engineering/Reactor/Engine Room lighting to be more orange.

Horizontal speed is still 24km/h, and descent speed is still 15km/h. Ballast hull sizes and engine values have been lightly tweaked to have very similar numbers, (Neutral Ballast Level 0.4150 -> 0.4167)

**Testing done:**
- Bots appear able to navigate without issue. They were able to repair the entire submarine when completely destroyed, and navigate back to the sub when brought all the way around outside.
- No changes made to docking ports on either end, but were tested to still work normally.
- No changes made to gun placement or power values.

image
Orca_Refresh_2


Regalis11:
Thank you for your contribution! Merged to our private repo.

PR #12321, opened 2023-07-10
Component FalseOutput defaults to 0

mygamingaccount:
Reasons for not merging: makes newcomers learn the hard way. This must be the most often encountered issue with the wiring component system.
I propose this change in light of the recent QoL fixes.


Regalis11:
Thank you for the contribution! Merged to our private repo.

PR #12343, opened 2023-07-12
1-Handed Riot shield + visual fixes

ubertpendragon:
The riot shield isn't very useful, most use it for trolling. It also has some visual/audio issues.

Changes:

-Made 1-handed (left hand only). 1-h guns can shoot in front of the shield, 1-h melee can swing as long as the shield is aimed downwards
-Reduced health by 50% to compensate for the shield getting much better
-Changed collider size to match the sprite more closely
-Fixes to hit sound not playing each hit
-Changed break sound
-Fixes to hit particles, more visible and directional 
-Changed break particle, more visible
-Fixes to handle position

image
image
image


ubertpendragon:
I suck at PRs

PR #12312, opened 2023-07-08
Berilia Refresh

WJohn:
The Berilia is a rarely used Transport Submarine that is utterly overshadowed by all the other Tier 3 submarines, and even the much lower Tier 1 Camel in most ways. Additionally, it has a number of technical issues that make it difficult to use even for an experienced crew. This PR aims to improve quality of life across the submarine and usability, but leaves its existing mechanics intact.

<details><summary><b>Changelist and reasoning:</b></summary>
<p>
<li><b>Lighting has been dramatically reworked, and generally softened.</b> Both Junction Box rooms now have light in them where previously they were completely dark.</li>
<li><b>Medical Fabricator has been added to the Medbay, replacing the Hospital Bed.</b> As a Tier 3 submarine of this size, it really should have one when even the Tier 1 Camel does.</li>
<li><b>Gunnery's Periscope/Loaders have been reorganized and labelled to be much more intuitive to use.</b> They previously confused players constantly, as even the loaders did not correspond to what gun they were linked to.</li>
<li><b>Armory has been added to the right of the Gunnery,</b> as one single Secure Cabinet is insufficient for the crew sizes this Submarine demands.</li>
<li><b>Lower Bilge on the right of the Gamma Deck has been removed and replaced with just a Duct Block.</b> Players constantly mistook it for the lower Airlock Access ladder.</li>
<li><b>Redundant Diving Suits above the Lower Airlock have been moved further to the left above the Ballast,</b> as Engineers and Mechanics have to dive in those areas constantly. This helps bots respond in a more reasonable amount of time, too.</li>
<li><b>Linked Engineering's Status and Sonar Monitor, as well as the Medical Fab and its Cabinet.</b> Basic QoL usability.</li>
<li><b>Flow Rate of the Ballasts has been increased from 200 -> 300, and evened out the Ballast sizes.</b> Collectively, this reduces time to completely fill or empty the Ballasts from 28 seconds -> 18. Due to the separator walls on each tank and there only being one Hatch into the Ballasts, the actual drainage speed in case of a major flood is unchanged.</li>
<li><b>Flow Rate of the Lower Engine's Small Pump increased from 100 -> 200.</b> It took upwards of 70 seconds to fully drain that room alone. Since about a third of the ship routes its water towards the lower engine instead of the Ballasts, this often took significantly longer and mostly involved waiting around for 5+ minutes.</li>
[...]

PR #12356, opened 2023-07-14
(another) Berilia rework

ubertpendragon:
Koku and I were working on the Berilia at the same time when the other PR got posted so might as well upload this version too. 
This covers most of the issues brought up over at #12312.

There are some layout changes, the goal it to put the large empty space to use, improve navigation and give each room a purpose or a point of interest.

**Major layout changes**
- Cargo split into two: cargo and living area (transport goods and personel)
- 2 new docking ports for shuttles/docking, 1 above cargo and 1 near the bottom engine
- Ballast area is accessible from 3 different entry points, resized ballast catwalk
- 1 more ballast on the right side to compensate
- Clown room moved near ballast
- Larger reactor + reactor room
- Proper medbay and Armory
- Janitor closet on top right (flavor space for hiding bodies)
- Bottom right airlock merged in a single room
- Junction boxes more spread around the submarine
- Moved diving suit lockers around

Other notable changes:
- Bottom left Flak moved to top right, Hardpoint moved to bottom right
- Added one ventral coilgun to compensate
- Uses way less lights and shadow casting lights by repositioning them 
- NBL changed to 0.4883


ubertpendragon:
wikiimage

mygamingaccount:
Don't use custom doors, they no longer have a place on subs unless some very weird design requires it. Use normal doors, they now have an activate output.

Koku0:
> Don't use custom doors, they no longer have a place on subs unless some very weird design requires it (for example when the trigger is on the opposite end of the submarine). Use normal doors, they now have an activate output.

The custom door has a use; it's meant to be for the clown/traitor door. You find the button, activate it, and run into the secret room.

PR #12358, opened 2023-07-15
Simple no-rubberband airlock using the new Activate Out signal

mygamingaccount:
Airlock to replace the glitchy interlocking circuit which was notorious for rubberbanding in multiplayer games Often the player was halfway inside when the door got shut, and the player teleported to the other side of the door.

This simple circuit always opens the door being operated, and always closes the opposite one. Component count was kept by simplifying the pump logic.
The assembly can be copied without ordering issues.

image


Regalis11:
Merged to our private repo

PR #12383, opened 2023-07-20
Spineling genes rework

ubertpendragon:
# Problems with spineling genes:

- **Impossible to aim the spikes**, and impractical. You have to point your back at the enemy's direction and hope for the best
- **No gene quality scaling**. A 10% spineling gene has the same effect as a refined, 100% spineling gene
- **It's weak**. Damage is negligible as a ranged weapon, and the melee spike is weaker than a basic knife

# Changes

The big change is that spineling genes can now be used as an emergency melee/harpoon generator, at the cost of losing blood. This technically gives you infinite weapons, but it has downsides

- **Spikes drop as an item, rather than being shot out of your back.**
- **Ejecting a spike will cause bleed/bloodloss, based on the quality of your gene**. Qualities over 33, 66 and 100 will cause less self-damage. 
- **Spikes (including looted ones) can be shot out of harpoons.** Simply load the spike into a harpoon and shoot.
- **Spikes can still be used as a melee weapon**. Melee damage increased, BUT the spikes break after 7 melee hits.
- **Spike stack size reduced to 1.** Necessary to prevent combining shenanigans, and from a balance perspective to avoid spike stacking as a replacement for harpoons. 
- **Spikes no longer deconstruct into resources and cannot be sold.** Since you can generate infinite spikes now. 

https://github.com/Regalis11/Barotrauma/assets/73229309/c13dc0fe-a5af-405b-b1b2-4f8b8c2c2a9b

# Issues:

- Security's **Slayer** affects the melee damage (spikes are tagged as harpoons). Not sure what to think about it

### DIFF
--- a/Spinelingspine.xml
+++ b/Spinelingspine.xml
@@ -1,0 +1,49 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<huskappendage>
+  <limb id="13" width="10" height="60" name="Spike1" type="None" notes="" scale="0.7" flip="True" mirrorvertically="False" mirrorhorizontally="False" hide="False" spriteorientation="0" steerforce="0" radius="0" density="1" ignorecollisions="False" angulardamping="7" attackpriority="1" pullpos="0,0" stepoffset="0,0" refjoint="-1" mouthpos="0,0" constanttorque="600" constantangle="0" attackforcemultiplier="1" minseverancedamage="1" severedfadeouttime="10" applytailangle="False" sinefrequencymultiplier="1" sineamplitudemultiplier="1" blinkfrequency="0" blinkdurationin="0.2" blinkdurationout="0.5" blinkholdtime="0" blinkrotationin="0" blinkrotationout="45" blinkforce="50" blinktransitionin="Linear" blinktransitionout="Linear" healthindex="0" friction="0.3" restitution="0.05">
+    <sprite texture="Content/Characters/Spineling/Spineling.png" sourcerect="229,0,26,183" origin="0.5385919,0.47905606" depth="0.61" color="255,255,255,255" deadcolor="255,255,255,255" deadcolortime="0" />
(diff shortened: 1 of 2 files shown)


Abkim27:
Please, try to make them spawn inside inventory

G-M-twostay:
or just make it so that they don't shoot to random directions, so if a player is skilled enough they can aim the spikes, tho without corsair like guns.

Regalis11 (changes requested):
This does not seem ready for a review (should probably be a draft?).

There are several issues remaining here, as you already commented. I also agree with G-M-twostay: I think being able to aim the spikes would be a better way to make the spikes more useful.

ubertpendragon:
> This does not seem ready for a review (should probably be a draft?).
> 
> There are several issues remaining here, as you already commented. I also agree with G-M-twostay: I think being able to aim the spikes would be a better way to make the spikes more useful.

I'll clarify the "problems". Some are details some are more annoying:

# Minor problems

- **Balance**: it's in the eye of the beholder, I'm still fine tuning it, whether it is the damage of the spike itself, or the self-damage. From testing you cannot mitigate the self-damage which is good. 

- **Ejected Spikes**: The downside is hitting your allies when ejecting a spike. The upside is that ejecting the spike itself can hit enemies swimming towards your back. Which is fun.

- **Two spikes**: it might be confusing to have two types of spikes, one that yields resources and one that doesnt. But it's to prevent spawning resources out of the ether. Could simply remove spineling spike deconstruct yields since it's now a much better weapon/ammo. 

# Annoying problems:

- **Bleeding is located on the head** instead of Torso. I've tried but can't figure out how to specifically target the torso. 

- **Slayer** boosting melee damage: There is no way to make slayer NOT apply the bonus to melee, since both melee and projectile are in the same item.
The workaround is to add a `harpoonweapon` tag to harpoons, change Slayer to detect if a harpoon gun is equipped and then use `RangedAttackMultiplier`... which is not supported any more! 

image

The xml would look like:

```xml
    <AbilityGroupInterval interval="0.2" >
      <Conditions>
        <AbilityConditionHasItem tags="harpoonweapon"/>
      </Conditions>
      <Abilities>
        <CharacterAbilityModifyStat stattype="RangedAttackMultiplier" value="0.25"/>
      </Abilities>
    </AbilityGroupInterval>
```

It's up to your appreciation whether to keep the slayer bonus or suffer the process of fixing that.

ubertpendragon:
Updated. This should solve all of the problems

The potentially contentious change is that spineling spikes can no longer sold or deconstructed, since both spikes were merged into one; can split them again if need be. 

Still needs #12548 to be fixed

ubertpendragon:
Updated main post to reflect the changes. 

The only remaining issue is Slayer, but I gave a potential solution if the extra damage is a problem

Regalis11 (changes requested):
As mentioned in the previous review, I think being able to aim the spikes would be a better way to make the spikes more useful.

ubertpendragon:
> As mentioned in the previous review, I think being able to aim the spikes would be a better way to make the spikes more useful.

You aim the spikes with a harpoon :) 

The other option is turning the limb into a turret and have it aim but honestly shooting from the back simply isnt practical and you lose the spineling sword function

ubertpendragon:
I've done some more testing to try and ease the aiming with the spikes, but they just keep flopping around every time you move. 

This feels like the best compromise, and it also fixes an issue with the vanilla spikes which is firing them at point blank for massive damage (the only way to not miss), which feels counterintuitive as it turns into a melee weapon

Regalis11 (changes requested):
As I said in the previous review, I think it would be better if you could aim the spikes (and by this I mean actually being able to aim them with the mouse, like you can with ranged weapons).

PR #12421, opened 2023-07-26
Implement blinking for decorative sprites

Jade-Harleyy:
Does what the title says.
Shouldn't be any noticible performance impact.


Regalis11:
Finally got around to giving it a test. Couldn't spot any problems so I merged it to our private repo. Thank you for the contribution!

PR #12483, opened 2023-08-09
Mudraptor genes rework

ubertpendragon:
# Problems with Mudraptor genes:

- **No gene quality scaling**. A 10% mudraptor gene has the same effect as 100%
- **It's weak**. Damage is negligible, and compared to a crowbar/axe it is way too slow at breaking down doors

# Changes

- **4 beak sizes and colours at 0-33-66-100 thresholds**. Limbs use conditionals to appear/disappear dynamically. 
- **Damage scaling**. Both bitewounds and item damage scaled up with quality. Breaks down doors in 2 hits at 100.
- **Armor scaling.** The beak protects the head and cannot bleed. Damage resistance increases with quality
- **Added sound for feedback**.

image

```

# Issues

* Method to hide limbs to figure out. See comments below

### DIFF
--- a/Mudraptorbeak.xml
+++ b/Mudraptorbeak.xml
@@ -7,9 +7,10 @@
       <HideLimb duration="100000" />
       <conditional naturalmeleeweapon="gt 33" />
     </StatusEffect>
-    <damagemodifier armorsector="0,360" afflictiontypes="damage" damagemultiplier="0.4" damagesound="LimbArmor" damageparticle="brownchunks" deflectprojectiles="True" afflictionidentifiers="" probabilitymultiplier="1" />
-    <damagemodifier armorsector="0,360" afflictiontypes="bleeding" damagemultiplier="0.3" damagesound="LimbArmor" deflectprojectiles="True" afflictionidentifiers="" probabilitymultiplier="1" damageparticle="" />
-    <damagemodifier armorsector="0,360" afflictiontypes="burn" damagemultiplier="0.25" damagesound="" deflectprojectiles="True" afflictionidentifiers="" probabilitymultiplier="1" damageparticle="" />
+    <!-- Takes melee/projectile damage instead of the head because the limb has a larger collider (50 vs 45) (unless projectile pierces). Stronger with quality. Takes no damage from bleeding/burns/explosion because those hit in an AOE and will hit the limb AND the head -->
+    <damagemodifier armorsector="0,360" afflictiontypes="damage" damagemultiplier="0.4" damagesound="LimbArmor" deflectprojectiles="True" damageparticle="brownchunks" />
+    <damagemodifier armorsector="0,360" afflictiontypes="bleeding" damagemultiplier="0" deflectprojectiles="True"  />
+    <damagemodifier armorsector="0,360" afflictiontypes="burn" damagemultiplier="0" deflectprojectiles="True"  />
     <attack context="Any" cooldown="2" afterattack="Pursue" range="50" damagerange="50" duration="0.5" structuredamage="0" itemdamage="25" structuresoundtype="StructureSlash" targetforce="0" severlimbsprobability="0.3" torque="100" force="30" applyforceonlimbs="0,1" targettype="Any" hitdetectiontype="Distance" stun="0" onlyhumans="False" applyforcesonlyonce="True" targetimpulse="5" targetimpulseworld="0,0" targetforceworld="0,-5" stickchance="0" priority="1" secondarycooldown="2" cooldownrandomfactor="0.5" reverse="False" targetlimbtype="None" afterattackdelay="0" retreat="False" rootforceworldstart="0,0" rootforceworldmiddle="0,0" rootforceworldend="0,0" roottransitioneasing="Linear">
       <conditional targetself="true" naturalmeleeweapon="lt 33" />
       <Affliction identifier="bitewounds" strength="10" probability="1" />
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by Regalis11 — CHANGES_REQUESTED** [review]
Not ready for a review, there are still some issues remaining here as you commented.

**Comment by ubertpendragon** [comment]
I've updated to fix the issues. 

It also turns out that the armor aspect of the genes never worked because the beak collider was smaller than the head collider, which I've fixed too. 

Using weldjoint required to rotate the beak sprite. Here it is:

Genetic

**Comment by ubertpendragon** [comment]
One hard-coded problem I'm running into is the inability to properly activate limbs based on the level of the affliction, or the ability to "truly" disable a limb. 

Currently I'm switching between sprites using `Conditionalsprite`

The condition checks for the level of the afflcition and makes the proper sprite visible.

```xml
    <sprite sourcerect="0,0,1,1" origin="0.36,0.64" depth="0.049" color="230,200,150,255" deadcolor="255,255,255,255" deadcolortime="0" texture="Content/Items/Genetic/Genetic.png" ignoretint="true" />
    <Conditionalsprite exclusive="True">
      <Conditional naturalmeleeweapon="gte 66" />
      <Conditional naturalmeleeweapon="lte 99" />      
      <sprite sourcerect="410,18,97,67" origin="0.36,0.5" depth="0.049" texture="Content/Items/Genetic/Genetic.png" />
    </Conditionalsprite>
```

Unfortunately the sprite collider itself doesnt disappear. 
This can be seen on Hammerhead Matriarch:

image

The head has "exploded", but since this only changes sprites the collider is still the same.

ubertpendragon:
Just changed the logic to make scaleable armor values to work.
Before all limb colliders were active instead of only 1

This uses a statuseffect to hide limbs based on condition:

```xml
    <!-- hides the limb using conditions -->
    <StatusEffect type="Always" target="This" interval="1" >
      <HideLimb duration="100000" />
      <conditional naturalmeleeweapon="lte 66" />
      <conditional naturalmeleeweapon="gt 99" />
    </StatusEffect>
```

Need to validate performance

Mudraptorbeak.xml
```diff
@@ -0,0 +1,96 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<huskappendage>
+  <limb id="13" name="Beak0" type="None" inheritlimbdepth="Head" width="0" height="0" notes="" flip="True" healthindex="0" attackpriority="0" steerforce="0" stepoffset="0,0" radius="50" mass="0" density="0" pullpos="0,0" refjoint="-1" ignorecollisions="False" friction="0.3" restitution="0.05" spriteorientation="90" mirrorvertically="False" mirrorhorizontally="False" hide="False" angulardamping="100" mouthpos="0,0" constanttorque="0" constantangle="0" scale="0.5" attackforcemultiplier="1" minseverancedamage="0.8" severedfadeouttime="0" applytailangle="False" applysinemovement="True">
+    <sprite sourcerect="410,18,97,67" origin="0.29,0.6" depth="0.049" color="180,115,115,255" deadcolor="255,255,255,255" deadcolortime="0" texture="Content/Items/Genetic/Genetic.png" ignoretint="true" />
+    <!-- hides the limb using conditions -->
+    <StatusEffect type="Always" target="This" interval="1" >
```
Regalis11, line 6: This feels like a somewhat hacky way to go about this. Why does the effect need to run every second? Why does it hide the limb for 100000 seconds instead of indefinitely? I think it would be preferable to only run this once (I'm not sure if there's existing support for that, but could be done e.g. by making OnSpawn effects run on the limbs when the limb spawns) and hide the limb forever (can be done just by omitting the duration).

Regalis11 (changes requested):
To me it feels like the approach for disabling the inactive limbs could be implemented in a cleaner way, see comments

Mudraptorbeak.xml
ubertpendragon: If I recall, it's to refresh the beak when you equip or unequip genes. 

You can only check the genetic affliction to determine whether to hide the limb or not, so the statuseffect does it regularly. I can't think of another way.

As you said the duration of 100000 is basically forever, couldnt think of  another way to do it

Will run some tests

ubertpendragon: I have tried both:

`    <StatusEffect type="OnSpawn" target="This,Limb" >`
which doesn't do anything

`    <StatusEffect type="OnActive" target="Limb" >`
which does work but probably runs every second, and also causes issues with advanced gene splicer if you add two mudraptor genes that take you over the threshold

not specifying duration sends an error message

Regalis11: Yeah, OnSpawn effects don't work on limbs atm, I think that should be implemented though.

> not specifying duration sends an error message

That is also something that should be fixed. I'm not seeing any errors when removing the duration though, what kind of an error are you getting?

Regalis11: > I still think the advanced gene splicer issue would remain with OnSpawn, if you equip two partial mudraptor genes and the affliction goes over 33/66/100, the limbs become invisible

I don't quite understand why that's an issue with OnSpawn, but not when the effects run constantly. What unhides the "higher-tier" limb when you use Always?

ubertpendragon: I assume the limbs are all spawned at once when you get the mudraptor genes affliction (any value)
It then checks the affliction every second and hides 3 out of 4 limbs based on affliction value

OnSpawn will only check the initial value of the affliction when the gene is equipped.
If the affliction value is changed by adding a 2nd mudraptor gene in Advanced Splicer slot 2, the value changes, but since the limb is hidden forever "OnSpawn" (with no duration) it cannot appear again. 

OnSpawn only works if you completely remove the gene, which is not a problem with normal splicer, but an issue with advanced splicer

Regalis11: > If the affliction value is changed by adding a 2nd mudraptor gene in Advanced Splicer slot 2, the value changes, but since the limb is hidden forever "OnSpawn" (with no duration) it cannot appear again.

I understand that, but what unhides the limb when you use Always?

ubertpendragon: The fact that it runs on an interval so it checks the conditionals I assume? 

I honestly don't know but it works that way 😔

Regalis11: It hides the limb every x seconds, but it doesn't seem like the limb is ever unhidden.

Regalis11: I gave the latest version a test now, and it seems the limbs are never unhidden. If you put a let's say 50% quality mudraptor gene in an advanced gene splicer, and then add another one, no limbs are visible (because the highest-tier limb is hidden and there's nothing that'd unhide it).

I feel like getting this to work robustly might require some code changes. Another option could be to only hide the limbs for some duration (one second for example), and re-check the conditions periodically (with a slightly smaller interval than the limbs are hidden).

ubertpendragon: True, I think it's a bigger issue with advanced gene splicer which doesn't "update" when you insert new genes, I've tried several methods

Hiding limbs is possible, unhiding them isnt

ubertpendragon: > I feel like getting this to work robustly might require some code changes. Another option could be to only hide the limbs for some duration (one second for example), and re-check the conditions periodically (with a slightly smaller interval than the limbs are hidden).

Tried 

    <StatusEffect type="Always" target="Limb" interval="0.5">
      <HideLimb duration="1" />
      <conditional naturalmeleeweapon="gt 33" />
    </StatusEffect>
    
    
    No luck

ubertpendragon:
Pushed a new version that is "functional".

Only the damage scales now

Issue #12595, opened 2023-08-31
Wrecked diving suit lockers can't contain all suits, look weird


ubertpendragon:
### Disclaimers

- [X] I have searched the issue tracker to check if the issue has already been reported.
- [ ] My issue happened while using mods.

### What happened?

Wrecked diving suit lockers are using item identifiers for containables instead of broader item tags

`      <Containable items="divingsuit,abyssdivingsuit,combatdivingsuit" />`

This means they cannot contain modded suits or even Disposable diving suits
They should be using the same line as regular lockers:

`      <Containable items="deepdiving" excludeditems="deepdivinglarge" />`

Also when containing a suit, the lockers look completely different, the suit's contained sprite overriding the dirty window:

![image](https://github.com/Regalis11/Barotrauma/assets/73229309/6d1eff91-e29a-46eb-a570-e200fd5a7fbc)

This can be solved by
1. Using the regular diving locker sprite as a base Sprite,
2. Change the Wrecked Locker sprite's opacity to 70% / 180 alpha (in the spritesheet)
3. Add the wrecked locker sprite as a Decorativesprite with less depth

_Left to Right: Empty locker, Filled locker, Vanilla wrecked empty locker_
![image](https://github.com/Regalis11/Barotrauma/assets/73229309/7100eb3d-2677-4e16-92c9-341bf5430197)

```xml
  <Item name="" identifier="divingsuitlockerwrecked" nameidentifier="divingsuitlocker" tags="divingsuitcontainerwindow,wreckdivingsuitcontainer" pickdistance="50" scale="0.5" category="Wrecked">
    <!-- regular locker sprite -->
    <Sprite name="Diving Suit Locker Vertical" texture="Content/Items/Containers/containers.png" depth="0.84" sourcerect="714,1,99,256" origin="0.5,0.5" />
    <!-- 70% opacity wrecked locker sprite on top -->
    <DecorativeSprite texture="Content/Items/Shipwrecks/MiscWrecked.png" depth="0.83" sourcerect="1126,1710,99,254" origin="0.5,0.5" />    
    <ItemContainer hideitems="false" drawinventory="true" capacity="1" slotsperrow="1" itempos="0,0" iteminterval="0,0" canbeselected="true" msg="ItemMsgInteractSelect">
      <GuiFrame relativesize="0.2,0.25" anchor="Center" minsize="140,170" maxsize="280,280" style="ItemUI" />
[...]

PR #12657, opened 2023-09-11
Adds a "broken" diving suit for wrecks/corpses and visual adjustments

ubertpendragon:
Adds a "broken" diving suit from the unused broken diving suit sprite. 

-Uses the assests from Husked_DivingSuit.png to create a new wearable diving suit
The suit doesn't give any pressure resistance and will leak oxygen, can be deconstructed for lesser yields than a regular diving suit
image

-Diving suit added to corpses loot as well as wrecked diving lockers
Some corpses will wear the suits, and the suit will spawn in wrecked diving suit lockers

image

~~-Changed wrecked diving lockers sprite (see #12595)~~

---
# Sprite changes:

- ~~**MiscWrecked.png** adjusted the wrecked diving suit locker sprite to 75% opacity to be used as a decorativesprite~~

- **Headgears.png** color adjusted the broken helmet to match the suit tone and color. Also adjusts husks

- **Broken_Divingsuit.png** basically husked_divingsuit but with mild adjustments to shape and size to exactly match the regular diving suit model

- **Broken_DivingSuit_contained.png** assembled sprites for diving suit as an item; on the floor, contained, and inventory icon


Regalis11:
Thank you for the PR! I'm not entirely sure of the reasoning behind these changes - is it to reduce the value of loot from the wrecks? Just for flavor (makes more sense for the suits to be degraded in wrecks)?

> MiscWrecked.png adjusted the wrecked diving suit locker sprite to 75% opacity to be used as a decorativesprite

I don't see why the texture needs to be adjusted for this - imo a better way would be to just draw the sprite with 75% opacity on the locker sprite (which would also mean the look of the cabinet doesn't change, since the base sprite is still the same). Also, for some reason this PR over doubles the file size of the atlas. The issue [has also already been fixed in our private repo](https://github.com/Regalis11/Barotrauma/issues/12595#issuecomment-1713348777).

ubertpendragon:
> Thank you for the PR! I'm not entirely sure of the reasoning behind these changes - is it to reduce the value of loot from the wrecks? To add more loot? Just for flavor (makes more sense for the suits to be degraded in wrecks)?

Mostly flavor, the sprite was unused, and it was weird seeing corpses dead without even trying to equip suits, or suits in lockers that are completely pristine

Balance wise this does dilute the "Loot Pool" for lockers, but also adds diving suits on the corpses themselves, which balances out.

> I don't see why the texture needs to be adjusted for this - imo a better way would be to just draw the sprite with 75% opacity on the locker sprite

If I knew it was possible I would've done that 😅 

> Also, it's practically impossible to see what has been changed from this PR, since the default files and the changes to them have been added in the same commit.

My bad.

I changed 2 things:
-The locker visuals (which I will remove)
-Added broken suits to Corpses inventory


All I did with corpses was sprinkle this on some existing ones:

```xml
      <Item identifier="brokensuit" equip="true">
        <Item identifier="oxygentank"/>
      </Item>
```

Feel to add to existing corpses or duplicate corpses (naked/suit) however you like


I'll update the PR to remove the locker stuff

Regalis11:
Merged this to our private repo now (with some minor modifications), thank you for the contribution. :)

PR #12703, opened 2023-09-21
Add option in multiplayer lobby settings to disable the alive crew list

27alaing:
### What this PR does
Adds an option in the lobby settings under gameplay section to disable the crew list

### Why
Roleplay purposes, you may not want to have a list that tells you everyone in the game and if they're alive
traitor mode, disabling the crew list will add more mystery to the mode if someone dies

https://github.com/Regalis11/Barotrauma/assets/45323883/e6a8c290-fbc7-44c3-af94-1b18feef385f

image

The crew list is enabled by default

### DIFF
--- a/Barotrauma/BarotraumaClient/ClientSource/Screens/NetLobbyScreen.cs
+++ b/Barotrauma/BarotraumaClient/ClientSource/Screens/NetLobbyScreen.cs
@@ -2275,6 +2275,7 @@ namespace Barotrauma
             options.Add(new ContextMenuOption("Rank", isEnabled: canManagePermissions, options: rankOptions.ToArray()));
 
             Color clientColor = client.Character?.Info?.Job.Prefab.UIColor ?? Color.White;
+            if (GameMain.NetworkMember.ServerSettings.DisableCrewList == true) { clientColor = Color.White; }
 
             if (GameMain.Client.ConnectedClients.Contains(client))
             {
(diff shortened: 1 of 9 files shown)


Barotrauma/BarotraumaClient/ClientSource/Screens/NetLobbyScreen.cs
```diff
@@ -2275,6 +2275,7 @@ public static void CreateModerationContextMenu(Client client)
             options.Add(new ContextMenuOption("Rank", isEnabled: canManagePermissions, options: rankOptions.ToArray()));
 
             Color clientColor = client.Character?.Info?.Job.Prefab.UIColor ?? Color.White;
+            if (GameMain.NetworkMember.ServerSettings.DisableCrewList == true) { clientColor = Color.White; }
```
Regalis11, line 2278: A minor note, but there's no reason to check equality to true here, `if (GameMain.NetworkMember.ServerSettings.DisableCrewList)` is enough.

What this line here is doing is essentially doing the check twice, "if it is true that the value of this variable equals to true", when it could be "if the value of this variable is true".

Applies to most of the other checks for `DisableCrewList` in this PR.

Regalis11 (changes requested):
Thank you for your contribution! However, I don't think just hiding the crewlist is an adequate solution here, because it also prevents you from assigning orders to bots (or other crewmates) using the crew list, makes it impossible to reprioritize orders, prevents you from seeing who's talking in the voice chat, and (in the upcoming traitor update), from voting for the traitor.

27alaing:
Yeah, I thought about that .  another possible solution would be you need to report the body in which, when someone dies it doesn't update on the crew list until the body is reported (you would need to interact the corpse with use key)

Regalis11:
I think another option could be to just leave the character in the crewlist regardless if they're dead or not. But regardless, I don't think we can use the approach in this PR because of the issues mentioned above, so I'll close it. Thank you for the PR and ideas regardless!

PR #12794, opened 2023-10-12
Barsuk visual tweaks and bug fixes

NotWendy:
General gameplay changes:

- Minor hull box adjustments and waypoint changes. Did a lot of testing with bots and they seem to like it.
- Made bots capable of using the airlock by linking buttons to doors. Previously the bots would get confused by the wiring and wouldn't know how to open or close the outer door.
- Added an assistant spawnpoint with an appropriate ID in the cargo room.
- Fixed the decorative shuttle shell piece placement around turrets causing rooms to flood when the shell is broken.
  For example, you could break the shell around the top turret without damaging the main hull pieces, and the cargo room would flood.
- Removed the shuttle shell piece above the laser. Replaced it with a decorative fin piece.
  It was barely visible and only served to make you question what exactly you're welding, when you were repairing the front of the submarine.
- Barsuk's batteries now start out fully charged, with their charge rate set to 20%.
  Reinforcing the idea of Barsuk being an easy to understand starter sub, this quality of life change makes the use of batteries slightly more intuitive.

Wiring changes:
- Made the powerful relays that control Barsuk's batteries non-detachable.
  Their higher than normal power limit (5000 vs 1000) meant you could start on a Barsuk, detach the relays, then use said relays to make other submarines more powerful.
- The state of the primary battery relay is now inverted via a NOT component, instead of the Signal Check component. This is a more intuitive setup, which matters for a highly customizable starter submarine that Barsuk is supposed to be.
- Added a lamp in the oxygen compartment. It is powered by the bottom junction box in the gunnery. The oxygen generator itself no longer serves as a source of light.
  Thanks to the oxygen generator now having an animation when it's active, having a light component associated with it isn't necessary and a separate lamp looks more natural.

Lighting changes:
- Reduced the range and slightly increased the intensity of emergency lights on cabinets. The purpose of those lights is to guide players in the dark when the submarine is out of power.
- Made the buttons in the airlock, outside of the submarine, and next to the depth charge loader non-detachable and placed non-interactable light components behind them, emitting a faint light in a range that matches the button's size. 
  Those three buttons are important, but hard to see in the dark. Highlighing them with non-interactable light components makes them easier to find and press.
- Made the emergency light in the bottom front room brighter and made it cast shadows, to prevent it from shining through the command room's floor.

Other visual changes:
- Fixed countless sprite depth conflicts across the entire submarine.
  Decorative boxes rendering behind junction boxes in the docking port room, reactor overlapping the platform it's on, metal strips around the docking hatch sometimes being hidden, shuttle shell pieces around guns sometimes rendering on top of the main hull, all fin parts having the same sprite depth, ladders having the same sprite depth as hatches, and many more.
- Reduced the amount of functional wires being used for decorative purposes. (Primarily in the reactor room.) Also smoothed out the curves on remaining wires by moving the nodes around.
  The jagged shapes you get when trying to decorate with real wires don't look right. Instead, the messy wiring is now implied with decorative wires, some of which show real wires underneath.
- Gave Barsuk a rudder and increased the sprite depth of the BARSUK label, so that the label's letters can't appear above wires, if those happened to be placed there mid-mission. Made other exterior decorations a bit more visually appealing too. Also, more rust.
- Added a platform on top of the docking hatch. Its bright texture makes the hatch more noticeable from the outside.
- Moved some items like vents and small supply cabinets around to align them with wall textures better.
- Moved the navigation terminal higher to save the captain from having to lean down all the time.
- Recolored some items (primarily vents) to make them stand out less.
- Moved lots of boxes around. They look better this way, I think.

Images:
BarsukRemake
BarsukRemakeLighting


Zalaxx:
Really like this small refresh! (based on image comparison with the current version and the list of changes)
It makes it a bit more "coherent" and easier to understand (clutter/decoration is better integrated with the submarine)
I particuliarly like the wirings mixing function and decoration, very creative and looks really good!
I hope it will be merged and make it for the next release :)

NotWendy:
Thank you, @Zalaxx.
Also, @JeroonBosch, I just pushed one more update for this, a few small visual tweaks I wanted to do for a while. Mostly lighting and sprite depth changes to make it so less wires are covered up by boxes. That's something that wasn't so good about the last version.

NotWendy:
Another update. Cleaned up the waypoints around ladders, did a lot more testing with bots. Added a spawnpoint for the assistant with an appropriate ID in the cargo room. Probably won't ever make a difference, but it's good to have for consistency in design.
Important note: Bots have no idea how to use the airlock without the player's help. This is the case in the original Barsuk too. I tried fixing this problem by marking the outer airlock door as a blue waypoint (indicating that it leads outside). This works sometimes, but did not help in this case. Might not be fixable in the editor if we want to stick with buttons that aren't integrated with the door itself.

NotWendy:
Rewired the airlock so that bots can use it without any problems. I'm so happy about this. They were confused before because the buttons were only connected to the inner door, and the outer door's state changed depending on whether the inner door is closed or open.

It still works the same way in my version, I just send redundant signals to the outer door, so the bots recognize that the button controls it too.
The outer button sends a signal to the inner door and to the outer door. The signal to the outer door gets overwritten by the Not component immediately, so it's completely irrelevant on its own. But the bots like it.
The central button works the same way. And the innermost button is only wired to the inner door.
image

NotWendy:
The airlock is bot-friendly now. I linked the buttons to doors, so the bots don't get confused by the wiring.

mygamingaccount:
What about the weird and misleading airlock text display? "Wait for the airlock to drain before exiting"?? It has alternating hatches, there's no way to wait for the airlock to drain. Either a different circuit is needed, or the text display has to go.

The change gets an automatic :-1: for having non-interactable, non-detachable parts. These buttons might need to be removed in order to fix the airlock circuit during the campaign, for example due to the previously mentioned issue.

NotWendy:
> What about the weird and misleading airlock text display? "Wait for the airlock to drain before exiting"?? It has alternating hatches, there's no way to wait for the airlock to drain. Either a different circuit is needed, or the text display has to go.

Damn, no fun allowed. Barsuk's shitty construction is the point. The airlock text is a joke from the devs. You're always free to rewire the buttons. Anyways, submarine PRs don't really get merged, no downvotes are needed to make sure this work goes to waste.

NotWendy:
Closing this PR to redo it later.

PR #12831, opened 2023-10-17
GoTo mission type enhancements

dakkhuza:
Currently the GoTo mission type completes at the end of the round if either:
- The loaded level type is an outpost
- The sub is at the end exit of the level

There is no way to affect if the mission should count as completed through XML as it will always complete regardless of the state of the mission.

This is especially strange as if you look at the [client code](https://github.com/Regalis11/Barotrauma/blob/ca2c30edca3e0edf6ad69cf35ea62024793c4730/Barotrauma/BarotraumaClient/ClientSource/Events/Missions/GoToMission.cs#L5) that handles displaying if the mission is finished or not, it will only display as completed if the mission state is >= the MaxProgressState (which is defined in XML). This leads to an unintuitive situation where the mission will complete at the end of the round even though the MaxProgressState was not met. I believe in vanilla this only affects the clown haven mission.

This PR adds two new properties to the GoTo mission:  `maxprogressdeterminescompleted` (bool) * `failstate`(int)

If `maxprogressdeterminescompleted` is true (default false) it alters the `DetermineCompleted` logic of the GoTo mission to require the mission State to be >= to the MaxProgressState for the mission to count as completed at the end of the round (reflecting the logic for displaying as completed for the client).

 `failstate` (default -1) determines the state the GoTo mission must be in to display as failed for the client.

This shouldn't affect any existing vanilla or modded missions as the change to completion logic is locked behind setting `maxprogressdeterminescompleted` to true.

### DIFF
--- a/Barotrauma/BarotraumaShared/SharedSource/Events/Missions/GoToMission.cs
+++ b/Barotrauma/BarotraumaShared/SharedSource/Events/Missions/GoToMission.cs
@@ -4,9 +4,11 @@ namespace Barotrauma
 {
     partial class GoToMission : Mission
     {
+        private readonly bool maxProgressStateDeterminsCompleted;
         public GoToMission(MissionPrefab prefab, Location[] locations, Submarine sub)
             : base(prefab, locations, sub)
         {
+            maxProgressStateDeterminsCompleted = prefab.ConfigElement.GetAttributeBool("maxprogressdeterminescompleted", false);
         }
 
         protected override void UpdateMissionSpecific(float deltaTime)
@@ -19,6 +21,11 @@ namespace Barotrauma
 
         protected override bool DetermineCompleted()
         {
+            if (maxProgressStateDeterminsCompleted)
+            {
+                return Prefab.MaxProgressState == State;
+            }
+
             if (Level.Loaded?.Type == LevelData.LevelType.Outpost)
             {
                 return true;
--- a/Barotrauma/BarotraumaShared/SharedSource/Events/Missions/GoToMission.cs
+++ b/Barotrauma/BarotraumaShared/SharedSource/Events/Missions/GoToMission.cs
@@ -5,10 +5,12 @@ namespace Barotrauma
     partial class GoToMission : Mission
     {
         private readonly bool maxProgressStateDeterminsCompleted;
+        private readonly int failState;
         public GoToMission(MissionPrefab prefab, Location[] locations, Submarine sub)
             : base(prefab, locations, sub)
         {
             maxProgressStateDeterminsCompleted = prefab.ConfigElement.GetAttributeBool("maxprogressdeterminescompleted", false);
+            failState = prefab.ConfigElement.GetAttributeInt("failstate", -1);
         }
 
         protected override void UpdateMissionSpecific(float deltaTime)
--- a/Barotrauma/BarotraumaClient/ClientSource/Events/Missions/GoToMission.cs
+++ b/Barotrauma/BarotraumaClient/ClientSource/Events/Missions/GoToMission.cs
@@ -3,6 +3,6 @@
     partial class GoToMission : Mission
     {
         public override bool DisplayAsCompleted => State >= Prefab.MaxProgressState;
-        public override bool DisplayAsFailed => false;
+        public override bool DisplayAsFailed => State == failState;
     }
 }


Regalis11 (changes requested):
Thank you for you contribution! This seems like a very useful addition, although I would consider implementing it in another way (see comments). I also think the "failstate" field doesn't work as intended now.

Barotrauma/BarotraumaShared/SharedSource/Events/Missions/GoToMission.cs
```diff
@@ -4,9 +4,13 @@ namespace Barotrauma
 {
     partial class GoToMission : Mission
     {
+        private readonly bool maxProgressStateDeterminsCompleted;
+        private readonly int failState;
         public GoToMission(MissionPrefab prefab, Location[] locations, Submarine sub)
             : base(prefab, locations, sub)
         {
+            maxProgressStateDeterminsCompleted = prefab.ConfigElement.GetAttributeBool("maxprogressdeterminescompleted", false);
```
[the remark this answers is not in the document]
dakkhuza: Originally the intention was to match the functionality of the client side display (were meeting MaxProgressState displays the mission as completed) and to change as little as possible. However it seems that a more intuitive functionality is desired. I'll work on a more complete functionality.

Barotrauma/BarotraumaShared/SharedSource/Events/Missions/GoToMission.cs
```diff
@@ -4,9 +4,13 @@ namespace Barotrauma
...
+            failState = prefab.ConfigElement.GetAttributeInt("failstate", -1);
```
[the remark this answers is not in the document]
dakkhuza: The mission implicitly fails if it's in the fail state as it's not in the success state at the end of the round, unless the state is changed afterwards in which case it will no longer be in the fail state and will thus not display as failed. Unless both the success state and fail state are the same, which imo would be a fault of the mission creator but I could put in a warning for it just to catch mistakes.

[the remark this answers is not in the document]
dakkhuza: In the current impl presenting by this pull request you could have a fail state that's higher than the MaxProgressState which would lead to weird situations where the mission completes even though it's displayed as failed, however the upcoming changes will fix that.

Barotrauma/BarotraumaShared/SharedSource/Events/Missions/GoToMission.cs
```diff
@@ -4,9 +4,13 @@ namespace Barotrauma
 {
     partial class GoToMission : Mission
     {
+        private readonly bool maxProgressStateDeterminsCompleted;
```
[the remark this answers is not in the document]
dakkhuza: Resolved as this property is no longer in the proposal, it was a bit of a mouthful anyways as noted above.

dakkhuza:
Updated the PR to address requested changes, functionality should now be more clear and easy to understand based on attribute names alone.

- `maxprogressdeterminescompleted` is replaced with `statecontrolscompletion`
- Added new property `successstate`
- When `statecontrolscompletion` is set to true, the mission will only display as completed and complete at the end of the round when `State == successState`
- Renamed `failState` -> `displayAsFailedState` to make it's functionality a bit more clear to end users

Barotrauma/BarotraumaShared/SharedSource/Events/Missions/GoToMission.cs
```diff
@@ -4,9 +4,21 @@ namespace Barotrauma
 {
     partial class GoToMission : Mission
     {
+        private readonly bool stateControlsCompletion;
```
Regalis11, line 7: To me this field seems unnecessary, this could work the same way as displayAsFailedState (= -1 means the value does nothing).

Barotrauma/BarotraumaShared/SharedSource/Events/Missions/GoToMission.cs
```diff
@@ -4,9 +4,21 @@ namespace Barotrauma
...
+        private readonly int successState;
+        private readonly int displayAsFailedState;
+
         public GoToMission(MissionPrefab prefab, Location[] locations, Submarine sub)
             : base(prefab, locations, sub)
         {
+            stateControlsCompletion = prefab.ConfigElement.GetAttributeBool("statecontrolscompletion", false);
+            successState = prefab.ConfigElement.GetAttributeInt("successstate", 1);
+            displayAsFailedState = prefab.ConfigElement.GetAttributeInt("displayasfailedstate", -1);
```
Regalis11, line 16: It's generally better to use `GetAttributeInt(nameof(displayAsFailedState))`, because it allows finding the reference to the field more easily, changes the argument automatically if you rename the field, and prevents doing any typos in the argument (a typo wouldn't compile).

Barotrauma/BarotraumaShared/SharedSource/Events/Missions/GoToMission.cs
```diff
@@ -4,9 +4,21 @@ namespace Barotrauma
 {
     partial class GoToMission : Mission
     {
+        private readonly bool stateControlsCompletion;
+        private readonly int successState;
+        private readonly int displayAsFailedState;
```
Regalis11, line 9: I would've still expected this to control whether the mission fails, not whether it _appears to_ fail to clients. The latter seems like it would be a very useful feature (you can control the failure state with scripted events for example), but I don't really see what the current behavior would be useful for.

Regalis11: Looking at your earlier comment, I think there might be a small misunderstanding here

> The mission implicitly fails if it's in the fail state as it's not in the success state at the end of the round

This is not the case. Not succeeding is not the same thing as failing: some missions have conditions for failure, such as if the hostages die in a rescue mission, but if you e.g. don't even try to do the objective, you'll neither fail or succeed. At the moment, goto missions can never fail (there's no failure condition).

Regalis11 (changes requested):
Spotted a couple more minor issues. I think the "failed state" behavior is still pretty strange, see comment.

Barotrauma/BarotraumaShared/SharedSource/Events/Missions/GoToMission.cs
dakkhuza: Good point.

Barotrauma/BarotraumaShared/SharedSource/Events/Missions/GoToMission.cs
dakkhuza: The intended use case is in conjunction with the state controlled success logic to display to clients that the mission has been failed. As you said, GoTo missions can never fail so using it without also using the state controlled success logic wouldn't make much sense.

dakkhuza: I've updated the logic to only run if the mission's success state is controlled by the state to hopefully make the purpose of failstate clearer as it only makes sense to use with GoTo missions that are controlled by the state.

dakkhuza:
Updated PR to address new concerns, changes:
- Use `nameof` for attributes (this is low key brilliant, I'll be updating the mission types in my mod to use this)
- `successState` now defaults to -1
- `StateControlsCompletion` is now a property, it's value is true if `successState` is set to anything other than -1
- `displayAsFailedState` -> `failState` and now only has an effect if `StateControlsCompletion` is true, as it's intended to only be used with the state controlled completion logic as otherwise it would be confusing with the mission completing at the end of the round even though it displays as failed to clients

Regalis11 (changes requested):
I think the behavior of `failState` is still confusing (or arguably even incorrect). `SuccessState` controls which state the mission needs to be in to succeed, but `failState` does not control which state it needs to be in to fail, as one might assume based on the name.

Could you clarify why you want it to behave like this, just making the mission display as if it has failed, instead of actually making it fail?

PR #12866, opened 2023-10-20
Add WearableOverlay

justinbrick:
feat: Add WearableOverlay property to Wearable to show an overlay when items are equipped

### DIFF
--- a/Barotrauma/BarotraumaClient/ClientSource/Items/Item.cs
+++ b/Barotrauma/BarotraumaClient/ClientSource/Items/Item.cs
@@ -1237,7 +1237,7 @@ namespace Barotrauma
         }
 
         public void DrawHUD(SpriteBatch spriteBatch, Camera cam, Character character)
-        {
+        { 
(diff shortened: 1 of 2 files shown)


justinbrick:
I forgot to add some context to this PR:
Currently, the only way to make an overlay appear when equipping items, is to apply a fake affliction with an overlay specified. With a WearableOverlay property, it can instead be directly specified through the Wearable component on an item, and will apply accordingly.
This looks like this:
```xml
<Wearable>
  <WearableOverlay texture="texture.png"/>
</Wearable>
```

Regalis11 (changes requested):
I think there are a couple of issues with the changes:
- I believe there's a potential memory leak here: the sprite doesn't seem to ever be disposed. That should most likely be done in `Wearable.RemoveComponentSpecific`.
- It also seems that the sprite is always drawn, regardless if the item is worn or not.
- If the item has multiple `Wearable` components (i.e. can be worn on multiple different limbs), only the first one is taken into account.

justinbrick:
Thanks for the feedback! I'll take a look this weekend when I've got a chance.

Issue #3604, opened 2020-08-20
docking with a shuttle lets you inside a ghost outpost instead of ending round on mission mode


bananaofhell:
- [x] I have searched the issue tracker to check if the issue has already been reported.

- [x] I have searched the issue tracker to check if the issue has already been reported.

**Description**
Using an escape shuttle thats attached to a submarine and docking at the end outpost does not end round

**Steps To Reproduce**
Go on a multiplayer server, Grab an escape shuttle thats attached to a submarine and fly to an outpost then dock, on a cargo mission, do not bring cargo with you

**Version**
Windows, v0.10.4.0

**Additional information**
I was not the server owner but the server owner was the one who docked

Regalis11:
This is intentional. The vessel with the most living players inside it needs to dock with the outpost for the round to finish. Otherwise it'd be easy to grief by stealing a shuttle and forcing the round to end by docking.

SpencerDawson:
What about survivors that escape with scooter? We had a fantastic session, where we were dropping like flies due to some sabotage while trying to salvage a falling sub going to crush depth. One of us got away as a lone survivor, made it to the station, and got inside, but the round wouldn't end. There needs to be some other conditions for a round to end outside of a ship docking.

Regalis11:
I completely agree that in that situation the round should've ended, and I assumed it would have if everyone in the sub was dead. Although, that's only if you're playing a single mission; in the campaign mode you always have to bring some vessel to the destination. I'll reopen this and take a closer look at the end conditions.

SpencerDawson:
In our scenario, there was a single crew member left alive, but they were paralyzed with no resources for us to fix him, so he eventually expired before the last player go the the station. This was, in fact, a campaign.

I noticed there is still a dummy terminal in the ghost station. We did technically complete our mission, which was kill a moloch, so perhaps letting there be an interface to acknowledge that the mission was completed trigger an jump to the station? Most missions give enough to buy a dugong, so perhaps just opening to a menu to purchase a submarine (or buy a new one if the credits are available).

In the event that a new sub can't be purchased, providing a trigger for a hard ending would be nice as well (just make it a big ol' "retire" button, you just survived hell!).

hubismal:
[...]

PR #12903, opened 2023-10-24
support multiple subs for ending the round

deltanedas:
Currently if you play with multiple subs (I am working on a mod that does this) you can only use the main sub to dock to the outpost.
This PR lets you use any player-team sub to end the round, meaning:
- You can use drones or escape pods as a last ditch effort to save the round facing certain death in the sub
- If your "sub" is actually multiple subs glued together, any of those sub parts can dock to it (basis of my mod)

PvP is unaffected and I believe pirates use a different team id, so pirates spawning at natural formation end zone shouldn't do anything to the round.

### DIFF
--- a/Barotrauma/BarotraumaServer/ServerSource/Networking/GameServer.cs
+++ b/Barotrauma/BarotraumaServer/ServerSource/Networking/GameServer.cs
@@ -415,16 +415,38 @@ namespace Barotrauma.Networking
                             !c.Character.IsDead && !c.Character.IsUnconscious &&
                             c.Character.Submarine != Level.Loaded.EndOutpost);
 
-                        //level finished if the sub is docked to the outpost
+                        //level finished if any player sub is docked to the outpost
                         //or very close and someone from the crew made it inside the outpost
-                        subAtLevelEnd =
-                            Submarine.MainSub.DockedTo.Contains(Level.Loaded.EndOutpost) ||
-                            (Submarine.MainSub.AtEndExit && charactersInsideOutpost > 0) ||
-                            (charactersInsideOutpost > charactersOutsideOutpost);
+                        if (charactersInsideOutpost > charactersOutsideOutpost)
+                        {
+                            subAtLevelEnd = true;
+                        }
+                        else
+                        {
+                            //all subs are checked you can have a drone/escape pod/etc dock to the outpost and end the round
+                            foreach (var sub in Submarine.Loaded)
+                            {
+                                if (sub.TeamID != Submarine.MainSub.TeamID)
+                                    continue;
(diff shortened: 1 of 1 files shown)


Fachoj:
cool feature will be to finish round if any sub docked with n% player-team crew on board.

Regalis11 (changes requested):
This does look like a potentially cool feature! I'm not entirely convinced the current approach is enough to implement this though. 

Does this handle swapping the submarine, i.e. that the vessel you dock with becomes the new main submarine and the others get left behind? It also seems that this is only implemented in the server-side round ending logic, in my opinion the behavior should be the same in both single player and multiplayer.

deltanedas:
> Does this handle swapping the submarine, i.e. that the vessel you dock with becomes the new main submarine and the others get left behind?

how i have it, non-main subs at the end area *should* not be left behind.

> It also seems that this is only implemented in the server-side round ending logic, in my opinion the behavior should be the same in both single player and multiplayer.

where would i add logic for singleplayer? i thought it was run on dedicated srrvers and solo play

Regalis11:
> how i have it, non-main subs at the end area should not be left behind.

What I mean is that if your main sub is laying at the ocean floor, and you make it to the end outpost on a drone or an escape pod, the main sub should in my opinion be left behind (and you'd need some way to retrieve/replace it). Otherwise it would be easy to abuse this feature by just always breezing through the level on a drone. This is probably not trivial to implement - it requires a lot of new logic, some UI work for the option to restore the main sub, and probably some changes to the saving and networking logic.

Which is why we still haven't gotten around to addressing [this years-old ticket about this feature](https://github.com/Regalis11/Barotrauma/issues/3604).

> where would i add logic for singleplayer? i thought it was run on dedicated srrvers and solo play

The code you've edited only affects the server, so it wouldn't work in solo play. I'm not entirely sure what would be the appropriate place to define the logic, but preferably the single player and multiplayer should reuse the same code. I believe figuring out (or _adding_) the proper place for it would be a part of the non-trivial amount of work in getting this feature implemented, and I'm afraid I currently don't have the time to start investigating it. :/

deltanedas:
> if your main sub is laying at the ocean floor, and you make it to the end outpost on a drone or an escape pod, the main sub should in my opinion be left behind

agreed, you should only keep all subs if they all make it (either individually piloted or docked together). if the main sub was not at the end zone a way similar to drone replacement could be used to replace any lost sub(s) but until then setting whatever sub docked as the main sub.

i assume that theres no shared client-server logic so that would take a lot of future work to get it going...

Regalis11:
Not mergeable due to the issues mentioned above, closing

PR #12906, opened 2023-10-24
Vanilla Events Overhaul

ubertpendragon:
Some vanilla events are played entirely through text, with a lack of visuals: _"tell, don't show"_

These events only happen for the one player who interacts with them. 

This aims to overhaul some of vanilla events by adding visual effects, NPCs or options that make the event more present in the world than just the text box.

Might add more events later

# Changes:

    Crawler outbreak
    -researcher runs past you to trigger the event
    -added spawnpoints to the research modules "testsubject", used to spawn crawlers
    -tagged a door "containementdoor", door breaks down
    
    Propaganda
    -added a visual and audio drawing effect
    -added a failure option
    -added text for failure option

    Firefighting
    -added smoke visuals for the trash can
    -gives a cigar if extinguished 
    
    Sound in the vent
    - added a requirement to equip screwdriver/flashlight to remove the grate (doesnt lock you out of the option if you dont have one))
    - changed text to reflect the above
    - added text for the above
    - increased trigger radius
    - spawns a transparent loose vent to make it interactable
    
    Captive souls
    - crawler spawns when breaking the glass + effects
    - crawler is aggressive
    - removed the text line about crawler going into a vent
    - looting curio is -2 reputation (make curio a stolen item instead?)

    Terrorism 101
    - NPC "operates" the vending machine  (BUGGED, disabled)
    - Arming the bomb spawns a detonator + locked UEX
    - UEX explodes after 15 seconds, makes a ticking sound and particles
    - added text lines to reflect the above
    - NPC flees to airlock
    - talk to NPC again to collect reputation
    - added text lines to reflect the above
    - disarming bomb gives wires
    - added text if you refuse to help with the bomb

    Stuck in the middle
    -spawns a handcuffed, bruised NPC
    -need to interact with NPC
    -added event sprites
    -TODO: change text to reflect the fact that the NPC isn't in a chair, or actually make the NPC able to pick the chair and sit somehow

    Young Cultists
    -Minor fix, changed trigger radius to interaction, the event triggered through walls before
    
    
# Videos:

_Not up to date with events/text but a good demonstration:_

https://github.com/Regalis11/Barotrauma/assets/73229309/00eb7ee7-aa16-400e-aff9-aa2c0b283caa

https://github.com/Regalis11/Barotrauma/assets/73229309/6bab917c-ce36-4f17-8d7c-6e816a15d169

https://github.com/Regalis11/Barotrauma/assets/73229309/da09f87d-92a0-4558-bcf7-ac8044c085e9

[...]

PR #12965, opened 2023-11-02
Explosive Mishap Event changes

ubertpendragon:
Just like events in #12906 `explosivemishap` is a very text heavy event that tells a lot but doesnt show. 

There's also a pretty big gap between what the text says and how it actually works (happens in the wrong module, explodes on you not the door). 
I've tried to make the door explode previously but without targetting the door specifically doing so is impossible

This salvages as much of the event as possible while making the necessary changes.

https://github.com/Regalis11/Barotrauma/assets/73229309/a27eacbd-c96a-43dc-b19b-e38f973abf76

**Changes:**
No mention of a door in text
Spawns an alien weapon on the floor and two researchers
Triggers on proximity with player
Explosion kills the researchers
Interact with the weapon to continue
Security runs to the explosion site
Security will talk to you if you picked up the weapon
Security tries to arrest you if you have the weapon
Can turn in the weapon to security


Regalis11:
Thank you for the contribution! Moved this to a PR in our private repo with some minor adjustments.

PR #13045, opened 2023-11-16
Make "fuel_percentage_left" pin output a float

Quavarx:
Made "fuel_percentage_left" pin on reactor output a float instead of an int (then to a string) for better precision. 

I just want to be able to display how much longer the fuel in the reactor will last👍

### DIFF
--- a/Barotrauma/BarotraumaShared/SharedSource/Items/Components/Machines/Reactor.cs
+++ b/Barotrauma/BarotraumaShared/SharedSource/Items/Components/Machines/Reactor.cs
@@ -378,7 +378,7 @@ namespace Barotrauma.Items.Components
             item.SendSignal(((int)-CurrPowerConsumption).ToString(), "power_value_out");
             item.SendSignal(((int)Load).ToString(), "load_value_out");
             item.SendSignal(((int)AvailableFuel).ToString(), "fuel_out");
-            item.SendSignal(((int)fuelLeft).ToString(), "fuel_percentage_left");
+            item.SendSignal((fuelLeft).ToString(), "fuel_percentage_left");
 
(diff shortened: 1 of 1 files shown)


Regalis11:

I'm not entirely sure whether we should change this though. Dealing with floats is much more difficult than integers: you can't e.g. make any assumptions of the length of the string when checking the output with other signal components, and equality checks will no longer work reliably because the float value might never be exactly, let's say "50" or "100".

And even if we wanted to change this, it would need to be done in a way that maintains backwards compatibility, and doesn't break every circuit that assumes the signal is an integer between 0 and 100. We also need to do the cast to a string using invariant culture, otherwise the output would depend on your system language.

MapleWheels:
What I would recommend is to make it a bool property set by the sub designer that defaults to false so it maintains backwards compatibility.

If true, try to make use of this function to format the float value output. The `NumberFormatInfo` runs as `Invariant Culture` with English as standard by default.

`public static string FormatToDecimalPlace(float value, int decimalPlaces) =>
        string.Format(new NumberFormatInfo() { NumberDecimalDigits = decimalPlaces }, "{0:F}", value);`

Example:
```
if (FuelRemainingWithDecimal)
{
    item.SendSignal(FormatToDecimalPlace(reactorFuelPercent, 2), "fuel_percentage_left");
}
else
{
    // vanilla process
}
```

Regalis11:
@MapleWheels The FormatToDecimalPlace method creates a new NumberFormatInfo object every time it's called, so I don't think that's a good way to go about this because it unnecessarily allocates memory that needs to be cleaned up by the garbage collector.

Regalis11:
Not mergeable in the current state, closing.

PR #13066, opened 2023-11-20
Bombscare event changes

ubertpendragon:
The event `bombscare` plays out entirely through text.
Happens ina random location, it mentions a bomb but said bomb doesn't even spawn, and both options defuse the bomb anyway. 

image

This changes the event to spawn an actual bomb, with a twist

_video too big for github, note that text isn't the proper one because overriding text isnt possible_

https://cdn.discordapp.com/attachments/668819422919524352/1175591635564900382/Desktop_2023.09.10_-_17.38.09.06_20.mp4?ex=656bca07&is=65595507&hm=2ab3f4a154cbea30d2ea237a58ea0313cf4a137d2d64dd81931690624f099e20&

Changes: 

-After talking with the guard, a random steel cabinet is selected
-Getting near the cabinet prompts a text
-A traitor time bomb spawns inside the cabinet and is set active
-Grab the bomb to continue event
-Trying to defuse the bomb has 2 outcomes
-On success, bomb is defused (same as before), can be given to guard
-On failure, bomb stops ticking for a moment then ticks twice as fast. 
-Added/Changed text to reflect the bomb changes


Regalis11:
Thank you for the contribution! Moved this to a PR in our private repo with some minor modifications.

PR #13071, opened 2023-11-22
Double-clicking mods in the mod settings menu now moves them to the top of the opposite list instead of to the bottom

SomeRandomNoobKekeke:
This seems more logical, since top mods take precedence over bottom ones, and if you move one to the very bottom, it will be instantly overridden - an action that makes no sense

+It'll be easier to find them if you move them to the top

https://github.com/Regalis11/Barotrauma/assets/122838333/0f95d73a-7621-4403-a405-6e43e541eb4b

https://github.com/Regalis11/Barotrauma/assets/122838333/1aced107-5bf5-47c2-8256-5aca9e47a420

Here's pack of mods with colored wrenches if you need them for testing
[3 mods with wrenches.zip](https://github.com/Regalis11/Barotrauma/files/13433852/3.mods.with.wrenches.zip)

and proof that top mods have priority

https://github.com/Regalis11/Barotrauma/assets/122838333/bb53fd0b-d333-4bc6-a179-5f79da33f706

### DIFF
--- a/Barotrauma/BarotraumaClient/ClientSource/Steam/WorkshopMenu/Mutable/InstalledTab.cs
+++ b/Barotrauma/BarotraumaClient/ClientSource/Steam/WorkshopMenu/Mutable/InstalledTab.cs
@@ -142,6 +142,7 @@ namespace Barotrauma.Steam
                 {
                     frame.Parent.RemoveChild(frame);
                     frame.RectTransform.Parent = to.Content.RectTransform;
+                    frame.RectTransform.SetAsFirstChild();
                 }
                 from.RecalculateChildren();
                 from.RectTransform.RecalculateScale(true);


Regalis11:
I'm not entirely sure about this. I can see the reasoning, but on the other hand, I think there are arguments for keeping it as-is too.

> This seems more logical, since top mods take precedence over bottom ones, and if you move one to the very bottom, it will be instantly overridden - an action that makes no sense

It only gets overridden if there is some conflicting content with the other mods. I would also argue it can make sense for it to get instantly override: if the user has mods enabled with content they want to keep enabled, I don't think we can assume the user always wants to override that content when they enable another mod.

I the current behavior is also much more common as far as UI design goes: moving elements between lists tends to move them to the end of the list.

SomeRandomNoobKekeke:
Yes, also true

So there are two cases: when you want mod to be overriden and when you don't
In one case you can double click, in other you must drag and drop

The problem is: drag and dropping mods one by one become more and more annoying the more mods you have

And just flipping this behavior will still cause pain in other case

So i think actual solution should be something like:
"Shift clicking mods in the mod settings menu now moves them to the top of the opposite list"

Regalis11:
Is it really that common that you have to drag and drop mods one by one? I.e. that every single mod you're using conflicts with another one, and that you need to manually reorganize all of them to make sure the ones you want take priority?

I generally don't play with that many mods enabled, so I honestly don't know how common that is. But when I've been testing mods, I've at most had to move one or two mods to get them to take priority over some other mod.

SomeRandomNoobKekeke:
It's not very common, i had mb 2-3 conflicting pairs of mods

I think it's noticeable in two cases:
If you encounter some bug, and you disabling all mods to figure out what causing it, and then you have to re-enable them in same order 

Or if you making mods, and making a lot of compatibility patches, like [this guy](https://steamcommunity.com/id/big_m_/myworkshopfiles/)

SomeRandomNoobKekeke:
> So i think actual solution should be something like:
> "Shift double clicking mods in the mod settings menu now moves them to the top of the opposite list"

It's bad idea because shift + double click will overlap with shift + click, same with ctrl
Alt + dclick? I don't know if guys with macbooks have alt, i can't test that and won't implement
And it's another hardcoded key combo to keep in mind
And it's counterintuitive, it's not normal for ui, you need to inform players somehow about this feature

I think it should be left as it is or be flipped
I can't speak for all community, i'm not even playing a lot recently so i guess we need a poll

SomeRandomNoobKekeke:
The people have spoken
almost 50% / 50%
https://www.reddit.com/r/Barotrauma/comments/19525zl/which_behavior_do_you_find_more_convenient/

so let's just keep it as it is, since they're so indecisive

i love democracy

PR #13172, opened 2023-12-15
Add beds as wandering targets depending on health

gunbuilderguy:
I've given the AI the ability to seek beds when idling, allowing them to heal during downtimes without being a constant drain on supplies for minor wounds, with the chance of choosing a bed increasing the more wounded they are every time a check passes.

To make it work, you'll need to add `tags="bed"` to the respective bunks and hospital beds in their XML files.
It essentially allows you to slowly heal NPCs at the cost of having to dissmiss them of any active duties like maintenance or operating machinery, which is a worthy tradeoff in my opinion as you trade away threat readiness for some weak, cost-free healing.

### DIFF
--- a/Barotrauma/BarotraumaShared/SharedSource/Characters/AI/Objectives/AIObjectiveIdle.cs
+++ b/Barotrauma/BarotraumaShared/SharedSource/Characters/AI/Objectives/AIObjectiveIdle.cs
@@ -71,7 +71,7 @@ namespace Barotrauma
 
         private Character tooCloseCharacter;
 
-        const float chairCheckInterval = 5.0f;
+        const float chairCheckInterval = 4.0f;
         private float chairCheckTimer;
 
         private float autonomousObjectiveRetryTimer = 10;
@@ -382,12 +382,18 @@ namespace Barotrauma
                     {
                         foreach (Item item in Item.ItemList)
                         {
-                            if (item.CurrentHull != currentHull || !item.HasTag(Tags.ChairItem)) { continue; }
-                            //not possible in vanilla game, but a mod might have holdable/attachable chairs
+                            
+                            if (item.CurrentHull != currentHull) { continue; }
+                            //not possible in vanilla game, but a mod might have holdable/attachable ones
                             if (item.ParentInventory != null || item.body is { Enabled: true }) { continue; } 
                             var controller = item.GetComponent<Controller>();
                             if (controller == null || controller.User != null) { continue; }
-                            item.TryInteract(character, forceSelectKey: true);
+                            //make it more likely to seek a bed when hurt, (68% chance after 10 checks at 80% health, 40% at 90%H 10th try, 5% at 90%H first try, 56% at 20%H first try...)
+                            if(item.HasTag(Tags.BedItem) && Rand.Value() > Math.Sqrt(character.HealthPercentage/100)){
+                                item.TryInteract(character, forceSelectKey: true);
+                            } else if (item.HasTag(Tags.ChairItem) && Rand.Value() > 0.4){
+                                item.TryInteract(character, forceSelectKey: true);
+                            }
                         }
                         chairCheckTimer = chairCheckInterval;
                     }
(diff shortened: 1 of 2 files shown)


Barotrauma/BarotraumaShared/SharedSource/Characters/AI/Objectives/AIObjectiveIdle.cs
```diff
@@ -382,12 +382,18 @@ public void Wander(float deltaTime)
                     {
                         foreach (Item item in Item.ItemList)
                         {
-                            if (item.CurrentHull != currentHull || !item.HasTag(Tags.ChairItem)) { continue; }
-                            //not possible in vanilla game, but a mod might have holdable/attachable chairs
+                            
+                            if (item.CurrentHull != currentHull) { continue; }
```
Regalis11, line 386: The removal of the Chair tag check is a problem for performance. Previously this line allowed the loop to skip non-chair items early, now it continues much further into the method and does the somewhat expensive `GetComponent` call (expensive for such a tight loop at least). We have done an optimization to this method in our private repo that makes this loop only search through items with the chair tag, since enumerating through the entire ItemList can get very expensive. That change conflicts with this one, so I think we'll need to wait until it's merged before working on this further.

Barotrauma/BarotraumaShared/SharedSource/Characters/AI/Objectives/AIObjectiveIdle.cs
```diff
@@ -382,12 +382,18 @@ public void Wander(float deltaTime)
...
+                            //not possible in vanilla game, but a mod might have holdable/attachable ones
                             if (item.ParentInventory != null || item.body is { Enabled: true }) { continue; } 
                             var controller = item.GetComponent<Controller>();
                             if (controller == null || controller.User != null) { continue; }
-                            item.TryInteract(character, forceSelectKey: true);
+                            //make it more likely to seek a bed when hurt, (68% chance after 10 checks at 80% health, 40% at 90%H 10th try, 5% at 90%H first try, 56% at 20%H first try...)
```
Regalis11, line 391: The comment is somewhat misleading, the probability doesn't effect the probability of `seeking` a bed/chair, but the probability of using one when there's one within interaction range.

Barotrauma/BarotraumaShared/SharedSource/Characters/AI/Objectives/AIObjectiveIdle.cs
```diff
@@ -382,12 +382,18 @@ public void Wander(float deltaTime)
...
+                            if(item.HasTag(Tags.BedItem) && Rand.Value() > Math.Sqrt(character.HealthPercentage/100)){
+                                item.TryInteract(character, forceSelectKey: true);
+                            } else if (item.HasTag(Tags.ChairItem) && Rand.Value() > 0.4){
+                                item.TryInteract(character, forceSelectKey: true);
+                            }
```
Regalis11, line 396: The formatting here is very inconsistent: sometimes there's a space after an if, sometimes not. We also use Allman style brackets in the codebase instead of K&R.

Barotrauma/BarotraumaShared/SharedSource/Characters/AI/Objectives/AIObjectiveIdle.cs
```diff
@@ -382,12 +382,18 @@ public void Wander(float deltaTime)
...
+                            //not possible in vanilla game, but a mod might have holdable/attachable ones
                             if (item.ParentInventory != null || item.body is { Enabled: true }) { continue; } 
                             var controller = item.GetComponent<Controller>();
                             if (controller == null || controller.User != null) { continue; }
-                            item.TryInteract(character, forceSelectKey: true);
+                            //make it more likely to seek a bed when hurt, (68% chance after 10 checks at 80% health, 40% at 90%H 10th try, 5% at 90%H first try, 56% at 20%H first try...)
+                            if(item.HasTag(Tags.BedItem) && Rand.Value() > Math.Sqrt(character.HealthPercentage/100)){
+                                item.TryInteract(character, forceSelectKey: true);
+                            } else if (item.HasTag(Tags.ChairItem) && Rand.Value() > 0.4){
+                                item.TryInteract(character, forceSelectKey: true);
```
Regalis11, line 395: Duplicate code, the exact same line is repeated twice

Regalis11 (changes requested):
Sorry for taking so long to take a look at this. I spotted some issues with the code (mainly formatting issues), but there's at least one thing here that could cause noticeable performance issues.

PR #13202, opened 2023-12-19
QOL Turret Ammunition

UnhappyWithMyUsername:
QOL Improvements to ammo boxes and railgun shells.

Changes:

1. Can now be carried with one hand. - Previously two-handed items.
    - Allows to carry up to two boxes/shells.

2. Added movement penalty for balance.
    - Slows down the character by 20% per box/shell carried.

3. Adjusted hold positions:
    - Ammo Boxes: Tucked under the arm.
    - Railgun Shells: Rested on the shoulder while walking. Held onto chest while swimming.
    
4. Adjusted Gun Runner talent:
    - Added "ammoboxslowdown" affliction resistance to the gun runner talent.

6. Adjusted sprites:
    - Fixed cut-off look on all ammo boxes
    - Merged all ammo box sprites into a single AmmoBoxes.png file.
 
QOLTurretAmmo
QOLRailgunShell


Content/Afflictions.xml
```diff
@@ -2756,4 +2756,24 @@
...
+    indicatorlimb="Torso"
+    maxstrength="2"
+    healableinmedicalclinic="false">
+    <Effect minstrength="0" maxstrength="1" strengthchange="-1">
+      <StatusEffect type="OnActive" target="Character" SpeedMultiplier="0.8" PropulsionSpeedMultiplier="0.8" setvalue="true" duration="0.1" />
+    </Effect>
+    <Effect minstrength="1" maxstrength="2" strengthchange="-2">
+      <StatusEffect type="OnActive" target="Character" SpeedMultiplier="0.6" PropulsionSpeedMultiplier="0.6" setvalue="true" duration="0.1" />
+    </Effect>
+  </Affliction>
```
Regalis11, line 2777: It's not clear to me why this is implemented using an affliction, couldn't the effects be in the items themselves?

Regalis11 (changes requested):
This looks great! I found one thing that seemed unclear to me, see the comment

Content/Afflictions.xml
ubertpendragon: Because the slowdown effects don't stack, Statuseffects will use setvalue

From moloch shield:

`      <StatusEffect type="OnActive" target="Character" SpeedMultiplier="0.6" setvalue="true" />`

Regalis11: Ah, I see! I think a cleaner solution would be to use stat values for this, because they do stack. Those only seem to be supported on Wearables though, but seems like it'd be easy to add support to holdable items as well.

UnhappyWithMyUsername: Yes, switching to stat values if they become functional in holdable components would definitely be the better solution.

Regalis11:
I've now implemented support for StatValues in holdables and opened a PR in our private repo. Thank you for your contribution!

I've still got a couple of questions though: I'm not sure what you mean by "Fixed cut-off look on all ammo boxes". I can see that the sourcerect on the coilgun ammo box is off by one pixel, but are there others too? It's also unclear what the reasoning behind moving the ammobox sprites to a separate texture was. It's better for performance to use texture atlases to reduce the number of textures, is there some benefit to moving them to a separate one in this case?

UnhappyWithMyUsername:
Very nice!
I made a gif to better visualize the change, but basically (at least to me) the boxes all looked like they were cut a few pixels too short on their right side when compared to how they look on their left side as if the sprite was originally wider at some point so I just made it match with how the left side is looking:
ammobox

As for putting them in a separate file was done for my convenience. I didn't think about the fact the game has to keep another texture in memory... So yeah it might be best to put them in the atlas again and so I did while reducing the spacing between the sprites and also adding the focused flak shell ammo box as that one was inside the TurretsAndDepthCharges.png for some reason. I guess it was done as to not change the sourcerects of previous ammo boxes for backwards-compatibility with mods? Though with these changes, mods altering the ammo boxes should probably update nonetheless anyways I'd say.

New containers.png:
containers

And one with sourcerects written on the ammo boxes for quick reference if you want to change the sourcerects in your private repo PR:
containers_ammobox_sourcerects

Regalis11:
This has been adapted to a PR in our private repo (https://github.com/FakeFishGames/Barotrauma-development/pull/4812), closing. Thank you for your contribution! :)

PR #13354, opened 2024-01-26
Door sprite depth changes

ubertpendragon:
There's too much of a difference between background and foreground depth for doors.

This means you don't have a choice: every item above 0.890 shows in front of the door background, this includes wires, cable holders, some walls... You can't change door depth if you want the background to hide those.

I simply changed the background to be closer to the foreground, so you can adjust the door depth if you want these things to show

On the left: vanilla door
On the right: modified door at different depths
image

Changes:

_For reference Foreground is 0.51_
-Changed door background depth to 0.61 (from 0.89)
-Changed door buttons depth to 0.62 (from 0.75)
-Added complete texture file path instead of just "door.png"


Regalis11 (changes requested):
If I'm not mistaken, this would affect the sprite depths in existing subs too. I think some kind of backwards compatibility logic would be necessary to prevent messing up the sprite depths if these are changed.

PR #13455, opened 2024-02-20
BeaconStation2 rework

NotWendy:
BeaconStation2 has always been the most broken of the vanilla beacons, so I tried to fix all of its bugs/ways in which players could accidentally break it beyond repair. Also gave it a visual refresh in the process.
Tested in the submarine editor and in the level editor. The circuits work, the enemies spawn and behave correctly. Have not forced it to spawn in a real campaign level.
Tested the waypoints by leading bots around the whole beacon, they handle it well.

Changes:
- Rewired the beacon's "puzzle" to be more intuitive for the player and for the modders looking at how the beacon is made. The wiring now only involves one relay and one And component. Relay turns on when the switch labeled as "reactor" is pressed, and the "water intake valve" lever is flipped. Just like in the old version, relay sends power to the main junction box from the alien reactor.
- Moved the interactable puzzle pieces to make them easier to spot.
- Locked the wiring of the parts of the puzzle that involve non-interactable components. (Ripping a wire out of the interactable end would make the beacon impossible to activate.)
- Made the main junction box interactable.
- Made the duct block above the station's fake reactor non-interactable, invulnerable, and hidden behind the shell.
- Made the hidden pump behind the reactor non-interactable.
- Made all wires connected to non-interactable devices invulnerable to damage, so they don't get automatically disconnected when the station is generated.
- Gave the two pumps at the bottom floor water detectors. They're on by default in the original design anyways.
- Made the airlock more compact, made the airlock's pump vulnerable to damage (since it's not longer hidden away), and made it only turn on when the outer door is closed.
- Used the space freed up by the simplified airlock to expand the "safe room" of the beacon. More room for drying meat, yum.

Images:
BeaconStation2


ubertpendragon:
This beacon has always rubbed me the wrong way because it's supposed to be a "starter" beacon but teaches players the wrong thing by not having a nuclear reactor

JeroonBosch:
Was merged already, closing

PR #13461, opened 2024-02-22
Orca rework

Waylonek:
First off English is not my native language, there may be translation inaccuracies.

My desire was to make vanilla sub more attractive without "over-decorate" it. The rework emphasizes decoration and combining elements with the wall pattern + fit wall's pattern borders in room.

### List of changes:
1. Removed separate docking compartment for being redundant (combined with airlock).
2. Removed gaps between ballasts (NBL preserved) - lower docking port moved to the right (stairs now lead to the airlock) - fabricator room expanded in size - depth charge loaders moved to a separate compartment - oxygen generator moved to the upper deck ( splitting of power consuming devices throughout the boat).
3. Pump output modified( 600 -> 420 ) to fix excessive pumping + the size of the pumps was reduced slightly. NBL saved.
4. Wiring became more branching (the first junction box on the left is responsible for powering the batteries and the next two junction boxes, which in turn are responsible for the next two pairs (in the same room, separated by a column) and the ones on the far right.
5. The searchlight has been moved to the bottom as it was not used at the top due to the presence of two turrets.
6. OR circuits moved to the circuit box + a switch added (to disable the annoying flood warning siren).
7. The lower docking bay now doubles as an airlock.
8. Expanded the living compartment vertically to reduce gaps between the shell and compartments (Crawlers do not cause damage to the compartment walls when chilling in before).
9. Drains now cover the entire boat evenly.
10. The firing angles of the bow and lower turret have been changed. The changes are made because there are no apparent restrictions on the angles. The lower turret is moved closer to the tail for the same reason (see screenshot).
11. Alpha level of Hull Ambient is set to 50 -> 15.
12. All doors and hatches now have buttons for more comfortable gameplay.
13. Removed labels.
14. Added a double diving suit locker at the bottom near the airlock (not counting the three at the top).
15. Interior and exterior lighting is presented in 3 colours (red, candle and light blue) with alpha 100-125, except ballast and  depth charge rooms with 75. Exterior and ballast lighting is slightly pulsing. Near the engine compartment the lights flicker a bit.
16. Fabricators are connected to adjacent cabinets (only one of them is visible, to avoid overlap). The navigation terminal is connected to the status monitor (when the status monitor is activated, it will open (the screen size is thus higher) for a more comfortable visibility.
17. Was made small correction of sub shell shape(small changs of shells size + a bit smaller upper docking floor) and fins.
18. Added easter egg for clown in crew room in box(hidden suppliens cabinet) with clown clothes under clown lable.

### Debatable design choices:
1. Side drain in the crew room (works on the greater component and two water detectors).
2. All relays for diving suits, lamps, battery array and docking hatch under juckbox(don't hide in walls). Relays h are "unholdable".
3. Depth charge room.
4. Turret angles change.

orca

orca_light

orca_compare

orca_angle


JeroonBosch:
Thank you Waylonek for your submission, but at this moment we're not looking to update the Orca to this extent. We feel that removing some of the features, such as removing the double airlock at the top, adding buttons to the bottom hatch or the hanging wires to the pumps remove some of it's identity. Therefor we decided to close this PR.

We'll be looking to implement some of the changes ourselves though, such as

> Alpha level of Hull Ambient is set to 50 -> 15.

PR #13667, opened 2024-03-25
Add distance_out pin to periscopes.

Jade-Harleyy:
Adds support for a "distance_out" pin to controllers, which outputs the distance between the focused object and the targeting reticle in meters.

### DIFF
--- a/Barotrauma/BarotraumaShared/SharedSource/Items/Components/Machines/Controller.cs
+++ b/Barotrauma/BarotraumaShared/SharedSource/Items/Components/Machines/Controller.cs
@@ -47,7 +47,7 @@ namespace Barotrauma.Items.Components
         private Character user;
 
         private Item focusTarget;
-        private float targetRotation;
+        private float targetRotation, targetDistance;
 
         public Vector2 UserPos
         {
@@ -388,6 +388,7 @@ namespace Barotrauma.Items.Components
                 Vector2 offset = character.CursorWorldPosition - centerPos;
                 offset.Y = -offset.Y;
                 targetRotation = MathUtils.WrapAngleTwoPi(MathUtils.VectorToAngle(offset));
+                targetDistance = offset.Length();
                 return false;
             }
 
@@ -413,6 +414,7 @@ namespace Barotrauma.Items.Components
                 Vector2 offset = character.CursorWorldPosition - centerPos;
                 offset.Y = -offset.Y;
                 targetRotation = MathUtils.WrapAngleTwoPi(MathUtils.VectorToAngle(offset));
+                targetDistance = offset.Length();
             }
             return true;
         }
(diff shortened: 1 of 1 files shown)


MasterReef:
yess please...

ssjarchon:
Amazing, thank you for doing this; I'll write a wiring and math guide for those that want to use it for the primary purpose I suggested on discord when this gets merged. In the mean time, I'll pull this branch and test it out when I get a chance (some time after this coming weekend.)

mygamingaccount:
I need this
:heart::heart::heart:

Barotrauma/BarotraumaShared/SharedSource/Items/Components/Machines/Controller.cs
```diff
@@ -423,6 +425,10 @@ public Item GetFocusTarget()
             if (positionOut == null) { return null; }
 
             item.SendSignal(new Signal(MathHelper.ToDegrees(targetRotation).ToString("G", CultureInfo.InvariantCulture), sender: user), positionOut);
+            if (item.Connections?.Find(c => c.Name == "distance_out") is { } distanceOut)
```
Regalis11, line 428: There is no reason to find the connection like this, the signal could instead be sent using the SendSignal override that takes in the name of the connection (which is also much faster than this, because the connection is fetched from a dictionary).

Barotrauma/BarotraumaShared/SharedSource/Items/Components/Machines/Controller.cs
```diff
@@ -423,6 +425,10 @@ public Item GetFocusTarget()
...
+            {
+                item.SendSignal(new Signal((targetDistance / 100).ToString("G", CultureInfo.InvariantCulture), sender: user), distanceOut);
```
Regalis11, line 430: It's not obvious why the distance is divided by 100 here. Using `ConvertUnits.ToSimUnits` would make it more clear..

Regalis11: This also creates new string objects every frame - it would be more efficient to just store the distance as a string when caching it.

Regalis11 (changes requested):
Thank you for the contribution! I found some minor issues with the changes, but I think I'll address those myself when merging this to our private repo.

Regalis11:
Merged to our private repo in https://github.com/FakeFishGames/Barotrauma-development/commit/80db478d7e825a9353872e40a51c5d6973c791e4

Regalis11:
We unfortunately had to revert the change, because it turned out this doesn't work in multiplayer, because the exact position of the player's cursor isn't synced, just the direction they're aiming towards.

https://github.com/FakeFishGames/Barotrauma/issues/14111

PR #13687, opened 2024-03-27
Temp fix to attack cancelling in closein stage of circle attack pattern

SomeRandomNoobKekeke:
See [this "issue"](https://github.com/FakeFishGames/Barotrauma/discussions/13601)

This is the easiest "fix" to that
Just don't reset CirclePhase to start onTargetChanged in CloseIn phase

It works perfectly for Charybdis
She always strike, and doesn't fallback if target hull is changed in CloseIn phase

https://github.com/FakeFishGames/Barotrauma/assets/122838333/ae59941a-e20d-4b35-9195-ff4f0a68e22b

I tested all of the vanilla potentially affected characters and i think they are behaving normally
Screenshot_1


itchyOwl:
Thank you for debugging it! I think the fix is correct. Created a PR in our private repo (FakeFishGames/Barotrauma-development/pull/5104) based on it.

PR #13754, opened 2024-04-08
Remove RepairTool progress bars from indestructible structures and items.

Jade-Harleyy:
This PR makes the progress/health bars of items and structures that appear when cutting/repairing not appear if the structure/item is unable to be damaged.
In addition, it adds extra debug info that tells if structures/items are invincible.

image

### DIFF
--- a/Barotrauma/BarotraumaClient/ClientSource/Items/Components/RepairTool.cs
+++ b/Barotrauma/BarotraumaClient/ClientSource/Items/Components/RepairTool.cs
@@ -76,19 +76,22 @@ namespace Barotrauma.Items.Components
 
         partial void FixStructureProjSpecific(Character user, float deltaTime, Structure targetStructure, int sectionIndex)
         {
-            Vector2 progressBarPos = targetStructure.SectionPosition(sectionIndex);
-            if (targetStructure.Submarine != null)
+            if (!targetStructure.Indestructible && targetStructure.Submarine is not { GodMode: true })
             {
-                progressBarPos += targetStructure.Submarine.DrawPosition;
+                Vector2 progressBarPos = targetStructure.SectionPosition(sectionIndex);
+                if (targetStructure.Submarine != null)
+                {
+                    progressBarPos += targetStructure.Submarine.DrawPosition;
+                }
+
+                var progressBar = user.UpdateHUDProgressBar(
+                    targetStructure.ID * 1000 + sectionIndex, //unique "identifier" for each wall section
+                    progressBarPos,
+                    MathUtils.InverseLerp(targetStructure.Prefab.MinHealth, targetStructure.Health, targetStructure.Health - targetStructure.SectionDamage(sectionIndex)),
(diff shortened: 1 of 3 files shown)


Regalis11:
I'm not actually entire sure about this change. I think it might be a good thing that the player gets some kind of feedback (a progressbar that won't go down) when they're trying to cut an indestructible wall, otherwise in some situations it could seem like the tool just isn't able to reach the wall.

Jade-Harleyy:
> it might be a good thing that the player gets some kind of feedback (a progressbar that won't go down) when they're trying to cut an indestructible wall

I'd say it's the opposite actually—_not_ having the progress bar appear tells players that nothing is happening, whereas _having_ the progress bar tells players that their action _is_ doing something, albiet slowly if they can't see the bar decreasing.

I've had this happen to both myself and people I've played with, where we'd try cutting walls for a longer than average time before finally realizing that the walls were invincible, despite the health bar not decreasing.

> in some situations it could seem like the tool just isn't able to reach the wall.

I think this is still communicated clearly by the particle effects that only occur when the beam touches a surface.

JeroonBosch:
> I think it might be a good thing that the player gets some kind of feedback

I agree with this, but not sure if

> (a progressbar that won't go down)

is the clearest way to do so, players might just think it's going really slowly.

> In addition, it adds extra debug info that tells if structures/items are invincible.

I think this would be more in the right direction, some text on the wall with the text "Indestructible" would be more clear than a healthbar. So overall I quite agree with this PR, except I'd not use "[Godmode]" as a text but rather "Indestructible" or similar

JeroonBosch:
Ah, I see now in the code it shows either [INDESTRUCTIBLE] or [GODMODE] depending on the reason. I don't think the brackets are needed, nor the capital letters but that's just some style preferences.

ubertpendragon:
I must agree with jade, showing the bar gives the impression that you can damage the wall. 
Sometimes those walls have very high health like ruins one, so you end up with confusion such as:

- waste time on a an invulnerable wall
- give up on cutting a very tough wall

Not showing the health bar clearly means you can't cut it

Regalis11:
Yeah, good points, I suppose this could be an improvement.

WrillWasTaken:
Since so many others are giving their grains of salt, I might as well give mine too. 

Not seeing a health bar is, and always has been indicative of you being out of range / obstructed instead of reaching the wall. Removing all indication of that would just make it unclear why the wall isn't breaking. On the other hand, it can be confusing with very very high health walls as ubertpendragon said if a healthbar shows and nothing changes.

But wouldn't all these issues be resolved by replacing the health bar with another symbol if a wall is indestructible? Replace it with a moderately sized red X perhaps? It's a different solution and probably out of scope of this PR, but to me it seems like the best solution possible.

Regalis11:
I agree with the comments by @JeroonBosch and @WrillWasTaken. Simply removing the progress bar does not seem like a good way to address this.

PR #13810, opened 2024-04-23
Debug Menu

Jade-Harleyy:
Adds entity lists and inspectors, a la UnityExplorer, for easier debugging.

Inspector mode - Accessed with the `debugmenu inspector` command; Click to view a list of entities below the cursor.
Entity explorers - Accessed with the `debugmenu` or `debugmenu explorer` commands, or via the Inspector mode; Click an entry to open an editor.
Too many windows? Use the `debugmenu close` or `debugmenu closeall` commands to clean up your view!

More features are planned such as an in-depth item spawner window and individual toggles for certain debug overlays, but I wanted to get this out now in case I never end up working on them.

image
image

### DIFF
--- a/Barotrauma/BarotraumaClient/ClientSource/DebugConsole.cs
+++ b/Barotrauma/BarotraumaClient/ClientSource/DebugConsole.cs
@@ -572,7 +572,7 @@ namespace Barotrauma
             {
                 GameMain.SpriteEditorScreen.Select();
             }));
-            
+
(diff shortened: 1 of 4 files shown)


kumilanka (changes requested):
This debug menus stuff is really great and I would like to see it included in the game. I would focus on considering if some of these features could be worked on separately instead of being a single package, would make it easier to review and test.

PR #13921, opened 2024-05-03
Allow resizing structures down to 2 cm instead of 16 cm.

Jade-Harleyy:
I've ran into this problem a couple times, where I needed to have a resizable structure smaller than 16 cm, but I wasn't allowed to make it any smaller. This PR fixes that.

I chose 2 cm instead of 1 cm since the handles would overlap at 1 cm.
image


mygamingaccount:
Grid size seems arbitrary, but then again so does 2. Why not 1?

Jade-Harleyy:
> Fitting to grid size seems like an arbitrary choice, but then again so does 2. Why not 1?

As I stated in the PR, I chose 2 as the minimum since, even when fully zoomed in, the resizing handles would overlap, making objects harder to work with.

lewri:
While I'm not a person accustomed to the Barotrauma code I do question whether Barotrauma has a policy to deter magic numbers. You've replaced a standardised variable (`Submarine.GridSize.X` / `Submarine.GridSize.Y`) with a magic number (`2`) without context here given to it. It may be beneficial to give the number a single referenced point that then equals 2.
(of course if magic numbers aren't an issue to the devs I'll stay quiet and run away again 😅)

mygamingaccount:
So basically `int minResizableItemSize = 2;`? Sounds fair.

Regalis11 (changes requested):
In my opinion it's a good thing that the sub editor by default tries to adhere to the grid when resizing. With this smaller size, if you for example placed a horizontal wall, it'd be very likely that the right side of the wall wouldn't match the grid, and it'd be difficult to align a vertical wall to the right side, or make hulls align with the walls.

You can freely resize the structure by holding shift anyway, so imo this change isn't necessary.

I also agree with the comments about the use of magic numbers. A constant should be used here to prevent repetition and to clarify what this magic number means.

PR #13933, opened 2024-05-07
Modlist Improvements:

Jade-Harleyy:
This PR:
- Adds options for renaming and deleting local mods.
- Adds an option to copy a workshop mod to a local mod.
- Fixes the context menu's 'unsubscribe' button not working.
- Cleans up context menu creation code, hopefully making it easier to read and work with.

### DIFF
--- a/Barotrauma/BarotraumaShared/SharedSource/Utils/SafeIO.cs
+++ b/Barotrauma/BarotraumaShared/SharedSource/Utils/SafeIO.cs
@@ -332,6 +332,53 @@ namespace Barotrauma.IO
         {
             return System.IO.Directory.GetLastWriteTime(path);
         }
+
+        public static void Copy(string src, string dest, bool overwrite = false)
+        {
+            if (!Validation.CanWrite(dest, true))
+            {
+                DebugConsole.ThrowError($"Cannot copy \"{src}\" to \"{dest}\": modifying the contents of the destination folder is not allowed.");
+                return;
+            }
+
+            CreateDirectory(dest);
+
+            foreach (string path in GetFiles(src))
+            {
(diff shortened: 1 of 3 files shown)


mygamingaccount:
Reviewing:

- when I try to rename a mod to the same name, it gives me an error (*A local package with the name or directory "Modname" already exists*)
- rename text box overlaps with the "Cancel" button

Barotrauma/BarotraumaShared/SharedSource/Utils/SafeIO.cs
```diff
@@ -362,8 +409,8 @@ public static void Move(string src, string dest)
                 DebugConsole.ThrowError($"Cannot move \"{src}\" to \"{dest}\": modifying the contents of the destination folder is not allowed");
                 return;
             }
-            System.IO.File.Move(src, dest);
-        }
+                System.IO.File.Move(src, dest);
+            }
```
mygamingaccount, line 413: Indent mistake

mygamingaccount:
I cherry picked this and now it feels awful when Baro updates and this feature is not there

Regalis11:
> Are you sure you want to delete [amount] mod[plural]?\nThis cannot be undone!

This will not work in all languages, plurals don't work the same way as they do in English in every language. For example, in Finnish the singular here would be "modin" and the plural would be "modia", and in many languages the rest of the sentence needs to change too when the word is in plural.

Regalis11:
I addressed the localization issues myself and moved this to [a PR in our private repo](https://github.com/FakeFishGames/Barotrauma-development/pull/5340). Thank you for your contribution once again! :)

Issue #13402, opened 2024-02-06
Cabinet sorting bug


SomeRandomNoobKekeke:
### Disclaimers

- [X] I have searched the issue tracker to check if the issue has already been reported.
- [ ] My issue happened while using mods.

### What happened?

Sorting in cabinets with some specific items constantly rearranges backpacks 

https://github.com/FakeFishGames/Barotrauma/assets/122838333/4e10d79e-4bce-481f-9249-fdae0ff24ca0

It's surprisingly hard to find a set of items that can cause this
So i just extracted a cabinet from my sub to an [item assembly](https://steamcommunity.com/sharedfiles/filedetails/?id=3155310490)

The language is english if it matters
Note that silicon also jumping in loops
Also works with duffle bags
I couldn't reproduce it with suitcases

### Reproduction steps

Take this [item assembly](https://steamcommunity.com/sharedfiles/filedetails/?id=3155310490)
Place it somewhere in sub editor
Press sort button
You don't even need to run test

### Bug prevalence

Happens regularly

### Single player or multiplayer?

Single player

### Version

v1.2.8.0 (Winter Update hotfix 2)

### Which operating system did you encounter this bug on?

Windows

Regalis11:
Fixed in https://github.com/FakeFishGames/Barotrauma-development/pull/4929

PR #13938, opened 2024-05-07
Fix Cabinet Sorting Bug #13402

egunde:
Fixed an issue where stacks of the same item in a cabinet would be rearranged when sorting alphabetically.

Added a function `IsContainer` to return whether or not an Item is an ItemContainer.

Also modified sorting to ignore any ItemContainer. Not sure if this is the ideal solution, but it was easier than figuring out a way to break ties between similar containers.

Edit: This is my first time contributing to something open source, so if I'm doing something wrong, feedback is appreciated.


Regalis11:
Thank you for the contribution! We've actually got a fix pending for this in our private repo, it's just waiting to be tested internally, so I'll close the PR.

PR #13948, opened 2024-05-10
Added requiredDestinationTypes on missionevents (solves #13941)

ubertpendragon:
Simply adds requiredDestinationTypes="outpost,city,research,military,mine" to all missionevents that give cargo/escort missions so they do not generate if there's no viable adjacent destination. 

No more of this:
218220031-6ebe2701-3360-4565-b76e-107c0caac5bc

_A side effect is that those missions will spawn less in practice as they will be picked at random by the eventset, but not trigger in the case of no adjacent outposts. If eventmanager generates its default 3 missionevents and 2 are cargo/escort, you will in practice only have 1 mission offered._

This is acceptable imo.

**NOTE: STILL NEEDS ANYOUTPOST SUPPORT FOR MODS.**
**THESE SHOULD BE USING requiredDestinationTypes="AnyOutpost" WITH THE ABILITY TO DEFINE IF A LOCATIONTYPE CAN BE USED BY ANYOUTPOST**

https://github.com/FakeFishGames/Barotrauma/discussions/13941

### DIFF
--- a/FactionEvents.xml
+++ b/FactionEvents.xml
@@ -868,11 +868,11 @@
     <!-- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -->
     <!--                          Church of the Husk                               -->
     <!-- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -->
-    <ScriptedEvent identifier="missionevent_escort1huskcult" commonness="100">
+    <ScriptedEvent identifier="missionevent_escort1huskcult" commonness="100" requiredDestinationTypes="outpost,city,research,military,mine">
       <ConversationAction text="EventText.missionevent_escort1huskcult.c1" speakertag="huskcultist" endeventifinterrupted="false" dialogtype="Small" />
       <MissionAction missiontag="escortcommonershuskcult" />
     </ScriptedEvent>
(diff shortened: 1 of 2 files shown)


Regalis11 (changes requested):
This does not seem to be in a mergeable state yet, as you commented

> NOTE: STILL NEEDS ANYOUTPOST SUPPORT FOR MODS.
THESE SHOULD BE USING requiredDestinationTypes="AnyOutpost" WITH THE ABILITY TO DEFINE IF A LOCATIONTYPE CAN BE USED BY ANYOUTPOST

ubertpendragon:
> This does not seem to be in a mergeable state yet, as you commented
> 
> > NOTE: STILL NEEDS ANYOUTPOST SUPPORT FOR MODS.
> > THESE SHOULD BE USING requiredDestinationTypes="AnyOutpost" WITH THE ABILITY TO DEFINE IF A LOCATIONTYPE CAN BE USED BY ANYOUTPOST

The PR itself does solve the issue for vanilla, but I'm not competent enough to make the necessary "AnyOutpost" changes that would ensure mod compatibility (for example, in DynamicEuropa, we add and override many location types, so we'd have to override the whole missionevents as well)

Regalis11:
Implemented this with AnyOutpost support in our private repo https://github.com/FakeFishGames/Barotrauma-development/pull/5341

PR #13977, opened 2024-05-19
Extra Missions

Jade-Harleyy:
Adds an XML tag to also start additional missions when a mission is started.
### Format:
- `ExtraMission`
  - `identifier`: Additional mission's identifier; takes priority over `tags`.
    - default: ""
    - type: String
  - `tags`: Additional mission's tags; selects a random mission that has any of the valid tags.
    - default: ""
    - type: String[]
  - `probability`: Chance for the mission to be added.
    - default: 1
    - type: float (0 to 1)
  - `mindifficulty`: Minimum level difficulty for the mission to be added.
    - default: 0
    - type: float (0 to 100)
  - `maxdifficulty`: Maximum level difficulty for the mission to be added.
    - default: 100
    - type: float (0 to 100)

Any amount of `ExtraMission` elements can be added.

### DIFF
--- a/Barotrauma/BarotraumaShared/SharedSource/Events/Missions/Mission.cs
+++ b/Barotrauma/BarotraumaShared/SharedSource/Events/Missions/Mission.cs
@@ -7,6 +7,7 @@ using System.Collections.Immutable;
 using System.Globalization;
 using System.Linq;
 using System.Xml.Linq;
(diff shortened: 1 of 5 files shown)


Jade-Harleyy:
Marked as draft since this currently doesn't work in campaign mode.

PR #14010, opened 2024-05-25
Moping Jack retexture

EebleSheembl:
Makes Moping Jack not just a bigger Hammerhead by giving it a custom sprite and ragdoll

Ragdoll changes are a harpoon limb and the default Moping Jack stats in the ragdoll


EebleSheembl:
image

Regalis11 (changes requested):
I feel the texture needs some more work. There's some strange blurry spots around the edges and the top fins. The texture also has a ton of unnecessary empty space.

EebleSheembl:
This was mainly a recolor of the normal hammerhead sprite with some damage added to it which would mean that those would probably be problems on the normal one too, though I might remake it as I was new to this whole thing and I could make it better now

EebleSheembl:
image

ubertpendragon:
> The texture also has a ton of unnecessary empty space.

I agree, there's empty space in the texture 😉 
image

EebleSheembl:
Yeah there's that one the normal hammerhead sprite, fixed it with this sprite though

Regalis11 (changes requested):
> This was mainly a recolor of the normal hammerhead sprite with some damage added to it which would mean that those would probably be problems on the normal one too

Just because there's issues in an existing texture doesn't mean we should add more with similar issues, especially when they'd be trivial to address. The texture still consists of mostly empty space, there's no reason to make it 1024px1024x when it would easily fit in half of that.

EebleSheembl:
Reduced the empty space

711885:
The changes that were requested were made, so is this considered? You should probably resolve the conversation/close the review

Regalis11:
Sorry for taking so long to get back to this! This looks good to me now with the latest changes, merged to our private repo. Thank you for your contribution!

PR #14019, opened 2024-05-26
Label Improvements

Jade-Harleyy:
image

### DIFF
--- a/Barotrauma/BarotraumaClient/ClientSource/Items/Components/ItemLabel.cs
+++ b/Barotrauma/BarotraumaClient/ClientSource/Items/Components/ItemLabel.cs
@@ -41,7 +41,7 @@ namespace Barotrauma.Items.Components
         }
 
         private Alignment alignment;
-        [Editable, Serialize(Alignment.TopLeft, IsPropertySaveable.Yes, description: "The alignment of the label's text.")]
+        [Editable, Serialize(Alignment.Center, IsPropertySaveable.Yes, description: "The alignment of the label's text.")]
         public Alignment TextAlignment
         {
             get => alignment;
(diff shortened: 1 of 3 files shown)


Libraries/BarotraumaLibs/BarotraumaCore/Utils/MathUtils.cs
```diff
@@ -10,7 +10,7 @@ namespace Barotrauma
     [Flags]
     public enum Alignment
     {
-        CenterX = 1, Left = 2, Right = 4, CenterY = 8, Top = 16, Bottom = 32,
+        Left = 1, CenterX = 2, Right = 4, Top = 8, CenterY = 16, Bottom = 32,
```
Regalis11, line 13: Why was this changed? I think it could have some unintended side-effects, such as things previously defaulting to CenterX defaulting to Left, which could cause for example UI layout issues in tons of places.

Barotrauma/BarotraumaClient/ClientSource/Items/Components/ItemLabel.cs
```diff
@@ -60,6 +68,16 @@ public string Text
...
 
+        /*
+        private GUIFont font;
+        [Editable, Serialize("UnscaledSmallFont", IsPropertySaveable.Yes, description: "The label's font.")]
+        public GUIFont Font
+        {
+            get => font;
+            set => font = textBlock.Font = value;
+        }
+        */
```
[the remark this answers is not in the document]
Jade-Harleyy: This and the next were commented out since I couldn't find a way to implement font selection.
I meant to come back to it, but I suppose I had forgot (until now).

Libraries/BarotraumaLibs/BarotraumaCore/Utils/MathUtils.cs
Jade-Harleyy: I believe I changed this one so the values would appear "in order" (left -> right, top -> bottom), although in hindsight I see this wasn't very necessary and can be reverted.

Regalis11:
This does not seem to be in a reviewable/mergeable state yet (font selection not implemented).

PR #14047, opened 2024-05-29
New Docking, Admin, and Engineering outpost modules

NotWendy:
First batch of new outpost modules. Includes 6 admin modules, 3 docking modules (1 for colonies), and 4 engineering modules.
Engineering_G3 and Engineering_G4 are remakes of vanilla modules. Made them recently when I started remaking all vanilla outposts. Add these if you wish. The rest of the modules are made from scratch.


AzureNocte:
Imo, unless Engineering modules have genes research station like vanilla ones, I don't think they should be merged.
Otherwise I suggest an overhaul to the design of all Engineering modules and Medical ones.

NotWendy:
Oh yea, shit, research stations. That is something worth looking into. I did offer remakes of vanilla modules without research stations in them, if we're going to go through with that change. If not, I can readd those.

JeroonBosch:
> I still very much advocate for letting research outposts have an identity of their own.

Yeah to me also they feel out of place in engineering modules. I'd personally favor making research stations limited to research outposts, colonies and some subs (Winterhalter only, in vanilla's case)

NotWendy:
We can consider the husk modules as another valid place for research stations. They only show up sometimes and the Church is all about biological/genetic research.

JeroonBosch:
Has been merged.

> We can consider the husk modules as another valid place for research stations. They only show up sometimes and the Church is all about biological/genetic research.

Also true

PR #14049, opened 2024-05-29
New Clown and Husk outpost modules

NotWendy:
Third batch of new outpost modules. Includes 4 clown modules and 2 husk modules, all made from scratch.
Sorry, Church of the Husk, I didn't have time to make two more modules for you.

This set of modules in particular is dear to my heart because it gives both factions reasons to exist at outposts.
Clowns set up bars and other places of entertainment in the least maintained parts of outposts, bringing in the money. Thus, they are tolerated by the outposts' management.
And husks establish small trading areas at outposts, selling valuable medical items. Whatever goes on in the backrooms of those areas is left unspoken, as long as the outpost gets the supply of drugs from the Church.


JeroonBosch:
Thanks, again, for your submission! These have been merged now to the private repo.

PR #14217, opened 2024-06-26
Fix NPCOperateItemAction Operate="false" not doing anything

dakkhuza:
Related to #14216 

Fixes the issue mentioned in the above discussion by implementing the suggested fix, that being changing `operateItemObjective.OperateTarget == target` to `operateItemObjective.Component.Item == target`

This could also be fixed by actually assigning the `OperateTarget` in the objective constructor, however I didn't want to change how the action operated and potentially cause unexpected compatibility issues.

Here's a mod to test that it's working correctly, this is the same mod in the above issue.
[NPCOperateAction.zip](https://github.com/user-attachments/files/15995089/NPCOperateAction.zip)

To test:
1. Load up the "NPCOperateItemAction" sub in the sub editor
2. Type `triggerevent npcoperateaction_example` into the console

Expected behavior with this pull request is the NPC cancels the operate objective correctly and starts following the player.


Regalis11:
Thank you for taking the time to investigate this! :) The fix looks correct to me and seems to work with no issues. Merged to our private repo in https://github.com/FakeFishGames/Barotrauma-development/commit/c782d90e6d05ff20cacc828d08b1e8c961ce7b6b.

PR #14238, opened 2024-06-29
README.md: Update Copyright date range to 2024

Next-Door-Tech:
Just noticed that README.md still had 2022 as the end of its copyright range.


Regalis11:
Thank you for the PR! We've now addressed this in our private repo.

PR #14257, opened 2024-07-05
Changes to make adding scopes possible

Jade-Harleyy:
This PR does two things:
1. Adds a new attribute to the Holdable component: `CameraAimOffset`
2. Reworks vision obstruction to work with long range sight.

## Obstruction Rework Previews
**Lighting Disabled**
image
**Lighting Disabled w/ Scope**
image
**Lighting Enabled**
image
**Lighting Enabled w/ Scope**
image

## Reworked light sprites
**Content/Lights/lightcone.png**

**Content/Lights/visioncircle.png**

### DIFF
--- a/Barotrauma/BarotraumaClient/ClientSource/Characters/Character.cs
+++ b/Barotrauma/BarotraumaClient/ClientSource/Characters/Character.cs
@@ -425,6 +425,10 @@ namespace Barotrauma
                 {
                     cam.OffsetAmount = targetOffsetAmount = item.Prefab.OffsetOnSelected * item.OffsetOnSelectedMultiplier;
                 }
+                else if (HeldItems.FirstOrDefault(item => item.GetComponent<Holdable>()?.Aimable ?? false) is Item item2 && IsKeyDown(InputType.Aim))
+                {
+                    cam.OffsetAmount = targetOffsetAmount = item2.GetComponent<Holdable>().CameraAimOffset;
+                }
                 else if (SelectedItem != null && ViewTarget == null &&
                     SelectedItem.Components.Any(ic => ic?.GuiFrame != null && ic.ShouldDrawHUD(this)))
                 {
--- a/Barotrauma/BarotraumaClient/ClientSource/Map/Lights/LightManager.cs
+++ b/Barotrauma/BarotraumaClient/ClientSource/Map/Lights/LightManager.cs
@@ -72,7 +72,7 @@ namespace Barotrauma.Lights
 
         public float ObstructVisionAmount;
 
-        private readonly Texture2D visionCircle;
+        private readonly Texture2D visionCircle, visionCone;
 
         private readonly Texture2D gapGlowTexture;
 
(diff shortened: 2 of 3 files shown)


Barotrauma/BarotraumaClient/ClientSource/Characters/Character.cs
```diff
@@ -425,6 +425,10 @@ void ResetInputIfPrimaryMouse(InputType inputType)
                 {
                     cam.OffsetAmount = targetOffsetAmount = item.Prefab.OffsetOnSelected * item.OffsetOnSelectedMultiplier;
                 }
+                else if (HeldItems.FirstOrDefault(item => item.GetComponent<Holdable>()?.Aimable ?? false) is Item item2 && IsKeyDown(InputType.Aim))
```
Regalis11, line 428: Pattern matching should be used instead of null coalescing here. The names `item` and `item2` are not very descriptive, and they're quite misleading because they're in this case the same item.

Barotrauma/BarotraumaClient/ClientSource/Characters/Character.cs
```diff
@@ -425,6 +425,10 @@ void ResetInputIfPrimaryMouse(InputType inputType)
...
+                {
+                    cam.OffsetAmount = targetOffsetAmount = item2.GetComponent<Holdable>().CameraAimOffset;
```
Regalis11, line 430: Fetching the `Holdable` component is done twice

Barotrauma/BarotraumaClient/ClientSource/Map/Lights/LightManager.cs
```diff
@@ -695,25 +696,18 @@ public void UpdateObstructVision(GraphicsDevice graphics, SpriteBatch spriteBatc
...
-                float originStartPosition = visionCircle.Width * relativeOriginStartPosition * MinHorizontalScale;
-                float relativeOriginLookAtPosModifier = -0.055f; //Increase this value increases how much the vision changes by moving the mouse
-                float originLookAtPosModifier = visionCircle.Width * relativeOriginLookAtPosModifier;
-
-                Vector2 scale = new Vector2(
-                    MathHelper.Clamp(losOffset.Length() / MaxOffset, MinHorizontalScale, MaxHorizontalScale), VerticalScale);
+                float scale = (1f - ObstructVisionAmount) * scaleMultiplier;
+                Vector2 circleOrigin = new Vector2(visionCircle.Width, visionCircle.Height) / 2f;
+                Vector2 coneOrigin = new Vector2(0f, visionCone.Height / 2f);
+                Vector2 coneScale = new Vector2(1f / visionCone.Width, scale / visionCone.Height * coneHeightMultiplier) * (losOffset.Length() + circleOrigin.X);
```
Regalis11, line 706: This seems to change the behavior of the existing vision circle too. I think this needs some clarification: how and why does it change the vision circle too?

Regalis11 (changes requested):
Found some code quality issues. It's also unclear to me why and how exactly the behavior of the vision circle logic was changed.

PR #14260, opened 2024-07-05
Fixed steam networking ping estimation being completely wrong on the server list

evilfactory:
# Old ping estimation
ping1
# New ping estimation
ping2
**(by the way, all the servers have high ping because I'm on Brazil)**

# Why old ping estimation gave wrong values
The old ping estimation only used https://partner.steamgames.com/doc/api/ISteamNetworkingUtils#EstimatePingTimeBetweenTwoLocations to estimate the ping between two locations, without combining it with GetLocalPingLocation to give the actual accurate results as pointed by the documentation. While also at it, steam API has https://partner.steamgames.com/doc/api/ISteamNetworkingUtils#EstimatePingTimeFromLocalHost which gives even more accurate results than combining GetLocalPingLocation and EstimatePingTimeBetweenTwoLocations according to the documentation, so the new ping code is now using that instead.

### DIFF
--- a/Barotrauma/BarotraumaClient/ClientSource/Networking/ServerList/PingUtils.cs
+++ b/Barotrauma/BarotraumaClient/ClientSource/Networking/ServerList/PingUtils.cs
@@ -144,9 +144,9 @@ namespace Barotrauma.Networking
 
             var pingLocation = NetPingLocation.TryParseFromString(pingLocationStr);
 
-            if (pingLocation.HasValue && Steamworks.SteamNetworkingUtils.LocalPingLocation.HasValue)
+            if (pingLocation.HasValue)
             {
-                int ping = Steamworks.SteamNetworkingUtils.LocalPingLocation.Value.EstimatePingTo(pingLocation.Value);
+                int ping = Steamworks.SteamNetworkingUtils.EstimatePingTo(pingLocation.Value);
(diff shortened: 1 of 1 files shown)


Regalis11:
The fix seems to work, but I don't quite understand why (I'm probably misunderstanding something about the documentation?).

> The old ping estimation only used https://partner.steamgames.com/doc/api/ISteamNetworkingUtils#EstimatePingTimeBetweenTwoLocations to estimate the ping between two locations, without combining it with GetLocalPingLocation to give the actual accurate results as pointed by the documentation.

The old ping estimation used estimated the ping between the remote location and `SteamNetworkingUtils.LocalPingLocation` which is fetched using `GetLocalPingLocation`, so I don't see why it's incorrect / why it returns incorrect results.

evilfactory:
Whoops, looks like I also misunderstood the documentation. Yeah the old ping logic was indeed using both functions correctly. I now don't know why the results were so inaccurate, since the documentation says that only on rare cases EstimatePingTimeFromLocalHost is more accurate than combining those both functions. 
> In rare cases this might return a slightly different estimate than combining GetLocalPingLocation with EstimatePingTimeBetweenTwoLocations. That's because this function uses a slightly more complete set of information about what route would be taken.

SomeRandomNoobKekeke:
AwareDev
Screenshot_2

Regalis11:
Merged this to our private repo now. Thank you for the contribution!

PR #14454, opened 2024-08-19
Burst-Fire Weapons

Jade-Harleyy:
This PR implements two types of burst-fire weapons:
- **Save**: The burst count will not reset when the trigger is released.
- **Forced**: The burst will continue even if the trigger is released.

They can be added to a RangedWeapon component using the following properties:
- **BurstMode** (BurstType) - The type of burst the weapon performs.
- **BurstCount** (int) - The amount of shots fired in a single burst.
- **BurstReload** (float) - How long the user has to wait before they can fire the weapon again after a burst (in seconds).
- **BurstReloadNoSkill** (float) - Burst reload time at 0 skill level. Reload time scales with skill level up to the Weapons skill requirement.

### DIFF
--- a/Barotrauma/BarotraumaShared/SharedSource/Items/Components/Holdable/RangedWeapon.cs
+++ b/Barotrauma/BarotraumaShared/SharedSource/Items/Components/Holdable/RangedWeapon.cs
@@ -184,7 +184,8 @@ namespace Barotrauma.Items.Components
         private bool tryingToCharge;
 
         private Character prevUser;
-        private int burstPos;
+        private int burstIndex;
+        private bool ShouldForceFire => BurstMode == BurstType.Forced && burstIndex > 0;
 
         public RangedWeapon(Item item, ContentXElement element)
             : base(item, element)
@@ -228,11 +229,11 @@ namespace Barotrauma.Items.Components
             if (ReloadTimer < 0.0f)
             {
                 ReloadTimer = 0.0f;
-                if (BurstMode == BurstType.Forced && burstPos > 0)
+                if (ShouldForceFire)
                 {
                     if (!HasRequiredContainedItems(prevUser, false))
                     {
-                        burstPos = 0;
+                        burstIndex = 0;
                     }
                     else if (Use(deltaTime, prevUser))
                     {
@@ -330,14 +331,14 @@ namespace Barotrauma.Items.Components
         private readonly List<Body> ignoredBodies = new List<Body>();
         public override bool Use(float deltaTime, Character character = null)
         {
-            if (character == null || character.Removed) return false;
-            if (ReloadTimer > 0) return false;
+            if (character == null || character.Removed) { return false; }
+            if (ReloadTimer > 0) { return false; }
 
-            if (BurstMode != BurstType.Forced || burstPos <= 0)
+            if (!ShouldForceFire)
             {
                 tryingToCharge = true;
-                if (item.RequireAimToUse && !character.IsKeyDown(InputType.Aim)) return false;
-                if (currentChargeTime < MaxChargeTime) return false;
+                if (item.RequireAimToUse && !character.IsKeyDown(InputType.Aim)) { return false; }
+                if (currentChargeTime < MaxChargeTime) { return false; }
             }
 
             IsActive = true;
@@ -416,10 +417,10 @@ namespace Barotrauma.Items.Components
 
             if (BurstMode != BurstType.None)
             {
-                burstPos++;
-                if (burstPos >= BurstCount)
+                burstIndex++;
+                if (burstIndex >= BurstCount)
                 {
-                    burstPos = 0;
+                    burstIndex = 0;
                     SetReloadTimer(character, BurstReload, BurstReloadNoSkill);
                 }
             }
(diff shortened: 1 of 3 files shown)


Barotrauma/BarotraumaShared/SharedSource/Items/Components/Holdable/RangedWeapon.cs
```diff
@@ -138,6 +183,9 @@ public Vector2 TransformedBarrelPos
         private float currentChargeTime;
         private bool tryingToCharge;
 
+        private Character prevUser;
+        private int burstPos;
```
Regalis11, line 187: Not clear what this means, needs a more descriptive name and/or comment. I think something like `burstCounter` or `burstIndex` would be more clear than `burstPos`.

Barotrauma/BarotraumaShared/SharedSource/Items/Components/Holdable/RangedWeapon.cs
```diff
@@ -233,34 +305,44 @@ private static float ApplyDualWieldPenaltyReduction(Character character, float o
...
+
+            ReloadTimer = newReload / (1 + character?.GetStatValue(StatTypes.RangedAttackSpeed) ?? 0f);
             ReloadTimer /= 1f + item.GetQualityModifier(Quality.StatType.FiringRateMultiplier);
+        }
 
+        private readonly List<Body> ignoredBodies = new List<Body>();
+        public override bool Use(float deltaTime, Character character = null)
+        {
+            if (character == null || character.Removed) return false;
+            if (ReloadTimer > 0) return false;
```
Regalis11, line 334: Curved brackets should be used in single-line if clauses too.

Regalis11 (changes requested):
Thank you for the contribution! I found some minor code quality issues that I feel should be addressed before I'm willing to approve of the PR.

Regalis11 (changes requested):
We took a look at this internally, and it seems I was a bit too hasty in approving this. I feel this will need quite a bit more work, and testing that all of it works as intended seems potentially so time-consuming that I unfortunately can't make any promises that we'd including this in the game even if all the issues found so far were addressed:

- The main issue is that `burstIndex` is not synced in multiplayer. If a clients' index is out of sync for whatever reason, for example if they join mid-round after the weapon has been fired, or if they miss a single shot due to networking issues, they'll reload timers will be completely off, leading to even worse syncing issues (e.g. the weapon going off client-side, but not server-side because the longer reload timer was active, or vice versa).
- It's impossible to create a burst-fire weapon that allows firing rapid single shots with this system. I.e. if the weapon is set to fire in bursts of three, if you fire 3 individual shots, you'll need to wait for the "longer" reload time. This doesn't feel that good to use: there's very little feedback to the player, so in practice it often just feels the gun doesn't fire for some reason. This also means weapons like the deadeye carbine cannot be implemented with this new system, which means we'd need to keep it unchanged, and in my opinion it's not a good thing if there are burst fire weapons implemented using two different systems (whose behavior is subtly different).
- There were some comments about the names of the burst fire modes being confusing, and in hindsight, I agree: if you look at an XML and see that a weapon's burst fire mode is "save", I think it'd be really hard to tell what it means unless you're familiar with how the feature works technically (that it "saves" the index of the burst).
- It's not obvious that the `BurstReload` and `BurstReloadNoSkill` values are affected by a separate property, not the "normal" weapons skill requirement that affects the normal reload. I don't quite see why a separate property is needed, but if it is, at the very least there should be a warning if the skill-based reload times have been configured without configuring the skill requirement.

Jade-Harleyy:
> It's not obvious that the `BurstReload` and `BurstReloadNoSkill` values are affected by a separate property, not the "normal" weapons skill requirement that affects the normal reload.

I'm not sure what you mean by this; `ReloadNoSkill` and `BurstReloadNoSkill` are both still dependent on `ReloadSkillRequirement`.

Regalis11 (changes requested):
Syncing the burst index doesn't work. There's no logic that would create any events for it. I also believe this approach would not work even if the events were created: the clients must be able to launch the projectile with no delay (= use client-side prediction) for the weapons to feel responsive. The network syncing however always has a delay. 

What this mean the burst index would be "out of sync" client-side: 
1. A client fires a gun, sets the burst index to let's say 1
2. The server notices the client fired the gun and sets it to 1 too and informs clients it was set to 1
3. The client keeps firing, sets the burst index to 2
4. The client receives an update from the server, saying the index is now 1
5. The end result being that the client will probably fire too many shots per burst, depending on the timing of the events.

I think addressing this will be very complicated and take a lot of testing to get working reliably, so even if you were to find a way to get it working, I can't make any promises that we'd be including it in the game.

Jade-Harleyy:
Even if the entire PR can't be implemented, I added a new property in the last commit that shouldn't have any problems, would be very useful, and could potentially be cherry-picked. :)

Regalis11 (changes requested):
The syncing issue I described in the previous review has not been addressed. I believe something like client-side prediction of the burst index with delayed corrections is needed to get it working properly.

Regalis11:
Closing for the reasons explained in the above reviews: this does not seem to be usable in it's current form in multiplayer.

PR #14470, opened 2024-08-22
Fix typo in item version update logic

Beyley:
Current code uses `.First` instead of `.FirstOrDefault` as was likely the intention, given the null check. Not 100% sure what exact XML causes this to bug happen, but this was hit with a heavily modded game.

image

mod list:
[lex!.xml.txt](https://github.com/user-attachments/files/16706977/lex.xml.txt)


Regalis11:
Thank you for the PR! This does seem like an oversight, and the fix looks correct and seems to work with no issues. Merged to our private repo in https://github.com/FakeFishGames/Barotrauma-development/commit/524d76d70efe73293ac03a2cb4e9706e26aac282

PR #14520, opened 2024-08-27
Add ForceSay to status effects

Emanthealmighty:
```
Added "ForceSay" command for status effects. 
ForceSay allows status effects to make a character say something in the text chat.

By default a forcesay message will be sent in local chat. Set "sayinradio" to true to try to send this message on a radio channel if applicable.
```

[Example mod](https://github.com/user-attachments/files/16766765/forcesaytestmod.zip)
[Example mod demonstration](https://youtu.be/OridMpt8vXU)

### DIFF
--- a/Barotrauma/BarotraumaShared/SharedSource/StatusEffects/StatusEffect.cs
+++ b/Barotrauma/BarotraumaShared/SharedSource/StatusEffects/StatusEffect.cs
@@ -616,6 +616,9 @@ namespace Barotrauma
 
         private readonly List<ItemSpawnInfo> spawnItems;
 
+        private readonly String forceSayIdentifier = null;
+        private readonly bool forceSayInRadio;
+
         /// <summary>
         /// If enabled, one of the items this effect is configured to spawn is selected randomly, as opposed to spawning all of them.
         /// </summary>
@@ -1133,6 +1136,10 @@ namespace Barotrauma
                         animationsToTrigger ??= new List<AnimLoadInfo>();
                         animationsToTrigger.Add(new AnimLoadInfo(animType, file, priority, expectedSpeciesNames.ToImmutableArray()));
                         
+                        break;
+                    case "forcesay":
+                        forceSayIdentifier = subElement.GetAttributeString("message", null);
+                        forceSayInRadio = subElement.GetAttributeBool("sayinradio", false);
                         break;
                 }
             }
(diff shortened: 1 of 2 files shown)


Barotrauma/BarotraumaShared/SharedSource/StatusEffects/StatusEffect.cs
```diff
@@ -1887,6 +1894,52 @@ protected void Apply(float deltaTime, Entity entity, IReadOnlyList<ISerializable
...
                 TryTriggerAnimation(target, entity);
 
+                if (forceSayIdentifier != null) 
+                {
+                    Character targetCharacter = null;
+                    if (target is Character character)
+                    {
+                        targetCharacter = character;
+                    }
+                    string messagetosay = null;
```
Regalis11, line 1904: Local variable names should be in camelCase, not all lower case.

Barotrauma/BarotraumaShared/SharedSource/StatusEffects/StatusEffect.cs
```diff
@@ -1887,6 +1894,52 @@ protected void Apply(float deltaTime, Entity entity, IReadOnlyList<ISerializable
                 
                 TryTriggerAnimation(target, entity);
 
+                if (forceSayIdentifier != null) 
+                {
+                    Character targetCharacter = null;
+                    if (target is Character character)
+                    {
+                        targetCharacter = character;
+                    }
```
Regalis11, line 1903: The above can be simplified to 
`Character targetCharacter = target as Character;`

Barotrauma/BarotraumaShared/SharedSource/StatusEffects/StatusEffect.cs
```diff
@@ -1887,6 +1894,52 @@ protected void Apply(float deltaTime, Entity entity, IReadOnlyList<ISerializable
...
+                            {
+                                GameMain.GameSession.CrewManager.AddSinglePlayerChatMessage(targetCharacter.DisplayName, modifiedMessage, messagetype, targetCharacter);
+                            }
+                            if (canUseRadio && forceSayInRadio)
+                            {
+                                Signal s = new Signal(modifiedMessage, sender: targetCharacter, source: radio.Item);
+                                radio.TransmitSignal(s, sentFromChat: true);
+                            }
+                            targetCharacter.ShowSpeechBubble(ChatMessage.MessageColor[(int)messagetype], messagetosay);
+                        }
```
Regalis11, line 1938: [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself). There's quite a lot of code duplication here. I also feel like a lot of this logic is already defined elsewhere. That means if we need to do changes to the chat message logic, we need to do remember to do it in multiple places.

Barotrauma/BarotraumaShared/SharedSource/StatusEffects/StatusEffect.cs
```diff
@@ -616,6 +616,9 @@ public IEnumerable<Explosion> Explosions
 
         private readonly List<ItemSpawnInfo> spawnItems;
 
+        private readonly String forceSayIdentifier = null;
+        private readonly bool forceSayInRadio;
```
Regalis11, line 620: These should be commented to allow them to be properly listed in the autogenerated modding documentation.

Regalis11 (changes requested):
Thank you for the contribution! This looks mostly fine, but I think there's way too much code duplication here for me to accept the PR. I also commented about some other minor issues.

Barotrauma/BarotraumaShared/SharedSource/StatusEffects/StatusEffect.cs
Emanthealmighty: Is there any way I can also add an automated description to the newly added "ForceSay" command as a whole or does that need to be done manually?

Regalis11: We have a script that manually parses the comments from the code for the modding doc, so these comments you added will do just fine. :)

Regalis11:
Moved to a PR in our private repo (https://github.com/FakeFishGames/Barotrauma-development/pull/5666), should be added to the vanilla game as soon as we manage to test this internally. Thank you for the contribution! :)

PR #14521, opened 2024-08-27
Add OnInserted and OnRemoved StatusEffect ActionTypes.

Jade-Harleyy:
This PR adds two new ActionTypes for StatusEffects:
- `OnInserted`
  - Executes once when a specific Containable is placed inside an ItemContainer.
  - Only valid for Containables defined in an ItemContainer component.
- `OnRemoved`
  - Executes once when a specific Containable is removed from an ItemContainer.
  - Only valid for Containables defined in an ItemContainer component.

### DIFF
--- a/Barotrauma/BarotraumaShared/SharedSource/Items/Components/ItemContainer.cs
+++ b/Barotrauma/BarotraumaShared/SharedSource/Items/Components/ItemContainer.cs
@@ -421,12 +421,11 @@ namespace Barotrauma.Items.Components
                         relatedItem ??= containableItem;
                         foreach (StatusEffect effect in containableItem.StatusEffects)
                         {
-                            activeContainedItems.Add(new ActiveContainedItem(
-                                containedItem, 
-                                effect, 
-                                containableItem.ExcludeBroken, 
-                                containableItem.ExcludeFullCondition,
-                                containableItem.BlameEquipperForDeath));
+                            ActiveContainedItem activeContainedItem = new(containedItem, effect, containableItem.ExcludeBroken, containableItem.ExcludeFullCondition, containableItem.BlameEquipperForDeath);
+                            activeContainedItems.Add(activeContainedItem);
+
+                            if (!TryProcessActiveContainedItem(activeContainedItem)) continue;
+                            activeContainedItem.StatusEffect.Apply(ActionType.OnInserted, 1, item, targets);
                         }
                     }
                 }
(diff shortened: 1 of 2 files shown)


Barotrauma/BarotraumaShared/SharedSource/Items/Components/ItemContainer.cs
```diff
@@ -657,37 +662,46 @@ void Inject(Item item)
...
                 effect.Apply(ActionType.OnContaining, deltaTime, item, targets);
-                if (wearing) { effect.Apply(ActionType.OnWearing, deltaTime, item, targets); }
+                if (item.GetComponent<Wearable>() is Wearable { IsActive: true })
+                {
+                    effect.Apply(ActionType.OnWearing, deltaTime, item, targets);
+                }
+            }
+        }
+
+        private bool TryProcessActiveContainedItem(ActiveContainedItem activeContainedItem)
```
Regalis11, line 679: I find the name of this method pretty undescriptive. Process in what way? What does the returned boolean represent? To me it doesn't seem like what it does could even be considered "processing" the contained item, it just seems to find the targets for the status effect. I think a more descriptive name and/or a comment explaining what this does is needed.

Barotrauma/BarotraumaShared/SharedSource/Items/Components/ItemContainer.cs
```diff
@@ -492,6 +491,12 @@ public override void Move(Vector2 amount, bool ignoreContacts = false)
 
         public void OnItemRemoved(Item containedItem)
         {
+            foreach (ActiveContainedItem activeContainedItem in activeContainedItems)
+            {
+                if (activeContainedItem.Item != containedItem || !TryProcessActiveContainedItem(activeContainedItem)) continue;
```
Regalis11, line 496: Single-line if clauses should also use curved brackets:

`if (activeContainedItem.Item != containedItem || !TryProcessActiveContainedItem(activeContainedItem)) { continue; }`

Barotrauma/BarotraumaShared/SharedSource/Items/Components/ItemContainer.cs
```diff
@@ -492,6 +491,12 @@ public override void Move(Vector2 amount, bool ignoreContacts = false)
...
+                activeContainedItem.StatusEffect.Apply(ActionType.OnRemoved, 1, item, targets);
```
Regalis11, line 497: Named arguments should be used if it's not obvious from the context what an argument means. Here it's not clear what the "1" means without looking at the method's signature.

Regalis11 (changes requested):
Thank you for the contribution! Functionally this looks fine, but I found some code quality issues I think should be addressed before I can approve of the PR.

Regalis11 (approved):
The changes look good to me now, merged to our private repo. :) Thanks, I'm sure many modders will love this addition!

PR #14583, opened 2024-09-06
Add "ConvulseAmount" affliction effect parameter

TheNinjaScout:
```
Added a new affliction effect parameter called "ConvulseAmount" which can be used to make characters convulse.

If an affliction's effect has a defined "ConvulseAmount" set to a float higher than zero, it will cause all of the afflicted character's limbs to shake around, similiarly to what happens to characters when they start dying from the husk infection, with the intensity of the shaking depending on the value of "ConvulseAmount".
```

https://github.com/user-attachments/assets/20264a5c-7495-4f63-b704-bb7d78da2514


SomeRandomNoobKekeke:
Now make 10 characters "convulse" at once and look at showperf

Emanthealmighty:
> Now make 10 characters "convulse" at once and look at showperf

Image on top shows 10+ characters convulsing around, second one is without the convulse affliction:
image



image

SomeRandomNoobKekeke:
Hmm weird
I thought this is exact copy of husk transformation, and husk transformation is very laggy
I thought it's laggy because of those convulsions, but perhaps there's some other reason

https://github.com/user-attachments/assets/6c4140f0-91b6-44f8-a0e0-c9e936f1c933

Emanthealmighty:
> Hmm weird I thought this is exact copy of husk transformation, and husk transformation is very laggy I thought it's laggy because of those convulsions, but perhaps there's some other reason
> 
>  10.husk.transformations.mp4

Here's 10+ characters convulsing due to husk infection, no FPS drop either. 
I manually spawned them in through console commands and gave them the affliction- maybe using events to do that causes the lag?

https://github.com/user-attachments/assets/009c757b-056f-4978-b2ee-1232bb056a72

SomeRandomNoobKekeke:
> maybe using events to do that causes the lag?

I don't think so

What is 1.5.0.2 unstable? :BaroDev:, is it unstable and they just forgot to change this label? or some custom version?
Mb it's just fixed there

Or maybe you have non potato pc

If you want to be sure you can take [this mod](https://steamcommunity.com/sharedfiles/filedetails/?id=3286179963), open "a test chamber" sub in sub editor and type triggerevent "husk lag" in console

Emanthealmighty:
> What is 1.5.9.2 unstable? :BaroDev:, is it unstable and they just forgot to change this label? or some custom version? Mb it's just fixed there
>
> If you want to be sure you can take [this mod](https://steamcommunity.com/sharedfiles/filedetails/?id=3286179963), open "a test chamber" sub in sub editor and type triggerevent "husk lag" in console


My "unstable version" is just me building the current repository as "unstable" instead of "release", it should in theory behave exactly the same as the current public steam version of the game, except for the modifications made in this PR. I'll experiment with your mod when I can and see what I can find out- there might be some sort of performance issue that only happens on the release builds an not on unstable.

SomeRandomNoobKekeke:
This seems to be caused by luatrauma directly 
BaroDev (wide)

PR #14599, opened 2024-09-09
Operators for Numerical Properties

Jade-Harleyy:
This PR lets modders use some operators when changing numerical property values with StatusEffects.
These operators are:
- `=`: Same as using `setValue`, but only for a single property.
- `+`: Same as using no operator and a positive number with `disableDeltaTime`, but only for a single property.
- `-`: Same as using no operator and a negative number with `disableDeltaTime`, but only for a single property.
(Negative numbers with no additional operators are still handled normally.)
- `*`: Sets the value to the original value multiplied by the given value.
- `/`: Sets the value to the original value divided by the given value.
- `^`: Sets the value to the original value to the power of the given value.
- `%`: Sets the value to the original value modulus the given value.

### DIFF
--- a/Barotrauma/BarotraumaShared/SharedSource/Serialization/SerializableProperty/SerializableProperty.cs
+++ b/Barotrauma/BarotraumaShared/SharedSource/Serialization/SerializableProperty/SerializableProperty.cs
@@ -303,10 +303,46 @@ namespace Barotrauma
                             case "identifierarray":
                                 PropertyInfo.SetValue(parentObject, XMLExtensions.ParseIdentifierArray((string)value));
                                 return true;
+                            case "int" or "float":
+                                PropertyInfo.SetValue(parentObject, ParseNumericalOperators((string)value));
+                                return true;
                             default:
                                 DebugConsole.ThrowError($"Failed to set the value of the property \"{Name}\" of \"{parentObject}\" to {value}");
                                 DebugConsole.ThrowError($"(Cannot convert a string to a {PropertyType})");
                                 return false;
+                        }
+
+                        float ParseNumericalOperators(string value)
+                        {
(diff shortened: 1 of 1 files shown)


Regalis11 (changes requested):
I don't think these changes work as intended. The operators are only handled in the `TrySetValue` overload that takes in an object, not the overloads that are used to set e.g. float or integer values. So in practice, I don't think this can be used by status effects.

I also feel parsing the operator should not be done on the fly like this for performance reasons, it should be done during initialization (of the status effect?).

Another issue here is how this plays together with the `setvalue` attribute. It's not at all clear to an user what happens when you try to both set a value and use one of the operators; there should probably be some kind of a warning if they are not supposed to be used together.

Barotrauma/BarotraumaShared/SharedSource/Serialization/SerializableProperty/SerializableProperty.cs
```diff
@@ -303,11 +303,47 @@ public bool TrySetValue(object parentObject, object value)
...
+                                return true;
                             default:
                                 DebugConsole.ThrowError($"Failed to set the value of the property \"{Name}\" of \"{parentObject}\" to {value}");
                                 DebugConsole.ThrowError($"(Cannot convert a string to a {PropertyType})");
                                 return false;
                         }
+
+                        float ParseNumericalOperators(string value)
+                        {
+                            float newVal = (float)PropertyInfo.GetValue(parentObject);
```
Regalis11, line 317: This will cause a crash if the value is an integer (you cannot cast a boxed integer to a float).

Regalis11 (changes requested):
Spotted one more issue I overlooked in the previous review

Regalis11:
This does not seem to be in a mergeable state, closing for now.

PR #14662, opened 2024-09-20
Add LocalMods to .gitignore

Jade-Harleyy:
This PR adds the content of the `Barotrauma/BarotraumaShared/LocalMods` folder to the .gitignore (excluding info.txt).

This helps with testing changes that require additional xml, without needing to about manually copy files to each output folder each time a change is made.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,5 +55,9 @@ Barotrauma/BarotraumaShared/Content/*
 Barotrauma/**/GameAnalyticsKeys.cs
 .github/ISSUE_TEMPLATE/release-checklist.md
 
+# Local mods
+Barotrauma/BarotraumaShared/LocalMods/*
+!Barotrauma/BarotraumaShared/LocalMods/info.txt
+
 #Rider
 *.DotSettings.user


Regalis11 (changes requested):
I don't quite understand the reasoning behind this change. How does this help with testing changes? Unless I'm missing something, .gitignore doesn't affect how files are copied from LocalMods to the output folder. If anything, this change makes testing more difficult because it makes impossible to commit or stash anything in the LocalMods folder.

Jade-Harleyy:
@Regalis11
When adding files to a repo in VS2022, each one appears as a changed file in the VS Git extension, making it difficult to tell what's changed that is important, and the files can be accidentally commited if not paying attention.
Additionally, I do not see the benefit over the cost of committing/stashing test mods vs. making new mods and enabling/disabling them in game.

Regalis11:
> When adding files to a repo in VS2022, each one appears as a changed file in the VS Git extension, making it difficult to tell what's changed that is important, and the files can be accidentally commited if not paying attention.

To me that sounds like an user issue - you could use some proper git client instead, and in my opinion you should always go through the files you're committing one-by-one to avoid accidentally committing something

> Additionally, I do not see the benefit over the cost of committing/stashing test mods vs. making new mods and enabling/disabling them in game.

I at least do that myself, and I believe many others in the dev team do too. Sometimes you just want to quickly hack something together in the LocalMods folder to test something, but not release them as actual mods, and maybe stash them away for future use instead of pushing them anywhere "public". 

All in all I feel like this PR essentially breaks a specific kind of workflow that's used by us devs (and I'd suspect, some modders too). If you want to prevent these files from showing up, you can edit the .gitignore locally.

PR #14673, opened 2024-09-21
Character flipping fixes and enhancements

TheNinjaScout:
```
Made changes to networking to properly sync characters' flipped state in a multiplayer server

Made player controlled monsters not turn in the direction of the mouse or turn instantly in multiplayer, instead they will need to walk in the opposite direction and wait for the animation's flip timers before turning around.

Added "flipcharacter" status effect command to make a character flip. Mainly only useful for non-humanoids, for example, to make them spawn already flipped through some conditions.

Made "spawncharacter" status effect command and husk infection spawn the new characters already flipped if the target character is also flipped.
```

Example video showcasing the changes including a modded status effect that transforms a crawler into a crawlerhusk on death, as well as showing what happens in the current stable build of the game:
https://youtu.be/bLEClFpJtZ0

### DIFF
--- a/Barotrauma/BarotraumaClient/ClientSource/Characters/CharacterNetworking.cs
+++ b/Barotrauma/BarotraumaClient/ClientSource/Characters/CharacterNetworking.cs
@@ -220,7 +220,16 @@ namespace Barotrauma
                 keys[(int)InputType.Ragdoll].Held = ragdollInput;
                 keys[(int)InputType.Ragdoll].SetState(false, ragdollInput);
 
-                facingRight = msg.ReadBoolean();
+            }
+
+            facingRight = msg.ReadBoolean();
+
+            if (AnimController.IsFlipped == facingRight)
+            {
+                if((AnimController is HumanoidAnimController && !CanMove) || AnimController is FishAnimController)
+                {
+                    TryFlipCharacter();
+                }
             }
 
             bool entitySelected = msg.ReadBoolean();
--- a/Barotrauma/BarotraumaServer/ServerSource/Characters/CharacterNetworking.cs
+++ b/Barotrauma/BarotraumaServer/ServerSource/Characters/CharacterNetworking.cs
@@ -386,8 +386,9 @@ namespace Barotrauma
 
                 tempBuffer.WriteBoolean(IsRagdolled || Stun > 0.0f || IsDead || IsIncapacitated);
 
-                tempBuffer.WriteBoolean(AnimController.Dir > 0.0f);
             }
+
+            tempBuffer.WriteBoolean(AnimController.Dir > 0.0f);
 
             if (SelectedCharacter != null || HasSelectedAnyItem)
             {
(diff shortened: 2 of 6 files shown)


PanmanS:
Very coolio PR

Barotrauma/BarotraumaClient/ClientSource/Characters/CharacterNetworking.cs
```diff
@@ -220,7 +220,16 @@ public void ClientReadPosition(IReadMessage msg, float sendingTime)
...
+            }
+
+            facingRight = msg.ReadBoolean();
+
+            if (AnimController.IsFlipped == facingRight)
+            {
+                if((AnimController is HumanoidAnimController && !CanMove) || AnimController is FishAnimController)
+                {
+                    TryFlipCharacter();
+                }
```
Regalis11, line 232: This seems like an incorrect way to handle this to me. The event we're reading here is not necessarily the latest one or received in order: that's why we add it to the `memState` buffer later in the method, and process that buffer in the proper order. I also feel this is not the correct place to be handling this - animation logic should be handled in the animation code, not a networking method that reads the remote state of the character.

I also find it unclear why we need to now add the facing to the buffer and handle it later, if it applies immediately.

Regalis11 (changes requested):
I think the fix here is flawed, see comments

Regalis11:
I went ahead an implemented a fix to the flipping syncing bug in our private repo.

> Fixed flipping monsters behaving inconsistently in multiplayer when controlling a monster: other clients saw the monster facing in the position of the cursor, even if it hadn't turned at all at the end of the client controlling it. There were two issues here:
> 
> Character.Control set the target dir based on the cursor position, disregarding if the character is a humanoid or a monster.
> The client controlling the monster reported the TargetDir of the monster to the server, but it doesn't do anything at the clients' end unless the monster moves. I think we can just report the actual direction the monster is facing as opposed to the direction it "wants to face".

TheNinjaScout:
> I went ahead an implemented a fix to the flipping syncing bug in our private repo.
> 
> > Fixed flipping monsters behaving inconsistently in multiplayer when controlling a monster: other clients saw the monster facing in the position of the cursor, even if it hadn't turned at all at the end of the client controlling it. There were two issues here:
> > Character.Control set the target dir based on the cursor position, disregarding if the character is a humanoid or a monster.
> > The client controlling the monster reported the TargetDir of the monster to the server, but it doesn't do anything at the clients' end unless the monster moves. I think we can just report the actual direction the monster is facing as opposed to the direction it "wants to face".

Can I reduce the contents of this PR to just the additions rather than fixes then? Namely, making a spawned character spawn already flipped if the character triggering the status effect or husk infection transformation is also flipped, as well as a status effect command to manually flip a character, in case we want to flip them in certain occasions, such as when spawning in.

Regalis11:
That sounds good, those seem like useful additions and I can't see any major problems in those changes.

PR #14812, opened 2024-10-12
Screenshake Slider

Jade-Harleyy:
Pretty much what the title says.
The game is much nicer to play when you can actually see and select the correct objects.

### New Localization
`ScreenShake`: Screenshake


Regalis11:
> The game is much nicer to play when you can actually see and select the correct objects.

To me this seems like an underlying issue that should be addressed, rather than allowing the shake to be disabled entirely.

PR #14829, opened 2024-10-14
New 'spawnnpc' and 'killall' Commands

Jade-Harleyy:
This PR adds two new console commands:
- `killall`: Immediately kills all characters in the level.
- `spawnnpc [npcsetidentifier] [npcidentifier] [near/inside/outside/cursor] [team (0-3)] [add to crew (true/false)]`: Spawns an NPC at a random spawnpoint. (Use the third parameter to select a specific set of spawnpoints.)


Regalis11:
Thank you for the contribution! I found some minor issues with the code: personally I think the heavy use of the ternary operator (as opposed to if-else clauses) makes some parts unnecessarily hard to read. I addressed those myself and moved this to a PR in out private repo.

PR #14844, opened 2024-10-16
Make smaller subs accelerate faster

mygamingaccount:
Every submarine in Barotrauma accelerates at the same rate because drag is calculated directly off the submarine's mass.
This is quite counter intuitive: I would definitely expect a tiny drone to achieve top speed much earlier than a container ship. It also makes shuttles drive like aßß, to which the suggested solution is to increase thrust, but that has an unwanted effect of raising the top speed.

To overcome these limitations, this change scales how hydrodynamic drag is calculated. It simply uses mass^⅔ and scales down the forces to remain 100% compatible.

Disclamer: I know that Update() is not the best place to calculate these factors, but I'd rather you gave it a try how the changes make the game feel before I spent more time researching a better place for it.

<details>
  <summary>(spoiler)</summary>

`horizontalDragCoefficient` and `verticalDragCoefficient`? They do absolutely nothing and are always 0.1

</details>

### DIFF
--- a/Barotrauma/BarotraumaShared/SharedSource/Items/Components/Machines/Engine.cs
+++ b/Barotrauma/BarotraumaShared/SharedSource/Items/Components/Machines/Engine.cs
@@ -3,6 +3,7 @@ using System;
 using System.Globalization;
 using System.Xml.Linq;
 using Barotrauma.Networking;
+using FarseerPhysics.Dynamics;
 
 namespace Barotrauma.Items.Components
 {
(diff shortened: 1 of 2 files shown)


Regalis11:
Have you considered calculating the drag "properly" based on the reference area of the sub instead? It could be cool if the shape of the sub affected the drag and the top speed.

mygamingaccount:
I guess that would be cool, yes. Intuition tells me that a sub that is more streamlined will tend to accelerate longer, and one that is thick will reach the top speed faster - at the same displacement. Meaning instead of size being the only factor, you could see different acceleration performance based on form. Interesting

Another case that I see where this could make a difference, is that of a submarine that has a variable geometry by the use of docking ports.

This is the reason I left `MassToAreaExponent` as a constant that can be changed: a lower value will exaggerate the effect done by this PR. I thought about serializing it too. But if the same effect could happen due to design decisions, that just feels a lot more engaging to me.

My fear is that it would incentivize streamlining, which might lead to more boring submarines overall. Currently builders can experiment with various shapes freely without concern for performance. This is one of those things that makes Barotrauma great, the crazy custom submarines.

jigardi:
@mygamingaccount

> My fear is that it would incentivize streamlining, which might lead to more boring submarines overall. Currently builders can experiment with various shapes freely without concern for performance. This is one of those things that makes Barotrauma great, the crazy custom submarines.
Would it be possible for it to be a per-submarine toggle?

This way you could get the best of both worlds; old subs would default to having it disabled, and sub builders could enable it if they wanted to.

mygamingaccount:
@jigardi setting `MassToAreaExponent` to be a serializable parameter of the sub would accomplish that.

I believe that existing subs equally benefit from this feature, so exempting them would really not be the best idea.

jigardi:
> @jigardi setting `MassToAreaExponent` to be a serializable parameter of the sub would accomplish that.
That's great news! : D
> I believe that existing subs equally benefit from this feature, so exempting them would really not be the best idea.
Good point!  I didn't consider that at all.  Hope this gets implemented!

PR #15133, opened 2024-11-13
Better Crosshair Accuracy

Jade-Harleyy:
This PR makes the crosshair (for RangedWeapons) more accurate by compensating for the barrel offset.


Regalis11:
Thank you for the contribution! I only spotted some very minor issues in the code (calculating values with double precision using `Math.Cos` and `Math.Sin` and then converting them to floats, instead of using `MathF.Cos` and `MathF.Sin`). Addressed those myself and moved this to a PR in our private repo.

PR #15216, opened 2024-11-21
Add "hideifnorecipe" bool to fabrication recipes

TheNinjaScout:
Adds "hideifnorecipe" boolean to fabrication recipe parameters, which, if set to true, makes the recipe not show up in fabricator recipe list if the character and anybody in their crew doesn't have the recipe unlocked through a talent. Only relevant if "requiresrecipe" is set to true.

(For the demonstrations, I set "hideifnorecipe" to true for the Funbringer "clownexosuit")
Demonstration 1: https://youtu.be/BIjkKoS6wag
Demonstration 2: https://youtu.be/CJD9t-N3ZDQ


Regalis11:
Thank you for the contribution! Merged to our private repo, coming up in the next unstable build and the upcoming release.

Issue #12772, opened 2023-10-07
[Unstable] Every time someone moves the text displayer GUI, all clients will disconnect from the server.


zhu-rengong:
### Disclaimers

- [X] I have searched the issue tracker to check if the issue has already been reported.
- [ ] My issue happened while using mods.

### What happened?

I moved the text displayer gui(Connection Panel), but I disconnected from the server and got:
```
Error while reading a message from server. 
Exception thrown while reading segment EntityEvent at position 32.
Inner exception: Message byte position incorrect after reading an event for the entity "Text Display (ID: 1196)" (ID 1196). The previous entity was "" (ID 0) Read 16 bits, expected message length was 80 bits.
   at Barotrauma.Networking.ClientEntityEventManager.Read(ServerNetSegment type, IReadMessage msg, Single sendingTime) in <DEV>\Barotrauma\BarotraumaClient\ClientSource\Networking\NetEntityEvent\ClientEntityEventManager.cs:line 233
   at Barotrauma.Networking.GameClient.<>c__DisplayClass118_0.<ReadIngameUpdate>b__0(ServerNetSegment segment, IReadMessage inc) in <DEV>\Barotrauma\BarotraumaClient\ClientSource\Networking\GameClient.cs:line 2120
   at Barotrauma.Networking.SegmentTableReader`1.Read(IReadMessage msg, SegmentDataReader segmentDataReader, ExceptionHandler exceptionHandler) in <DEV>\Barotrauma\BarotraumaShared\SharedSource\Utils\SegmentTable.cs:line 296
```

recent server log:
```
[10/07/2023 13:08:45]
  iivv has left the server.
[10/07/2023 13:08:46]
  iivv set the value "GuiFrameOffset" of the item "Text Display" to "-142,-72".
```

I think it's related to #10203 

### Reproduction steps

1. Host a server (I hosted a dedicated server)
2. Start a round
3. Use screw driver to interact with a text displayer and open its connection panel, then just move the gui frame.

### Bug prevalence

Happens every now and then

### Single player or multiplayer?

Multiplayer hosted using a dedicated server

### Version

Unstable (v1.1.14.0)

### Which operating system did you encounter this bug on?

Windows

Regalis11:
Fixed in https://github.com/Regalis11/Barotrauma-development/commit/1b40cd204926bfbe8a6675dc65dfe810e1fbd726 (commit in our private repo)

Jasontti:
[...]

Issue #10203, opened 2022-11-01
Adding Turret/ItemLabel before LightComponent and write light's properties' changes in server will cause desync.


zhu-rengong:
### Disclaimers

- [X] I have searched the issue tracker to check if the issue has already been reported.
- [X] My issue happened while using mods.

### What happened?

In the executing of `Item.WritePropertyChange` will gets all Editable properties of the specific item, and finds its related property index.
but i noticed in some cases that actual property which propertyIndex indicated is different in server and client.
Taking ItemLabel as an example, its property list looks like this:
|Editable Property Sequence|Server|Client|
|-|-|-|
1|TextColor|IgnoreLocalization|
2|TextScale|TextColor|
3||TextScale|

Assume I want the server to write TextScale to the client, but the client will actually sets its value to IgnoreLocalization

This problem doesn't occur in Vanilla, because Vanilla only provides the ability for the client to write `Item.WritePropertyChange` to the server, and the properties are limited to InGameEditable, and coincidentally, the client and server have the same sequence of InGameEditable-properties.

### Bug prevalence

Happens every now and then

### Version

0.19.14.0

### Which operating system did you encounter this bug on?

Windows

Regalis11:
This happens because the property in question is only implemented for clients, and doesn't exist server-side. 

I wouldn't consider this a bug, as non-ingame-editable properties cannot be edited in-game by clients in the vanilla game. If you want to change that, I think you need to come up with some other way to sync them than using the index of the property: perhaps the name? Or ensure that all Editable properties exist both client-side and server-side.

zhu-rengong:
The server always synchronizes some non-ingame-editable properties of the headset to the clients, if some modder override headset such as adding an ItemLabel for it, this will happens, and there're other ways which also can causes this. This could be a potential bug in the future, but as for now it's not a bug. Maybe it's better to look property up by name rather then index? But at the moment I don't have any ways to get all players having modding client installed, so I'd love to see there's a way to deal with this in the future.

PR #15256, opened 2024-11-26
Label Improvements

Jade-Harleyy:
This PR allows submakers and modders to:
- Change the padding of ItemLabels
- Change the alignment of ItemLabels
- Change the font of ItemLabels


zhu-rengong:
Writing changes mid-game via code modding may cause the client to asynchronize with the server and disconnect. This is mainly due to #12772 #10203 ; so here it needs to do the same thing in the server source, make `Padding` and `Alignment` editable, and add the `Font` property in the server project. There is a command `listeditableproperties` used to check for this.

Barotrauma/BarotraumaClient/ClientSource/Serialization/SerializableEntityEditor.cs
```diff
@@ -523,58 +523,62 @@ public GUIComponent CreateNewField(SerializableProperty property, ISerializableE
...
-                if (!Fields.ContainsKey(property.Name)) { Fields.Add(property.Name.ToIdentifier(), new GUIComponent[] { propertyTickBox }); }
-                return propertyTickBox;
+                    };
+                    refresh += () =>
+                    {
+                        propertyTickBox.Selected = (bool)property.GetValue(entity);
+                    };
+                    if (!Fields.ContainsKey(property.Name)) { Fields.Add(property.Name.ToIdentifier(), new GUIComponent[] { propertyTickBox }); }
+                    return propertyTickBox;
+                }
```
Regalis11, line 580: Unclear why this method was modified (seems unrelated to the rest of the changes in the PR) or what was actually modified about it.

Regalis11 (changes requested):
Spotted one oddity in the changes. I also believe there's an issue with syncing the new values in multiplayer [as zhu-rengong commented](https://github.com/FakeFishGames/Barotrauma/pull/15256#issuecomment-2503111580): since they only exist client-side, trying to sync properties of the label will now cause errors in multiplayer due to the mismatching number of editable properties between the server and the clients.

Barotrauma/BarotraumaClient/ClientSource/Serialization/SerializableEntityEditor.cs
Jade-Harleyy: Odd. I've manually compared the files between the main branch and my branch and there are no changes here, so it may be an issue with Git?

Barotrauma/BarotraumaServer/ServerSource/Items/Components/ItemLabel.cs
```diff
@@ -39,13 +39,19 @@ public float TextScale
...
         {
             get;
             set;
         }
 
+        [Editable, Serialize(Alignment.Center, IsPropertySaveable.Yes, description: "The alignment of the label's text.")]
+        public Alignment TextAlignment { get; set; }
+
+        [Editable, Serialize("UnscaledSmallFont", IsPropertySaveable.Yes, "The label's font.")]
+        public GUIFont Font { get; set; }
```
Regalis11, line 53: It seems potentially error-prone to me that we'd maintain two sets of these properties. It would be very easy for someone to do the mistake of editing these in the client project without realizing the changes need to be done to the server project as well (because surely UI stuff like this is client-only, so there's no need to touch the server code?). This also introduces some unnecessary code duplication.

I think it'd be preferable to define these in the shared project.

Regalis11 (changes requested):
I feel there are still some issues with the way the new properties have been implemented. I also don't fully understand the `GUIBridge` class. I would recommend syncing the font in some other way: it doesn't seem ideal to me to add this sort of a "mock" class that doesn't actually fully work outside this specific context.

Barotrauma/BarotraumaServer/ServerSource/Items/Components/ItemLabel.cs
Jade-Harleyy: Done, all duplicated properties are now defined in the shared source.

Barotrauma/BarotraumaServer/ServerSource/Networking/GUIBridge.cs
```diff
@@ -0,0 +1,14 @@
...
+    public static class GUIStyle
+    {
+        public readonly static ImmutableDictionary<Identifier, GUIFont> Fonts;
+    }
+
+    public class GUIFont
+    {
+        public readonly Identifier Identifier;
+    }
+}
```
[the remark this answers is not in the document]
Jade-Harleyy: I changed to defining them as partial classes in the shared source, and included the addition of fonts as well.

PR #15400, opened 2024-12-19
Fix duplicate message box when using forcesay in multiplayer

TheNinjaScout:
Since the messages are replicated to all players in multiplayer anyway, we don't need to run the clientside message code at all unless it's singleplayer.

Fixes https://github.com/FakeFishGames/Barotrauma/discussions/15395


Regalis11:
Thank you for the PR! Can't see any issues with the fix, merged to our private repo.

PR #15456, opened 2024-12-30
Fix custom sprite deformation not mirroring when character flips

TheNinjaScout:
Adds "isFlipped" parameter to all "GetDeformation" methods, only used in "CustomDeformation" for now to read the deformation rows in an inverse order when the character is flipped

Fixes https://github.com/FakeFishGames/Barotrauma/discussions/15434

https://github.com/user-attachments/assets/cc04ced6-b605-4693-b7a7-662ad6b39cf5

Without the fix ↓:

https://github.com/user-attachments/assets/090ec2e5-c7c1-459f-8caa-d12b96c7603b

### DIFF
--- a/Barotrauma/BarotraumaClient/ClientSource/Sprite/DeformAnimations/CustomDeformation.cs
+++ b/Barotrauma/BarotraumaClient/ClientSource/Sprite/DeformAnimations/CustomDeformation.cs
@@ -38,6 +38,8 @@ namespace Barotrauma.SpriteDeformations
         }
         private float phase;
 
+        private Vector2[,] FlippedDeformation;
+
         public CustomDeformation(XElement element) : base(element, new CustomDeformationParams(element))
         {
             phase = Rand.Range(0.0f, MathHelper.TwoPi);
@@ -107,25 +109,24 @@ namespace Barotrauma.SpriteDeformations
                         (normalizedY % divY) / divY);
                 }
             }
+            int width = Deformation.GetLength(0);
+            int height = Deformation.GetLength(1);
+            FlippedDeformation = new Vector2[width, height];
+
+            for (int x = 0; x < width; x++)
+            {
+                for (int y = 0; y < height; y++)
+                {
+                    FlippedDeformation[x, y] = Deformation[width - x - 1, y]; // read the rows from right to left
+                }
+            }
         }
 
         protected override void GetDeformation(out Vector2[,] deformation, out float multiplier, bool inverse, bool isFlipped = false)
         {
             if (isFlipped)
             {
-                Vector2[,] flippedDeformation = new Vector2[Deformation.GetLength(0), Deformation.GetLength(1)];
-
-                int width = Deformation.GetLength(0);
-                int height = Deformation.GetLength(1);
-
-                for (int x = 0; x < width; x++)
-                {
-                    for (int y = 0; y < height; y++)
-                    {
-                        flippedDeformation[x, y] = Deformation[width - x - 1, y]; // read the rows from right to left
-                    }
-                }
-                deformation = flippedDeformation;
+                deformation = FlippedDeformation;
             }
             else
             {
--- a/Barotrauma/BarotraumaClient/ClientSource/Sprite/DeformAnimations/CustomDeformation.cs
+++ b/Barotrauma/BarotraumaClient/ClientSource/Sprite/DeformAnimations/CustomDeformation.cs
@@ -109,9 +109,28 @@ namespace Barotrauma.SpriteDeformations
             }
         }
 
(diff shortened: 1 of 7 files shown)


Barotrauma/BarotraumaClient/ClientSource/Sprite/DeformAnimations/CustomDeformation.cs
```diff
@@ -109,9 +109,28 @@ public override float Phase
             }
         }
 
-        protected override void GetDeformation(out Vector2[,] deformation, out float multiplier, bool inverse)
+        protected override void GetDeformation(out Vector2[,] deformation, out float multiplier, bool inverse, bool isFlipped = false)
```
Regalis11, line 112: The difference between `inverse` and `isFlipped` seems unclear to me.

Barotrauma/BarotraumaClient/ClientSource/Sprite/DeformAnimations/CustomDeformation.cs
```diff
@@ -109,9 +109,28 @@ public override float Phase
...
         {
-            deformation = Deformation;
+            if (isFlipped)
+            {
+                Vector2[,] flippedDeformation = new Vector2[Deformation.GetLength(0), Deformation.GetLength(1)];
```
Regalis11, line 116: This causes a new array to be initialized every time the method is called. That's not good for performance: the flipped array should be cached and reused (perhaps it could be created in the constructor already).

Barotrauma/BarotraumaClient/ClientSource/Sprite/DeformAnimations/CustomDeformation.cs
```diff
@@ -109,9 +109,28 @@ public override float Phase
...
+
+                int width = Deformation.GetLength(0);
+                int height = Deformation.GetLength(1);
```
Regalis11, line 119: Would make sense to store the width and height before creating the array and reuse the values, instead of getting the lengths twice.

Regalis11 (changes requested):
I think there are a couple of issues here, see comments

Barotrauma/BarotraumaClient/ClientSource/Sprite/DeformAnimations/CustomDeformation.cs
TheNinjaScout: Inverse seems to only ever be used by customdeformation even though it's defined as a parameter for all deformations,

https://github.com/FakeFishGames/Barotrauma/blob/38ca721cb0fd94aadbb9c1886ba2338bbd5f1180/Barotrauma/BarotraumaClient/ClientSource/Sprite/DeformAnimations/CustomDeformation.cs#L112-L119

https://github.com/FakeFishGames/Barotrauma/blob/38ca721cb0fd94aadbb9c1886ba2338bbd5f1180/Barotrauma/BarotraumaClient/ClientSource/Sprite/DeformAnimations/SpriteDeformation.cs#L225-L233

 **and** it's only ever set to true when getting the deformation for deformable light sprites and not normal ones, 
https://github.com/FakeFishGames/Barotrauma/blob/38ca721cb0fd94aadbb9c1886ba2338bbd5f1180/Barotrauma/BarotraumaClient/ClientSource/Characters/Limb.cs#L831

after some testing I found out that deformable light sprites with custom deformation actually appear to deform correctly when the character is flipped in the current public build of the game, 


https://github.com/user-attachments/assets/6c40decc-44ba-441e-bf55-d0b3536599b6



If we set inverse to true for normal deformable sprites
https://github.com/FakeFishGames/Barotrauma/blob/38ca721cb0fd94aadbb9c1886ba2338bbd5f1180/Barotrauma/BarotraumaClient/ClientSource/Characters/Limb.cs#L827

like it is being done for light deformable sprites based on the direction of the character results in this happening, 


https://github.com/user-attachments/assets/c1062e27-f6da-4bca-b804-bb8bd0fee242


so I'm assuming inverse was made as a solution for light deformable sprites, but not normal deformable ones?

TheNinjaScout:
The custom deformation behaves a bit differently when comparing a non flipped limb to a flipped limb, not sure if it's a problem with my fix implementation or something else:

https://github.com/user-attachments/assets/c41702f6-6f49-428d-a9ba-d598b9e4db51

Regalis11:
Thanks for the clarification! I'm still a bit unhappy about `inverse` and `isFlipped`, it's not at all obvious how they differ - but I couldn't quite figure out either why `inverse` works the way it does and why it should be used. Moved this to a PR in our private repo (https://github.com/FakeFishGames/Barotrauma-development/pull/6231) with some additional comments on those.

PR #15469, opened 2025-01-04
Submarine Editor Transform Tool

Jade-Harleyy:
This PR adds a rotation & scaling tool to the submarine editor.
Its functions can be toggled with the tickboxes to the right of the "Generate Waypoints" button.

With "Edit Rotation" active, users can drag the widget around to rotate the selected entities about the center of the selection.
With "Edit Scaling" active, users can drag the widget around to scale the selected entities about the center of the selection.

The current rotation and/or scaling offset is visible in a tooltip next to the widget while it is being dragged.

Added content is in the mod provided with the PR.
The tickboxes also require their own art to be done, hence the duplicate styles I have added.

image

### DIFF
--- a/Barotrauma/BarotraumaClient/ClientSource/GUI/Widget.cs
+++ b/Barotrauma/BarotraumaClient/ClientSource/GUI/Widget.cs
@@ -112,8 +112,7 @@ namespace Barotrauma
         {
             PreUpdate?.Invoke(deltaTime);
             if (!enabled) { return; }
-            if (IsMouseOver) { Hovered?.Invoke(); }
-            if (PlayerInput.PrimaryMouseButtonHeld() && (IsMouseOver || !RequireMouseOn && SelectedWidgets.Contains(this)))
+            if (IsMouseOver || (!RequireMouseOn && SelectedWidgets.Contains(this) && PlayerInput.PrimaryMouseButtonHeld()))
             {
                 if (RequireMouseOn || PlayerInput.PrimaryMouseButtonDown())
                 {
--- a/Barotrauma/BarotraumaClient/ClientSource/GUI/Widget.cs
+++ b/Barotrauma/BarotraumaClient/ClientSource/GUI/Widget.cs
@@ -112,9 +112,9 @@ namespace Barotrauma
         {
             PreUpdate?.Invoke(deltaTime);
             if (!enabled) { return; }
+            if (IsMouseOver) { Hovered?.Invoke(); }
             if (PlayerInput.PrimaryMouseButtonHeld() && (IsMouseOver || !RequireMouseOn && SelectedWidgets.Contains(this)))
             {
-                Hovered?.Invoke();
                 if (RequireMouseOn || PlayerInput.PrimaryMouseButtonDown())
                 {
                     if ((multiselect && !SelectedWidgets.Contains(this)) || SelectedWidgets.None())
--- a/Barotrauma/BarotraumaClient/ClientSource/GUI/Widget.cs
+++ b/Barotrauma/BarotraumaClient/ClientSource/GUI/Widget.cs
@@ -112,7 +112,7 @@ namespace Barotrauma
         {
             PreUpdate?.Invoke(deltaTime);
             if (!enabled) { return; }
-            if (IsMouseOver || (!RequireMouseOn && SelectedWidgets.Contains(this) && PlayerInput.PrimaryMouseButtonHeld()))
+            if (PlayerInput.PrimaryMouseButtonHeld() && (IsMouseOver || !RequireMouseOn && SelectedWidgets.Contains(this)))
             {
                 Hovered?.Invoke();
                 if (RequireMouseOn || PlayerInput.PrimaryMouseButtonDown())
--- a/Barotrauma/BarotraumaClient/ClientSource/Screens/SubEditorScreen.cs
+++ b/Barotrauma/BarotraumaClient/ClientSource/Screens/SubEditorScreen.cs
@@ -8,6 +8,7 @@ using Microsoft.Xna.Framework.Input;
 using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
+using System.Globalization;
 using System.Linq;
 using System.Threading;
 using System.Xml.Linq;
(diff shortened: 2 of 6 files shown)


Barotrauma/BarotraumaClient/ClientSource/GUI/Widget.cs
```diff
@@ -112,7 +112,7 @@ public virtual void Update(float deltaTime)
         {
             PreUpdate?.Invoke(deltaTime);
             if (!enabled) { return; }
-            if (IsMouseOver || (!RequireMouseOn && SelectedWidgets.Contains(this) && PlayerInput.PrimaryMouseButtonHeld()))
+            if (PlayerInput.PrimaryMouseButtonHeld() && (IsMouseOver || !RequireMouseOn && SelectedWidgets.Contains(this)))
```
Regalis11, line 115: I don't understand this change. Doesn't this mean that the `Hovered` callback won't be called unless the mouse button is held, pretty much defeating the purpose of that callback?

Barotrauma/BarotraumaClient/ClientSource/Screens/SubEditorScreen.cs
```diff
@@ -36,6 +39,131 @@ public enum Mode
...
             Wiring
         }
+        
+        private GUITickBox RotateToolToggle, ScaleToolToggle;
+        
+        private static Vector2 SelectionCenter
+        {
+            get
+            {
+                List<MapEntity> nonWireEntities = MapEntity.FilteredSelectedList.Where(entity => (entity as Item)?.GetComponent<Wire>() == null).ToList();
```
Regalis11, line 49: To me it seems unnecessary to instantiate a new list here, could just use the `IEnumerable` returned by `Where`. It seems like this can be called every frame, which means instantiating a new list every time would lead to a considerable performance hit.

Barotrauma/BarotraumaClient/ClientSource/Screens/SubEditorScreen.cs
```diff
@@ -36,6 +39,131 @@ public enum Mode
...
+                                if (entity.ResizeHorizontal) { entity.RectWidth = (int)(oldData.rect.Width * scaleMult); }
+                                if (entity is Structure structure && oldData.texOffset.HasValue) { structure.TextureOffset = oldData.texOffset.Value * scaleMult; }
+                            }
+                            oldData.textScales?.ForEach(pair => pair.Key.TextScale = pair.Value * scaleMult);
+                            oldData.lightRanges?.ForEach(pair => pair.Key.Range = pair.Value * scaleMult);
+                            oldData.wires?.ForEach(pair => pair.Key.Width = pair.Value.width * scaleMult);
+                        }
+
+                        entity.Move(MathUtils.RotatePointAroundTarget(oldSelectionCenter + (oldData.pos - oldSelectionCenter) * scaleMult, oldSelectionCenter, -rotationAngleRad) - entity.DrawPosition);
+                        oldData.wires?.ForEach(pair => pair.Key.SetNodes(pair.Value.nodes.Select(nodePos => MathUtils.RotatePointAroundTarget(selectionWirePos + (nodePos - selectionWirePos) * scaleMult, selectionWirePos, -rotationAngleRad)).ToList()));
```
Regalis11, line 161: Instantiating a new list here seems unnecessary and seems like it could have some negative performance implications.

Regalis11 (changes requested):
Thank you for the contribution! This seems like a really handy addition. The code changes look mostly good, but I spotted a couple of things that could have some negative performance implications and would most likely be easy to optimize.

Barotrauma/BarotraumaClient/ClientSource/GUI/Widget.cs
Jade-Harleyy: I changed this originally due to the widget following the cursor even when released due to it not checking if the mouse was being held.

However, this did change the Hovered behaviour, so i've added an additional check for that.

Barotrauma/BarotraumaClient/ClientSource/Screens/SubEditorScreen.cs
Jade-Harleyy: Fixed by letting Wire.SetNodes() take any enumerable.

Barotrauma/BarotraumaClient/ClientSource/GUI/Widget.cs
```diff
@@ -112,9 +112,9 @@ public virtual void Update(float deltaTime)
         {
             PreUpdate?.Invoke(deltaTime);
             if (!enabled) { return; }
-            if (IsMouseOver || (!RequireMouseOn && SelectedWidgets.Contains(this) && PlayerInput.PrimaryMouseButtonHeld()))
+            if (IsMouseOver) { Hovered?.Invoke(); }
+            if (PlayerInput.PrimaryMouseButtonHeld() && (IsMouseOver || !RequireMouseOn && SelectedWidgets.Contains(this)))
```
Regalis11, line 116: This still changes the behavior of the method. Previously Hovered was called if the cursor is on the widget, or if it's being dragged and doesn't require the mouse to be on it to be draggable.

If you want the new widgets to behave in some other way, I think that should be handled in the callbacks or by disabling `RequireMouseOn`.

Regalis11 (changes requested):
I think there's still an issue with how the behavior of the `Hovered` callback was modified.

SomeRandomNoobKekeke:
1. If you press esc while the widget is grabbed you won't be able to release it, and if you quit to main menu and enter sub editor again the widget will become ungrabable. Because it's still in Widget.SelectedWidgets? God knows how baro gui works
So i think you should somehow clear SelectedWidgets on quiting

2. I think it should react to moving camera with wasd / mouse scroll while grabbed. Barotrauma widgets do, so i think it's possible

3. Widget should appear at the angle of rotation of the item and initial scale should be set to the scale of the item. I understand that you work with selections, but make an exception if only one item is selected, this is deceiving, it shows you that item always have 0 rotation and 1 scale whenever you grab it

4. Can numbers it the widget have fixed level of precision?

5.  When placing walls by dragging why do you snap them to grid? Placing and rotating them manually doesn't have this limitation

6. When you end placing wall by dragging, the wall jumps dramatically ruining your precise positioning

Regalis11:
This has been merged in our private repo now, closing

PR #15484, opened 2025-01-07
Mild XML Cleanup

711885:
There are many useless spaces that have been bothering me for quite a while and I took care of most, although there is a lot more cleaning up to do in regards of specifications such as deleting attributes whose value is already the serialized default, I might do that later.
There are no cons to this and only 2 pros: it's making me feel inner peace and it saves you around 4 KBs
## What I have done:
```
"<?xml version="1.0" encoding="utf-8" ?>" to "<?xml version="1.0" encoding="utf-8"?>",
" />" to "/>",
" >" to ">",
">                       " to ">"
```


Regalis11:

> There are no cons to this

I disagree with this, in my opinion there are several cons to this:
- This PR is an enormous amount of work to review - over 95k lines, and since there's just one commit, we can't see which of the lines have been actually modified.
- We cannot be certain that the changes are just formatting changes without reviewing the whole thing. There may be accidental (or intentional) functional changes somewhere in here too.
- This makes it a lot more inconvenient to browse the project history - if we want to check why some line in the XML has been done the way it is with `git blame`, it would in many cases just show this formatting change.

As such, I don't think we can accept this PR - I don't think it's reasonable for us to start reviewing such a massive PR and risk breaking things just for these minor formatting improvements.

711885:
Isn't it a bit early to close it? I could try to workout the issues by dividing it into multiple pull requests (XML Cleaning of Unfolded Files, XML Cleaning of BackgroundCreature Folder) and etc with almost all other folders, with committing first the unmodified files and then the changed ones, showing the only change was the formatting cleanup, that way there is no risk
I could also limit myself to ONLY do another PR AFTER the previous one was reviewed and accepted

PR #15497, opened 2025-01-08
Oxygen Generator Improvements

Jade-Harleyy:
This PR makes changes to oxygen generators that I believe will enhance the interactions available with them.

The first change adds the ability for players to find and change the oxygen generation rate either manually, or through wiring.

The second change adds a UI to oxygen generators that allows players to:
1. See if it is powered.
2. Manually change the oxygen generation rate.
3. See if the generation rate is being controlled through wiring.

The third change adds a new upgrade for oxygen generators that increases the oxygen generation rate.
This upgrade has a base cost of 2000 mk and 5 oxygenite shards, with tier restrictions of:
- Tier 1: Unavailable
- Tier 2: Level 1
- Tier 3: Level 2

Each level of this upgrade increases oxygen generation by 20%.

The intended effect of these changes is to add more depth to the mechanics of oxygen generators, with the end goal of getting players to engage with them beyond it just being a repairable device that fills oxygen tanks.

For example, traitors can set the generation amount to a lower value to silently starve the crew of oxygen.
Another example is players can create more complex "duty cycle" circuts to save fuel by dynamically changing the generation amount through wiring, therefore reducing the power consumption.

### DIFF
--- a/Barotrauma/BarotraumaShared/LocalMods/oxygen-changes/Upgrades.xml
+++ b/Barotrauma/BarotraumaShared/LocalMods/oxygen-changes/Upgrades.xml
@@ -5,7 +5,7 @@
     <ResourceCost item="oxygeniteshard" amount="5" levels="1-2"/>
     <MaxLevel tier="1" level="0"/>
     <MaxLevel tier="2" level="1"/>
-    <OxygenGenerator generatedamount="+20%"/>
+    <OxygenGenerator generatedamount="+20%" ratewarningindicatorlow="-20%" ratewarningindicatorextremelylow="-20%"/>
     <Sprite texture="Content/UI/CampaignUIAtlas2.png" sourcerect="768,768,128,128"/>
   </UpgradeModule>
 </UpgradeModules>
--- a/Barotrauma/BarotraumaShared/LocalMods/oxygen-changes/Upgrades.xml
+++ b/Barotrauma/BarotraumaShared/LocalMods/oxygen-changes/Upgrades.xml
@@ -1,11 +1,11 @@
 <UpgradeModules>
   <UpgradeCategory identifier="oxygengenerators" items="oxygengenerator"/>
-  <UpgradeModule identifier="increaseoxygengeneration" maxlevel="2" categories="oxygengenerators" increaseontooltip="25">
+  <UpgradeModule identifier="increaseoxygengeneration" maxlevel="2" categories="oxygengenerators" increaseontooltip="20">
     <Price baseprice="2000" increaselow="+0%" increasehigh="+200%"/>
     <ResourceCost item="oxygeniteshard" amount="5" levels="1-2"/>
     <MaxLevel tier="1" level="0"/>
     <MaxLevel tier="2" level="1"/>
-    <OxygenGenerator generatedamount="+25%" powerconsumption="-5%"/>
+    <OxygenGenerator generatedamount="+20%"/>
     <Sprite texture="Content/UI/CampaignUIAtlas2.png" sourcerect="768,768,128,128"/>
   </UpgradeModule>
 </UpgradeModules>
--- a/Barotrauma/BarotraumaShared/LocalMods/oxygen-changes/Upgrades.xml
+++ b/Barotrauma/BarotraumaShared/LocalMods/oxygen-changes/Upgrades.xml
@@ -5,7 +5,7 @@
     <ResourceCost item="oxygeniteshard" amount="5" levels="1-2"/>
     <MaxLevel tier="1" level="0"/>
     <MaxLevel tier="2" level="1"/>
-    <OxygenGenerator generatedamount="+20%" ratewarningindicatorlow="-20%" ratewarningindicatorextremelylow="-20%"/>
+    <OxygenGenerator generatedamount="+20%"/>
     <Sprite texture="Content/UI/CampaignUIAtlas2.png" sourcerect="768,768,128,128"/>
   </UpgradeModule>
 </UpgradeModules>
--- a/Barotrauma/BarotraumaShared/LocalMods/oxygen-changes/Upgrades.xml
+++ b/Barotrauma/BarotraumaShared/LocalMods/oxygen-changes/Upgrades.xml
@@ -1,0 +1,11 @@
+<UpgradeModules>
+  <UpgradeCategory identifier="oxygengenerators" items="oxygengenerator"/>
+  <UpgradeModule identifier="increaseoxygengeneration" maxlevel="2" categories="oxygengenerators" increaseontooltip="25">
+    <Price baseprice="2000" increaselow="+0%" increasehigh="+200%"/>
+    <ResourceCost item="oxygeniteshard" amount="5" levels="1-2"/>
+    <MaxLevel tier="1" level="0"/>
+    <MaxLevel tier="2" level="1"/>
+    <OxygenGenerator generatedamount="+25%" powerconsumption="-5%"/>
+    <Sprite texture="Content/UI/CampaignUIAtlas2.png" sourcerect="768,768,128,128"/>
+  </UpgradeModule>
+</UpgradeModules>
--- a/Barotrauma/BarotraumaShared/LocalMods/oxygen-changes/filelist.xml
+++ b/Barotrauma/BarotraumaShared/LocalMods/oxygen-changes/filelist.xml
@@ -1,0 +1,5 @@
+<ContentPackage name="oxygen-changes" modversion="1.0.0" gameversion="1.0.8.0">
+  <Text file="%ModDir%/Locale/English.xml"/>
+  <Item file="%ModDir%/OxygenGenerators.xml"/>
+  <UpgradeModules file="%ModDir%/Upgrades.xml"/>
+</ContentPackage>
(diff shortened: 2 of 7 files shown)


Barotrauma/BarotraumaShared/LocalMods/oxygen-changes/Upgrades.xml
```diff
@@ -0,0 +1,11 @@
+<UpgradeModules>
+  <UpgradeCategory identifier="oxygengenerators" items="oxygengenerator"/>
```
JeroonBosch, line 2: I believe these have no text description defined right now, unless I am missing something?

JeroonBosch (commented):
Missing text for oxygengenerator upgrade?

Barotrauma/BarotraumaShared/LocalMods/oxygen-changes/Upgrades.xml
Jade-Harleyy: I discovered unused localization tags in EnglishVanilla.xml, so it should be using those.

itchyOwl (changes requested):
Overall looks pretty good, spotted only some minor code quality issues. Also would have to change the content definitions a bit, if we'd include the changes in the vanilla game.

Only skimmed through the xml changes, didn't go the UI code very thoroughly, as it's easier to spot the issues by testing, which I didn't do.

No objections on these changes overall.

Barotrauma/BarotraumaShared/LocalMods/oxygen-changes/Upgrades.xml
JeroonBosch: Ah alright, yeah I didn't check for those
> Increase the amount of oxygen generated.
Maybe that should still be expanded to also mention the power consumption (which may be more relevant in most cases). But we could also take care of that.

JeroonBosch:
> For example, traitors can set the generation amount to a lower value to silently starve the crew of oxygen.

Without opening the UI of the oxygen generator, would this be visible in any way? At least, to me it would make sense if vents stopped turning when oxygen generation is at 0 (and slowing down at lower values, ideally), but also the generator itself no longer moving and/or showing a red light or orange light based on the percentage of oxygen generation.

Barotrauma/BarotraumaShared/LocalMods/oxygen-changes/filelist.xml
```diff
@@ -0,0 +1,5 @@
+<ContentPackage name="oxygen-changes" modversion="1.0.0" gameversion="1.0.8.0">
```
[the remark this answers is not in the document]
Jade-Harleyy: Hey there, I'm not sure where I should place these files after addressing this, or how I should handle differences between changes/additions and mixed types.
Please let me know!

Jade-Harleyy:
> Without opening the UI of the oxygen generator, would this be visible in any way? At least, to me it would make sense if vents stopped turning when oxygen generation is at 0 (and slowing down at lower values, ideally), but also the generator itself no longer moving and/or showing a red light or orange light based on the percentage of oxygen generation. 

Vents already stop turning when no oxygen is being produced, and both the vents and the generator make no sound when not producing anything.
I've also went ahead and stopped the generators from animating, though.

Barotrauma/BarotraumaShared/LocalMods/oxygen-changes/Upgrades.xml
Jade-Harleyy: Maybe that should still be expanded to also mention the power consumption (which may be more relevant in most cases).

I've decided to remove the reduced power consumption for the reason provided in commit fdb553, so this is no longer neccessary.

itchyOwl (approved):
Looks good to me from the technical perspective. Would have to decide what to do with this. As commented earlier, some changes are still required in the content definitions, if we'd include this in the vanilla game. But we can take care of that. For testing purposes, it's fine to have this implemented as a mod for now. GJ!

Regalis11:
One concern I have about this is whether it's clear enough what's happening if someone for example turns the oxygen generation rate to 10%. How would a player know it's possible to adjust it, where is it communicated? How can a player tell the oxygen generation has been adjusted down, would it seem like a bug if they start suffocating for seemingly no reason?

mygamingaccount:
How is the sub supposed to handle grid overloads upon thrust deduction if the O2 genny is not there to suck up the extra power? Or is overvoltage not a thing any more in upcoming versions? (I'm cool with that by the way)

This change is very similar to [a circuit that I've designed](https://steamcommunity.com/sharedfiles/filedetails/?id=3104296897) for a campaign which involved having to manage fuel, but on that same sub I had to [entirely delete any overvoltage action](https://steamcommunity.com/sharedfiles/filedetails/?id=3104164209) just to make that kind of gameplay possible.

ubertpendragon:
> I think one thing that would alleviate the problem a bit is if the oxygen generation rate was visible on the generator and the vents somehow. Maybe the animation speed (and the volume of the sounds too?) could be relative to the oxygen generation rate?

I think that defeats the purpose of "silently" sabotaging the oxygen as Jade suggested

The use cases for O2 generator changes are:
-Starving fires (you previously had to disconnect the O2 power wire)
-Discreet Sabotage (instead of completely shutting off O2)
-Saving power

> The third change adds a new upgrade for oxygen generators that increases the oxygen generation rate.

This just reads as "reduce O2 power cost", subs don't really need more O2 generation since they're made with giving enough O2 to all rooms to begin with (it's not FTL where you can overpower a breach leaking oxygen with upgraded O2).
I believe O2 upgrades were removed because they were largely useless


Another problem I can think of with Saving Power is that you can survive on O2 shelves filling bottles. O2 shelves don't need the O2 generator to function since they're not LINKED but rather powered (sub builders wish they had been from the very start)

Regalis11:
> I think that defeats the purpose of "silently" sabotaging the oxygen as Jade suggested

Possibly to some extent, but I think it's still important for there to be some kind of a cue that someone has sabotaged the oxygen. Maybe not something immediately obvious, but I feel like if we just add this sort of a "hidden feature" without communicating it anywhere in any way, it will just seem like a bug when the sub stops running out of oxygen for seemingly no reason or any clear way to fix it.

JoneKone:
> One concern I have about this is whether it's clear enough what's happening if someone for example turns the oxygen generation rate to 10%. How would a player know it's possible to adjust it, where is it communicated? How can a player tell the oxygen generation has been adjusted down, would it seem like a bug if they start suffocating for seemingly no reason?
> 
> I think one thing that would alleviate the problem a bit is if the oxygen generation rate was visible on the generator and the vents somehow. Maybe the animation speed (and the volume of the sounds too?) could be relative to the oxygen generation rate?

Or just vents are rotating slower than normal. Or both.

jigardi:
> > I think that defeats the purpose of "silently" sabotaging the oxygen as Jade suggested
> 
> Possibly to some extent, but I think it's still important for there to be some kind of a cue that someone has sabotaged the oxygen. Maybe not something immediately obvious, but I feel like if we just add this sort of a "hidden feature" without communicating it anywhere in any way, it will just seem like a bug when the sub stops running out of oxygen for seemingly no reason or any clear way to fix it.

https://en.wikipedia.org/wiki/Generalized_hypoxia
^This link suggests there are at least 18 discernible symptoms for hypoxia,  some of which are readily available to apply via the affliction system.

As a player/submarine builder,  I'd like those symptoms to gradually kick in at sustained low/decreasing oxygen levels.  Some of the crew will become sick with minor symptoms at first(hidden in the UI until some critical threshold),  and if they do not catch on or fail to communicate with their team;  that's on them - just like it is IRL.

IRL, low-oxygen is so dangerous and feared because it is often not detected in time;  and worse still sometimes you detect the symptoms,  but they fail to alarm you...

- Impaired judgment, confusion, memory loss and cognitive problems.
- Euphoria or dissociation

...and then you die horribly. xD

If you do your best to implement some reasonable aspects of those 18 symptoms,  and then put a big old disclaimer in the update text;  then I think the majority of people will be fine with it.
Or even better, put a new red O2 disclaimer in the loading tips!  Then people who fail to read updates will also get exposed to the changes over time.

> You could make oxygen generation matter a little more if flooding a hull also drained its oxygen, or making plants consume oxygen.

^I strongly second this notion.  But would probably make plants yield O2 and consume CO2 - at least the tobacco plant since it is presumably from Earth.

As it is now you can catastrophically flood the boat - complete-hull-breach, open-airlock,  doesn't matter it's all the same - the oxygen magically remains in the room!
   That is a massive error,  since it removes the primary need for the O2 generator in the first place(Imagine if the same was true for pumps and water;  the game would hardly make any sense!).

This also has the knock on effect of making oxygen detectors virtually meaningless.  The game has oxygen detectors sure,  but so far we haven't had any reason to really care about them(except maybe for wiring).

I can count on one hand,  how often people have asked about the oxygen levels in the status monitor - and all that has ever lead to has been:  _"Understood!  I'll fix the O2 gen if it breaks down."_ - and then that conversation ends for ever. xD

If you end up addressing this issue,  I think you should also consider revamping fire and vents.  Allowing fire to spread through vents would make fire a real danger.  As it is now,  fire is almost always contained to its starting hull,  unless a player lets it out.

Just my two cents as a humble player/submarine builder of some 2,5K hours.

PS:  Thank you for almost making my dream game <3

Edit:  Some of the quotes and text collapsed.  Should be properly separated.

PR #15514, opened 2025-01-11
Modular XML Cleanup 1 - BackgroundCreaturePrefabs, ContentPackages & Particles

711885:
Taking Regalis' concerns into account I've decided to split my pull requests into a few smaller ones for easier progressive review, and also made sure there were commits showing the changes I have made. I'll only work on the next pull request once the previous (in this case this current one) has been reviewed and merged.
In BackgroundCreaturePrefabs.xml I deleted attributes that are already the serialized default values.
```
swarmmin="1"
swarmmax="1"
```
Shown in https://github.com/FakeFishGames/Barotrauma/blob/master/Barotrauma/BarotraumaClient/ClientSource/Map/Levels/BackgroundCreatures/BackgroundCreaturePrefab.cs


711885:
I also want to note that needless specific values such as `1.0` & `20.0` can be _optimized_ into `1` & `20`
The `.0` in floating values is useless as far as I have tested, although I haven't made that optimization since I'm not sure you'd want it.
If you wish for it, you can request that change and all future pull requests will feature that optimization from me
Also for `0.20` into `0.2`

Regalis11:
I still feel iffy about these sort of "cleanup" PRs. Even though this one is a lot smaller and more "reviewable" than [the previous one](https://github.com/FakeFishGames/Barotrauma/pull/15484), I still feel doing these sorts of changes comes with several drawbacks, and the minor benefit of saving some bytes doesn't really make much difference in practice.

- This makes it a lot more inconvenient to browse the project history - if we want to check why some line in the XML has been done the way it is with git blame, it would in many cases just show this formatting change.
- I feel it's actually useful to have some of the default values in the XML files as well: it makes it clear to modders these attributes exist.
- I'd argue it's useful to format 1.0 with the decimal instead of converting it to 0: the latter makes it clear it's a float value and not an integer.

Regalis11:
> I mean, on the second point that's one of the reasons why the https://regalis11.github.io/BaroModDoc/ exists, it has almost all of the attributes, and the same attributes could be found somewhere else except with their default value.

That's a fair point, but think it can also be convenient to have somewhat "complete" examples of what can be edited in the vanilla files: I would assume that many modders copy vanilla content and modify them to suit their needs, and don't necessarily always check the documentation. I know I do at least. :D

And yeah, there are definitely other ways to browse the project history, it's just very handy to be able to look at any arbitrary line and easily see when and why it was last modified.

PR #15521, opened 2025-01-13
overriding with inheritance from overridden

EugenPolyakov:
Allow overriding with inheritance from overridden.
Know issues:
- the new element is always inherited from the last override (this is normal behavior now)
- not tested with <clear> and topMostOverrideFile because it is not clear when and for what it is used


Regalis11:
Thank you for the PR! It's worth noting that there's [another PR](https://github.com/FakeFishGames/Barotrauma/pull/9779) that implements this, although it has been in the works since 2022 and is still a work in progress. 

This seems like a vastly simpler implementation though - which makes me wonder if there are some edge cases that aren't taken into account here that would need more complex logic to solve. I don't have time to do a proper review on this right now, but I'll look into it in the future and try to think of potential edge cases or issues in the chosen approach though.

PR #15530, opened 2025-01-16
Pass all args to GetValidArgs in DebugConsole.AutoComplete

SomeRandomNoobKekeke:
> [!NOTE]
> Just a draft

It's impossible to organize args like a tree, all the args from different branches on the same depth will mix in one array
On the other hand you're getting string[][] from GetValidArgs and using only one layer

Example: i have a test command that should run some test method from some test class like that "test testClass testMethod"
testClass will autocomplete fine, but all testMethods from all classes will mix in one array because they all considered 2nd arg

It would probably require changing all the commands, but since you are already considering this i can't help but ask what do you think of this?
Screenshot_3

### DIFF
--- a/Barotrauma/BarotraumaShared/SharedSource/DebugConsole.cs
+++ b/Barotrauma/BarotraumaShared/SharedSource/DebugConsole.cs
@@ -2314,7 +2314,7 @@ namespace Barotrauma
                 int autoCompletedArgIndex = args.Length > 0 && command.Last() != ' ' ? args.Length - 1 : args.Length;
 
                 //get all valid arguments for the given command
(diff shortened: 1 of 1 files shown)


SomeRandomNoobKekeke:
if think it's possible to make some adapter
```C#
public class Command
    {
      public Func<string[], string[]> GetArgsForPosition;

      public Command(Func<string[][]> getValidArgs)
      {
        GetArgsForPosition = (string[] args) => getValidArgs().ElementAtOrDefault(args.Length);
      }

      public Command(Func<string[], string[]> getArgsForPosition)
      {
        GetArgsForPosition = getArgsForPosition;
      }
 }
```

PR #15698, opened 2025-03-02
Show circuit box labels in connection panel

MatthiasKunnen:
Previously, a circuit box's connection panel would show `SIGNAL_IN/OUT_1-8` instead of the labels that the user applied.

This fixes #15697.

### Example circuit box
Circuitbox example

### Previous
Previous connection panel

### Now
This PR makes it so the displayed connection's name is based on the circuitbox's label:
Current connection panel

Tested on both client and server.

I thought about doing this without an extra property but found no good way to restore the `DisplayName`.

### DIFF
--- a/Barotrauma/BarotraumaShared/SharedSource/Items/Components/Signal/Connection.cs
+++ b/Barotrauma/BarotraumaShared/SharedSource/Items/Components/Signal/Connection.cs
@@ -18,7 +18,15 @@ namespace Barotrauma.Items.Components
         public readonly int MaxWires = 5;
 
         public readonly string Name;
-        public readonly LocalizedString DisplayName;
+
+        private readonly LocalizedString _displayName;
+        public LocalizedString DisplayName
+        {
+	        get => DisplayNameOverride ?? _displayName;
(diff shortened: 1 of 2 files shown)


Barotrauma/BarotraumaShared/SharedSource/Items/Components/Signal/Connection.cs
```diff
@@ -18,7 +18,15 @@ partial class Connection
...
-        public readonly LocalizedString DisplayName;
+
+        private readonly LocalizedString _displayName;
+        public LocalizedString DisplayName
+        {
+	        get => DisplayNameOverride ?? _displayName;
+	        private init => _displayName = value;
+        }
+
+        public LocalizedString DisplayNameOverride;
```
MatthiasKunnen, line 29:
```suggestion
        public LocalizedString? DisplayNameOverride;
```
I wanted to make it nullable but the nullable context is not enabled here.

Regalis11:
Thank you for the PR! I've moved this to our private repo now, we'll test it internally and include it in a future update. :)

PR #15714, opened 2025-03-10
Display ballast ratio in sub editor

MatthiasKunnen:
The ballast ratio is the total hull volume divided by volume of hulls containing the word "ballast". This is the same approach as the "Wet room" detection.
The purpose is ease-of-use when building submarines and trying to figure out ballast tank sizes.

## Current workflow:
1. Change hull size.
2. Select all ballast tanks manually, this can be especially painful if a lot of linked hulls are present and the ballast tanks are scattered around the sub.
3. Divide total volume by selected volume using an external calculator. Now you know if the ballast-to-total volume has changed as you intended.
4. Tinker with hull size again, repeat these steps.

## New workflow
1. Change hull size.
2. Instantly see the current ballast ratio.
3. Happily fiddle with hull size enjoying immediate feedback.

image

## Note
I'm currently reusing the ballast translation from the roomname which is usually an i18n no-no. Since translations are not part of this repo, this will have to be addressed upstream.

## Potential future PR
I would like to propose making the _optimal neutral ballast level_ calculation automatic too in another PR. It might look like this:

```
Total hull volume:
7024156

Ballast:
1351447 (19.24 % of total volume)
optimal neutral ballast level is 0.3638

Selected volume:
512345 (optimal neutral ballast level is 0.9597)
```

Any thoughts?

### DIFF
--- a/Barotrauma/BarotraumaClient/ClientSource/Screens/SubEditorScreen.cs
+++ b/Barotrauma/BarotraumaClient/ClientSource/Screens/SubEditorScreen.cs
@@ -230,7 +230,22 @@ namespace Barotrauma
 
         private static LocalizedString GetTotalHullVolume()
         {
-            return $"{TextManager.Get("TotalHullVolume")}:\n{Hull.HullList.Sum(h => h.Volume)}";
+            float totalVolume = 0.0f;
+            float ballastVolume = 0.0f;
+            Hull.HullList.ForEach(h =>
+            {
+                totalVolume += h.Volume;
+
+                if (h.IsBallast())
+                {
+                    ballastVolume += h.Volume;
+                }
+            });
+
+            string retVal = $"{TextManager.Get("TotalHullVolume")}:\n{totalVolume}";
+            retVal += $" ({(ballastVolume / totalVolume * 100).ToString("0.00")} % {TextManager.Get("roomname.Ballast")})";
+
+            return retVal;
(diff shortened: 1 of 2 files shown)


Barotrauma/BarotraumaClient/ClientSource/Screens/SubEditorScreen.cs
```diff
@@ -230,7 +230,22 @@ private static string GetSubDescription()
...
+                totalVolume += h.Volume;
+
+                if (h.IsBallast())
+                {
+                    ballastVolume += h.Volume;
+                }
+            });
+
+            string retVal = $"{TextManager.Get("TotalHullVolume")}:\n{totalVolume}";
+            retVal += $" ({(ballastVolume / totalVolume * 100).ToString("0.00")} % {TextManager.Get("roomname.Ballast")})";
```
deeyonn, line 246: I noticed that if `totalVolume` is `0.0f`, the division here could result in `Infinity`, which might cause unexpected behavior when formatting the string.

MatthiasKunnen: I had the same thought when writing this, though `GetTotalHullVolume` is only called when at least one hull is selected so this should never happen.
That being said, I have no problem adding a defensive guard just in case.

PR #15774, opened 2025-03-26
Add minValue for CharacterAbilityPsychoClown.cs

NoahRoopull:
Added a minValue, and the strength is lerped between the two in order to avoid needing another statValue change elsewhere. Shouldn't conflict with existing implementations, given that the base was by default 0. This doesn't fix/change anything in vanilla, but it gives a little bit more flexibility in modding.


Regalis11:
Thank you for the PR! Merged this to our private repo now, fix coming up in a future update.

PR #15827, opened 2025-04-06
Improved Reactor modding compatibility

ubertpendragon:
Reactors have hard restrictions on items they can hold, and the fuel value of the rods in defined in reactors, not the rods themselves. #15826

430701713-fd521ebf-52ba-4a5f-a383-e541c0eb9c8d

This PR simply puts the fuel value in the rods and allows reactor to hold any time of item tagged as "reactorfuel" or "reactoritem", which allows modders to add new fuel/reactor items without overriding the reactor and causing conflicts. 

**Will this affect performance?**
Probably, I'm not sure to what extent: instead of having 5 statuseffects per reactor you have 1 per fuel rod. I spawned 100 fuel rods in a cabinet and it didnt kill the framerate. Adding an interval breaks the fuel rod. The average campaign will have 20 fuel rods at most, but I'll need more accurate profiling to know.

**Won't this break existing subs?**
I've tested sub editor and loaded existing campaigns, it seems just fine.
I'm not familiar with <Update /> at all, but the reactors and rods seem to have updated just fine, and the changes are quite simple as you can see

**Won't this break existing mods?**
Maybe, but most mods override reactors and fuel anyway partly due to this exact issue, and I'm in touch with modders who do reactor stuff, like M with Enhanced Reactors and Foxtrot with Hazardous Reactors which are the two big ones. I'll try notifying other modders as well.

### DIFF
--- a/Reactorfix/OutpostItems.xml
+++ b/Reactorfix/OutpostItems.xml
@@ -1,0 +1,1246 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Items>
+  <Item name="" identifier="op_clowncurtain2" width="199" height="381" scale="0.5" noninteractable="true" category="Decorative" subcategory="clownassets">
+    <sprite texture="Content/Map/Outposts/Art/ClownDistrict.png" sourcerect="2,2,199,381" depth="0.97" premultiplyalpha="false" origin="0.5,0.5" />
+  </Item>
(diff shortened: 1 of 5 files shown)


Regalis11:
As we discussed in Discord, I feel like this is a potentially risky way to do this, as it would break potentially a lot of mods that modify reactors and fuel rods. I decided to instead do this by adding the generic "reactorfuel" tag as a containable for the reactor (commit in our private repo https://github.com/FakeFishGames/Barotrauma-development/commit/883101ed43c506e24ffdd1b4a43028ea026f7d71).

PR #15832, opened 2025-04-07
Fix: Character Portrait Attachment Alignment in UI

ToumanLin:
Actually, the problem in issue #14646 wasn't fixed in the 1.6 update. The problem even get worse.
Sample XML:
HumanDefaultRagdoll.xml (limb = "head")
```
    <sprite texture="%ModDir:2809175631%/Content/Characters/Human/Human_[GENDER]_heads.png" sourcerect="0,0,160,228" depth="0.05" origin="0.6125,0.33" color="255,255,255,255" deadcolor="255,255,255,255" deadcolortime="0" ignoretint="False" />
```
Human.xml
```
      <Wearable type="Hair" commonness="0.05" replacewhenwearinghat="16" >
        <sprite name="Hair 1" texture="%ModDir:2809175631%/Content/Characters/Human/Headattachments/Human_decor.png" sourcerect="0,0,240,480" origin="0.6125,0.3229" inheritorigin="false" inheritsourcerect="false" />
      </Wearable>
```
In the current version, the hair attachment appears incorrectly offset from the head.
Screenshot 2025-04-06 222024
Screenshot 2025-04-06 222212

This is because, the offset compensates used in `DrawHeadAttachmentSprite` in `CharacterInfo.cs` would be affect both `DrawPortrait` and `DrawIcon` Methods
The offset: `drawPos += HeadSprite.Origin * scale;`
Neither `DrawIcon` Method need this offset, nor this offset is correct for `DrawPortrait` Method.

In my fix, I removed this offset in `DrawHeadAttachmentSprite`, so it runs correctly for `DrawIcon`, 
I duplicate the `DrawHeadAttachmentSprite` and renamed it as `DrawPortraitHeadAttachmentSprite`, and use a new offset to make the sprite aligns.

With this simple fix, the UI now looks better
Screenshot 2025-04-06 212956
Screenshot 2025-04-06 215200

### DIFF
--- a/Barotrauma/BarotraumaClient/ClientSource/Characters/CharacterInfo.cs
+++ b/Barotrauma/BarotraumaClient/ClientSource/Characters/CharacterInfo.cs
@@ -410,7 +410,7 @@ namespace Barotrauma
                     foreach (var attachment in attachmentsToDraw)
                     {
                         SetAttachmentEffect(spriteBatch, attachment);
-                        DrawAttachmentSprite(spriteBatch, attachment, portraitToDraw, sheetIndex, screenPos + offset, scale, depthStep, GetAttachmentColor(attachment, hairColor, facialHairColor), flip ? SpriteEffects.FlipHorizontally : SpriteEffects.None);
+                        DrawProtraitAttachmentSprite(spriteBatch, attachment, portraitToDraw, sheetIndex, screenPos + offset, scale, depthStep, GetAttachmentColor(attachment, hairColor, facialHairColor), flip ? SpriteEffects.FlipHorizontally : SpriteEffects.None);
                         depthStep += depthStep;
                     }
                 }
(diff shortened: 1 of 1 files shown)


Regalis11 (changes requested):

I don't think this is a good solution to the issue though. In my opinion it doesn't make sense to have two nearly-identical methods with a ton of duplicate code. It's also unclear _why_ the behavior is or should be different depending on whether you're drawing an icon or a portrait, or in which kind of a context `DrawIcon` should be used over `DrawPortrait`, or vice versa.

I ended up implementing a different solution to this in our private repo, getting rid of the DrawPortrait method entirely and using DrawIcon everywhere. The fix will be included in a future unstable build.

ToumanLin:
> Thank you for the contribution!
> 
> I don't think this is a good solution to the issue though. In my opinion it doesn't make sense to have two nearly-identical methods with a ton of duplicate code. It's also unclear _why_ the behavior is or should be different depending on whether you're drawing an icon or a portrait, or in which kind of a context `DrawIcon` should be used over `DrawPortrait`, or vice versa.
> 
> I ended up implementing a different solution to this in our private repo, getting rid of the DrawPortrait method entirely and using DrawIcon everywhere. The fix will be included in a future unstable build.

Sorry, but do you sure it get fixed? I used the Update today and the problem still there

Regalis11:
The fix unfortunately didn't make it in to yesterday's update, we still need to review and test it internally before we can include it in the public version. Sorry for the inconvenience!

Regalis11:
A2ur3 also pointed the issue out in https://github.com/FakeFishGames/Barotrauma/discussions/15722. :/ 

> It seems the issue happens when the origin of the head sprite isn't at the center horizontally, and it doesn't get correctly flipped when the head is rendered facing left in the bottom-right corner. IIRC we only tested this with the XML provided in the ticket and a custom hair we made ourselves, which is why this issue slipped past us.

We've implemented a fix in our private repo, hopefully this is the last of these misalignment issues and the issue finally fixed for good. Sorry for the trouble!

Issue #11350, opened 2023-02-24
aitarget doesn't originate in same place as sonar


GoodDays13:
### Disclaimers

- [X] I have searched the issue tracker to check if the issue has already been reported.
- [ ] My issue happened while using mods.

### What happened?

The aitarget for the sonar comes out of the nav terminal but the sonar comes out of the center of the sub. This causes some jank with the sonar, for stuff like the dugong it gives a slight advantage seeing forward and orca 2 gives a slight disadvantage. It also generally messes with directional ping; here's an example:
![image](https://user-images.githubusercontent.com/42877449/221087217-16335b95-1878-468d-a464-474de57312b1.png)
![image](https://user-images.githubusercontent.com/42877449/221087226-4e0da62c-a6b5-43ac-8955-f4476ede81b9.png)
This could possibly be exploited by using directional sonar in a way that lets you see enemies but doesn't let them see you.
You can also make a sub with a nav terminal in the back and a transducer in the front that exploits this to see a lot further ahead without alerting enemies.
As a side note, logically, why would the sound come from the monitor inside the sub?

### Reproduction steps

1. start inside a sub in the sub editor
2. turn on `aitargets` and `debugai` from the console
3. turn on sonar
4. observe how the aitarget and sonar originate in different places

### Bug prevalence

Happens regularly

### Version

Faction/endgame test branch

### Which operating system did you encounter this bug on?

Windows

Regalis11:
Duplicate of https://github.com/Regalis11/Barotrauma/issues/10225

Issue #10225, opened 2022-11-03
Sonar origin doesn't match sound


GoodDays13:
### Disclaimers

- [X] I have searched the issue tracker to check if the issue has already been reported.
- [ ] My issue happened while using mods.

### What happened?

The sonar interface is centered on the middle of the sub and sends out pulses from the center of the sub, but the sounds for both player and AI seem to be centered on the navigation terminal.

### Reproduction steps

Start the game
Do `debugdraw` then `aitargets` in the console.
Turn on the sonar and see where the origin is.
![image](https://user-images.githubusercontent.com/42877449/199842078-0666e823-2aff-44e7-915b-f4cfc4ab2ce3.png)
Use `freecam` and see what the origin of the sound is.
![image](https://user-images.githubusercontent.com/42877449/199842281-6b1f5667-76fe-4f75-89da-bea9c6408bf5.png)

You can also see how this affects directional sonar.
Turn on the directional sonar and find something to set it in relation to
![image](https://user-images.githubusercontent.com/42877449/199840627-8d4c3086-248f-4b11-8a04-bfbe231cf1ee.png)
Use `freecam` to see what it actually looks like.
![image](https://user-images.githubusercontent.com/42877449/199840712-7610c410-5091-40a7-8ccc-69c18f98fd05.png)

### Bug prevalence

Happens every now and then

### Version

0.19.14.0

### Which operating system did you encounter this bug on?

Windows

Regalis11:
This was actually an intentional choice, we felt the nav terminal feels better to use when it's centered on the sub's center of mass instead of the source of the pings. If you want a more realistic representation, it's possible to connect the nav terminal to a Sonar Transducer and enable "CenterOnTransducers" in the submarine editor.

GoodDays13:
Why shouldn't the sound the ai sees be centered in the same place then so that the ping matches what the player sees? Also, the sound comes from the nav terminal even when connected to transducers, so the transducer has to be placed inside the nav terminal for them to match

GoodDays13:
I have made a sub using this exploit by putting the nav terminal in the back and a transducer in the front for the ability to see pretty far in front with active without alerting anything

PR #16075, opened 2025-05-15
Add AITarget to Submaries that is used with sonar machines

GoodDays13:
### What
Sonar machines use the AITarget provided by the Entity part of Submarines instead of the machine itself.
### Why
Fixes the misalignment between the display and AITarget that I talk about in #10225 and #11350
### Draft
I made this PR a draft to indicate that I've not looked too much into edge cases yet.


SomeRandomNoobKekeke:
2 years in development, still a draft :BaroModder:

PR #16084, opened 2025-05-17
Feat/add dns srv record resolution to client direct connection

Reidond:
Hello Barotrauma team 😇

In this PR I've added support for resolving DNS SRV records when client tries to join via Direct Join and enters domain name like `barotrauma.example.com`.

Why this is needed? Imagine that your dedicated server is hosting game server with different port than default one (`27015`) for example `42069` so users have to write this `barotrauma.example.com:42069` which is ugly.

DNS SRV solves this and allows to host multiple servers under same domain name.

Here are some context about it:

- https://www.cloudflare.com/learning/dns/dns-records/dns-srv-record/
- factorio implements DNS SRV too - https://wiki.factorio.com/Multiplayer#DNS_SRV_Records

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -56,5 +56,8 @@ Barotrauma/**/GameAnalyticsKeys.cs
 Deploy/DeployAll/PrivateKey.*
 .github/ISSUE_TEMPLATE/release-checklist.md
 
-#Rider
+# Rider
 *.DotSettings.user
+
+# VSCode
+.vscode/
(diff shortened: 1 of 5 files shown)


mygamingaccount:
> Imagine that your dedicated server is hosting game server with different port than default one

I'm having a hard time imagining that. People don't normally go out of their way to make simple things such as joining a multiplayer game harder by randomly changing network addresses.

Less than 10% of Barotrauma servers are dedicated, most just use Steamworks or whatever it is Epic uses, so this feature is for a niche of the niche.

Or is this something that you would specifically use? Do you have one IP address where you want to host 10 Barotrauma servers, and need the links to be pretty too?

Reidond:
@mygamingaccount 
I understand this issue is niece. I have one server for dedicated games I play and there is possibility that some port could be different from defaults ones that game might use. Originally I found this from Factorio game, and it seemed like cool feature to give my friends {game}.example.com instead of ip adress or domain.com:port

PR #16290, opened 2025-07-08
Remove OSX if else statement for save path.

Cubea01:
Fix for bug report  #16268. An if-else statement in the saveutil breaks the file pathing for macOS. Previously, save files were stored under `~/Library/Application Support/`, but after installing the latest update, they're now found under `~/Documents/Library/Application Support/`. I'm guessing the project was updated from .NET 7 to .NET 8 or higher in the latest release, which changed `Environment.SpecialFolder.Personal` from `$HOME` to `$HOME/Documents`. 

The obvious fix is to remove the if-else statement altogether and use `SpecialFolder.LocalApplicationData` for all operating systems.

https://learn.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/8.0/getfolderpath-unix

```
Console.WriteLine("Home: " + Environment.GetEnvironmentVariable("HOME"));
Console.WriteLine("Personal: " + Environment.GetFolderPath(Environment.SpecialFolder.Personal));
Console.WriteLine("ApplicationData: " + Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData));
Console.WriteLine("LocalApplicationData: " + Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData));
Console.WriteLine(".NET Version: " + Environment.Version);
Console.WriteLine("macOS Version: " + Environment.OSVersion);
```
```
Home: /Users/cubea01
Personal: /Users/cubea01/Documents
ApplicationData: /Users/cubea01/Library/Application Support
LocalApplicationData: /Users/cubea01/Library/Application Support
.NET Version: 9.0.6
macOS Version: Unix 15.5.0
```

### DIFF
--- a/Barotrauma/BarotraumaShared/SharedSource/Utils/SaveUtil.cs
+++ b/Barotrauma/BarotraumaShared/SharedSource/Utils/SaveUtil.cs
@@ -65,23 +65,13 @@ namespace Barotrauma
         private static readonly string LegacySaveFolder = Path.Combine("Data", "Saves");
         private static readonly string LegacyMultiplayerSaveFolder = Path.Combine(LegacySaveFolder, "Multiplayer");
 
-#if OSX
         //"/*user*/Library/Application Support/Daedalic Entertainment GmbH/" on Mac
-        public static readonly string DefaultSaveFolder = Path.Combine(
-            Environment.GetFolderPath(Environment.SpecialFolder.Personal), 
-            "Library",
-            "Application Support",
-            "Daedalic Entertainment GmbH",
-            "Barotrauma");
-#else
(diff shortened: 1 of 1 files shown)


Regalis11 (commented):
Thanks for the PR, and sorry for taking so long to take a proper look at this.

I'm not at all familiar with how these folders are intended to work on MacOS - why is it more correct to use `/Library/Application Support/` rather than `/Documents/Library/Application Support/` ? Does this cause some kind of issues?

Also, wouldn't this PR break make the save files disappear again? I think if we were to change the path back, there should be some kind of backwards compatibility logic in place to find the saves from the previous folder too.

PR #16340, opened 2025-08-06
Pirate Captain Hat Changes

711885:
This adds a bandanaless version of the Separatist Captain Hat as I feel it should be included in the base game, also to add a lil bit more variety and flavour cuz I LOVE FLAVOUR.
It costs 15 marks less because the bandana costs 15 marks, decreased the normal Sep Captain Hat's min-maxavailables as an attempt to make it have sense, since idk about going up to a dirty merchant and them selling you a bandana, a hat, and a hat with a bandana all being separate options
I tried to integrate it into NPCs' prefabs and etc, wanting it to be like 50% chance for each, but I don't really know if I did that, you're gonna have to check if it works correctly
Also made the normal Sep Captain Hat's inventory sprite appear with the bandana as well, to help differentiate them a lil bit
image
image

### DIFF
--- a/Barotrauma/Content/Jobs.xml
+++ b/Barotrauma/Content/Jobs.xml
@@ -41,6 +41,24 @@
     </ItemSet>
     <ItemSet>
       <Item identifier="idcard" equip="true" showpreview="false"/>
+      <Item identifier="captainscap1" identifierteam2="piratecaptainhat2" equip="true"/>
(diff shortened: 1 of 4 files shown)


Regalis11:
Thank you for the PR! I merged the changes to our private repo now, the changes will be coming in a future update. Ended up doing some adjustments to the XML though, imo it's better to define the 2nd hat as a variant of the 1st one so it can just override the sprites instead of redefining everything.

PR #16365, opened 2025-08-20
Disallow using partial condition fpga for circuit box (in-game)

OlegBSTU:
This disallows players to use partial condition fpga circuit for adding components into a circuit box in-game. 
The goal of this is to remove a gameplay exploit where you could recover your fpga if it was damaged, without using any resource:

1. Damage your fpga circuit (enable a mod to do this, for example I was playing with NT Cybernetics Enhanced);
2. Use it to add a component in a circuit box;
3. Remove the component;
4. Obtain a full condition fpga circuit, without a cost.

Changes:
- Added full item condition check when adding a component into the circuit box in-game


Regalis11 (changes requested):
I don't think this is a good way to handle the check. There should not be a hard-coded check for a condition of 100 - what if the required item is something with less than 100 health?

PR #16930, opened 2026-02-11
Penetration for Explosions

NotAlwaysTrue:
Changes:
Added Penetration for Explosions.
Made necessary modifications to Character.AddDamage() to make things work. This change should not affact other calls.

Reason:
1. Doing this with Harmony/Lua is way too hard.
2. Explosion should have a penetration for armored targets since StatusEffect has one but Explosion doesn't.

This PR come with a DEBUGONLY mod for testing.
A Divingsuit with -100% explosiondamage damagemodifier. Name/Identifier: ExplosionTestSuit
Two 40mm grenades:
A grenade with 100% penetration and 100 explosiondamage when explode. Name/Identifier: 40mmgrenadefullpen
A grenade with no penetration defined (default mode) and 100 explosiondamage when explode. Name/Identifier: 40mmgrenadenopen

To Test:
1. Find a stationary targe, another player or bot with AI disabled will be fine.
2. Load a Grenade Launcher with "40mmgrenadenopen"
3. Shoot the target for damage results. Expected: The target was damaged
4. Revive the target and make the target wear a "ExplosionTestSuit"
5. Shoot the target for damage results. Expected: The target was intact
6. Load the same Grenade Launcher with "40mmgrenadefullpen"
7. Shoot the target for damage results. Expected: The target was damaged

Note:
If DeflectProjectile was not set to true, penetration defined in the test mod will not work and no damage on the target.


NotAlwaysTrue:
Folder "Libraries/moonsharp"
Folder "luacs-docs/lua/libs/ldoc"
from commit f60ff24

Both should be removed. They are from my LuaCsForBarotrauma Fork and somehow get to here.
I cannot perform this remove because Github said "Action unavailable"

Edited because in perview version github shows them as new files but in classic they are folders

NotAlwaysTrue:
This should be an issue instead of a PR so im closing it

PR #16936, opened 2026-02-13
Don't force Outpost spawn inside a hull (Allow custom starter outpost)

ubertpendragon:
When determining where to spawn the crew in outposts, the game checks for a valid spawnpoint **INSIDE A HULL**

This check is useless, as you will always put spawnpoints inside hulls anyway. 

https://github.com/FakeFishGames/Barotrauma/blob/783499f152669a89745e4f9fa46167b6f2bf5195/Barotrauma/BarotraumaShared/SharedSource/GameSession/CrewManager.cs#L393-L405

Removing this check will finally allow me to add a cool cinematic starter outpost intro to the campaign by spawning players in the descending shuttle.
I'm even willing to give you the whole thing for vanilla if you make it work

image

image

https://cdn.discordapp.com/attachments/668819422919524352/1372243853498912838/StarterTeleport.mp4?ex=69909dad&is=698f4c2d&hm=8c11d69a11df110f9b3908b3b457e5cc0be234db4fb127d79a9b5bdbae6d89ff&

### DIFF
--- a/Barotrauma/BarotraumaShared/SharedSource/GameSession/CrewManager.cs
+++ b/Barotrauma/BarotraumaShared/SharedSource/GameSession/CrewManager.cs
@@ -399,7 +399,6 @@ namespace Barotrauma
             return WayPoint.WayPointList.FindAll(wp =>
                     wp.SpawnType == SpawnType.Human &&
                     wp.Submarine == Level.Loaded.StartOutpost &&
-                    wp.CurrentHull != null &&
                     wp.CurrentHull.OutpostModuleTags.Contains("airlock".ToIdentifier()));
         }
 


Barotrauma/BarotraumaShared/SharedSource/GameSession/CrewManager.cs
```diff
@@ -399,7 +399,6 @@ public List<WayPoint> GetOutpostSpawnpoints()
             return WayPoint.WayPointList.FindAll(wp =>
                     wp.SpawnType == SpawnType.Human &&
                     wp.Submarine == Level.Loaded.StartOutpost &&
-                    wp.CurrentHull != null &&
```
evilfactory, line 402: This will throw a null reference exception if `wp.CurrentHull` is null because of the check below

evilfactory (changes requested):
We discussed this a bit on discord, but this needs a different approach, even if you were to remove the other check, it would cause you to just spawn in a random human spawnpoint in the station instead of the supposed airlock module.

Barotrauma/BarotraumaShared/SharedSource/GameSession/CrewManager.cs
ubertpendragon: I frankly have no idea how to do this, just thought removing that one line would be enough 😔 

I guess the issue is that it checks if the hull is inside an "Airlock" outpost module and that's how the game knows and spawns you in the airlock. At this point it would be simpler to just have a priority rule to spawn players in spawnpoints marked as "starterspawn" or something

Sorry for the crappy PR

PR #16963, opened 2026-03-04
Parallelize power grid updates and fix gap shuffle allocation

Renart-Celtique:
Reduce FPS drops on complex submarines by eliminating two main-loop bottlenecks: parallel power grid solving and zero-allocation gap shuffling.
The electrical grid solver was single-threaded despite grids being fully independent — now computed in parallel via Parallel.ForEach, with device notifications kept sequential to avoid race conditions. Gap updates were allocating a new sorted array every frame via LINQ — replaced with a pre-allocated Fisher-Yates shuffle buffer.

### DIFF
--- a/Barotrauma/BarotraumaShared/SharedSource/Items/Components/Power/Powered.cs
+++ b/Barotrauma/BarotraumaShared/SharedSource/Items/Components/Power/Powered.cs
@@ -592,45 +592,57 @@ namespace Barotrauma.Items.Components
             // NOTE: GridResolved is kept sequential below because some devices (e.g. RelayComponent)
             // read Voltage from the grid on the *other* side of the relay, which may not have
             // finished its parallel task yet.
-            Parallel.ForEach(Grids.Values, grid =>
+            try
             {
-                //Iterate through the priority src groups lowest first
-                foreach (PowerSourceGroup scrGroup in grid.PowerSourceGroups.Values)
+                Parallel.ForEach(Grids.Values, grid =>
                 {
-                    scrGroup.MinMaxPower = PowerRange.Zero;
-
-                    //Iterate through all connections in the group to get their minmax power and sum them
-                    foreach (Connection c in scrGroup.Connections)
+                    //Iterate through the priority src groups lowest first
+                    foreach (PowerSourceGroup scrGroup in grid.PowerSourceGroups.Values)
                     {
-                        foreach (var device in c.Item.GetComponents<Powered>())
+                        scrGroup.MinMaxPower = PowerRange.Zero;
+
+                        //Iterate through all connections in the group to get their minmax power and sum them
+                        foreach (Connection c in scrGroup.Connections)
                         {
-                            scrGroup.MinMaxPower += device.MinMaxPowerOut(c, grid.Load);
+                            foreach (var device in c.Item.GetComponents<Powered>())
+                            {
+                                scrGroup.MinMaxPower += device.MinMaxPowerOut(c, grid.Load);
+                            }
                         }
+
+                        //Iterate through all connections to get their final power out provided the min max information
+                        float addedPower = 0;
+                        foreach (Connection c in scrGroup.Connections)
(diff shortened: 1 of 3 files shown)


Regalis11:
Thank you for the PR!

Seeing that this is PR is AI generated, before we consider reviewing this more deeply, I need to ask whether you've actually verified the hypotheses the AI has made about the performance, or tested that the results do what they say.

Based on the profiling I have done, it doesn't seem to me that parallelizing the grid updates would give us much performance benefit. Updating the grids doesn't seem to be a major performance problem, and moreover, I'm not 100% sure that the claim "grids are independent of each other — cross-grid work could run in parallel" is true - there are devices like relays which treat the input and output sides as separate grids, and the input affects the behavior of the output and vice versa. Have you looked into whether it's actually safe to multithread, or are you just trusting what the AI says? Have you measured how big of a performance improvement this multithreading gives or whether the gains are significant enough to warrant the additional overhead of the multithreading?

The gap shuffling is also definitely not a major bottleneck, but the proposed optimization is safe and simple, so that one we could definitely use.

Renart-Celtique:
Hello !
Let me be fully honest about what happened here.

First, this PR landed on the wrong repo by mistake. I meant to keep it on my fork until it was actually ready. Truly sorry for the noise.

The real motivation is simple: I play the game regularly with friends and we get heavy FPS drops on complex submarines. I started auditing the codebase to understand why. While doing that, I noticed with btop that during a session Barotrauma saturates 3-4 threads while the other 12 stay at ~0%. That's what led me to the multithreading hypothesis. If the game loop is bottlenecked on a few threads while the rest of the CPU sits idle, parallelizing independent work seemed worth exploring for me.

That said, I know btop only shows thread saturation, not where in the game loop the time is actually spent. I don't have fine-grained profiling data yet. I have Linux build issues blocking me from going further right now.

On top of that, we play with several mods that change some game configs, and I still need to set up a proper test environment that accounts for those before drawing any conclusions. So this was still very much a hypothesis I wanted to validate, not something I was confident enough to propose as an actual fix.

I don't care about merging code for the sake of it, I just want the game to run better. It's a game I genuinely enjoy and want to see improve. 
If you've done profiling and the real bottleneck is somewhere else entirely, I'd genuinely rather know that and investigate there instead.

Do you have any insight into what's actually eating performance in multiplayer on large subs?

NotAlwaysTrue:
Hello.

I had done these in ProjetEP program and i'm afraid there are very limited improvements
The main issue I can tell is update lags caused by item update and you will usually see these lags in modded saves/late-game saves, typically with a bunch of items.
Under most circumstances Gaps/PowerGrids won't take as much time as items did.

It looks like client counts will also affact SV performance but I didn't track that

The main logic is single-threaded so it is possible your CPU useage was low :( and due to the bases the game is built on it is really hard to threading the game

Regalis11:
No worries, there were some useful bits here too (mainly the gap shuffling optimization which I think we can use in the vanilla game)!

As NotAlwaysTrue mentioned above, one of the main bottlenecks is the item update logic. Which unfortunately is an extremely difficult thing to multithread, since items can and do affect each other in various ways and the order in which they're updated can matter. Status effects and moddability also make this much more complicated, since the items can do "essentially anything" when they're being updated.

Another major performance bottleneck is the physics simulation, particularly in situations where there's lots of characters/monsters and/or multiple submarines around. Even though some parts of the physics simulation are multithreaded, a large portion of it cannot be easily parallelized for similar reasons as item updates: OnCollision and OnSeparation callbacks which trigger during the physics simulation can and do affect other things (again in arbitrary ways due to modding), and here too the update order matters.

In any case, I think we can close this PR especially if it wasn't intended to be published here. But if you do manage to implement some optimizations (even if it's with the help of AI), and those optimizations are based on real performance issues identified with profiling and the results of the optimization verified along with verifying that the code is safe to parallelize, we'd be happy to see a PR about those!

PR #16970, opened 2026-03-07
Fix CanReceive in VoipClient being called incorrectly while being entirely unneeded.

BramvanZijp:
Remove an incorrect & redundant check client-side.

Reasoning why is explained in https://github.com/FakeFishGames/Barotrauma/discussions/16969
To sum it up: The server never even sends a packet to the client if CanReceive returns false.
So even if the check was being called correctly, it is not needed, but does prevent mod makers from changing the outcome of CanReceive without clientside lua, which impacts mods such as Radio Broadcast, which I would use if it was server-only. (Its hard to get randoms to stay in a server if they need to install clientside lua)

However, it is also being called wrong, with the 2 arguments being swapped, now checking if the sender can receive the receiver, which should be the other way around, so we can kill 2 birds with 1 stone by removing the redundant check, fixing that bug in the process.

I brought both points up in the Luatrauma discord, to which Evil Factory said:
"yea i think its mostly redundant, probably just some really old code that didn't that much thought put on"

### DIFF
--- a/Barotrauma/BarotraumaClient/ClientSource/Networking/Voip/VoipClient.cs
+++ b/Barotrauma/BarotraumaClient/ClientSource/Networking/Voip/VoipClient.cs
@@ -120,7 +120,7 @@ namespace Barotrauma.Networking
                     var messageType = 
                         !client.VoipQueue.ForceLocal && 
                         ChatMessage.CanUseRadio(client.Character, out senderRadio) && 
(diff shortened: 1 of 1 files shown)


BramvanZijp:
Resolves #16969

BramvanZijp:
Based on the info provided by Nilanth, it is not as simple as removing this check, which was deliberately added to patch a bug. 
I will close this PR, and open one with a solution without that drawback.

Issue #16978, opened 2026-03-13
CanReceive being called in VoipClient both has the arguments swapped and is entirely redundant.


NilanthAnimosus:
### Discussed in https://github.com/FakeFishGames/Barotrauma/discussions/16969

<div type='discussions-op-text'>

<sup>Originally posted by **BramvanZijp** March  7, 2026</sup>
### Disclaimers

- [X] I have searched the issue tracker to check if the issue has already been reported.
- [X] My issue happened while using mods.

### What happened?

I wanted to make a serverside only mod to allow me to do something simular to the mod Radio Broadcast (https://steamcommunity.com/sharedfiles/filedetails/?id=3341600653), but a bit more complex. To water it down, I need to be able to override the CanReceive function under Barotrauma.Items.Components.WifiComponent.

And so I did, and it worked until VC was tested, where the other player was unable to hear me. (Though the speaker icon did appear in the player list next to my name, with chat messages being routed perfectly fine.)
I later found out this was because, despite the check being ran in VoipServer with all needed info, the check is ran AGAIN in VoipClient under Read(), at the following line: https://github.com/evilfactory/LuaCsForBarotrauma/blob/9e957a75b005f96006475d4562b3c43deaf4de34/Barotrauma/BarotraumaClient/ClientSource/Networking/Voip/VoipClient.cs#L123

Now, I was prepared to give up on my mod being server-only, until I realised 2 things.
1. The function is being called incorrectly. It is checking if the SENDER can RECEIVE the RECIPIENT, instead of the recipient receiving the sender. This is a bug that RadioBroadcast has already had to work around by inverting the values client-side as seen here: 
<img width="942" height="187" alt="image" src="https://github.com/user-attachments/assets/686bd005-917f-4837-aea0-5eb7b743f877" />

2. The function being called is entirely redundant. If CanReceive is false, the server never even sends the packet. So not only is the function being called wrong, its also not needing to be called at all, with the game being able to work just as well if that check is deleted from the code clientside.

So, given that it is being called incorrectly, without needing to be called at all: It would not only fix the issue, but also help mod makers a lot, if that canreceive check is removed from the client entirely, given it is simply not needed.
(This matters in my case because public servers are MUCH harder to get new players into if said new players need clientside lua installed)

(I will also note Evil Factory mentioned in the luatrauma discord that the relevant section of code is mostly redundant and likely hasnt had much thought put on it.)

### Reproduction steps

[...]

PR #16982, opened 2026-03-13
Change VoipClient to be informed of the message being radio or local by VoipServer instead of by re-running the checks ran server-side..

BramvanZijp:
Change VoipClient to be informed of the message being radio or local by VoipServer instead of by re-running the checks ran server-side.
 
Reasoning why is explained in #16969 (Moved to #16978).

Unlike #16970, which simply removes the check, this PR takes advantage of the fact that that exact check is already ran server-side, and simply includes the result of that check in the packet sent to the client as a boolean value.

NOTE: This WILL require testing with another person (which I am unable to alone), though I see no reason why it would not function.

### DIFF
--- a/Barotrauma/BarotraumaClient/ClientSource/Networking/Voip/VoipClient.cs
+++ b/Barotrauma/BarotraumaClient/ClientSource/Networking/Voip/VoipClient.cs
@@ -118,16 +118,22 @@ namespace Barotrauma.Networking
                     float rangeMultiplier = spectating ? 2.0f : 1.0f;
                     WifiComponent senderRadio = null;
 
-                    //Checks if the radio noise filter should be applied. CanUseRadio is there because the code after uses senderRadio.
-                    var messageType = ChatMessage.CanUseRadio(client.Character, out senderRadio) && isRadio ? ChatMessageType.Radio : ChatMessageType.Default;
+                    //Checks if the radio noise filter should be applied.
+                    var messageType = isRadio ? ChatMessageType.Radio : ChatMessageType.Default;
+                    
                     client.Character.ShowTextlessSpeechBubble(1.25f, ChatMessage.MessageColor[(int)messageType]);
 
                     client.VoipSound.UseRadioFilter = messageType == ChatMessageType.Radio && !GameSettings.CurrentConfig.Audio.DisableVoiceChatFilters;
                     client.RadioNoise = 0.0f;
                     if (messageType == ChatMessageType.Radio)
                     {
+                        //If the client cannot establish a radio, use a headsets default range as a fallback to calculate the radio noise.
+                        //This cannot happen in an un-modded setting as CanUseRadio is part of the server side check for isRadio to be true.
+                        ChatMessage.CanUseRadio(client.Character, out senderRadio);
+                        float senderRadioRange = (senderRadio == null) ? 35000.0f : senderRadio.Range;
+                    
                         client.VoipSound.UsingRadio = true;
-                        client.VoipSound.SetRange(senderRadio.Range * RangeNear * speechImpedimentMultiplier * rangeMultiplier, senderRadio.Range * speechImpedimentMultiplier * rangeMultiplier);
+                        client.VoipSound.SetRange(senderRadioRange * RangeNear * speechImpedimentMultiplier * rangeMultiplier, senderRadioRange * speechImpedimentMultiplier * rangeMultiplier);
                         if (distanceFactor > RangeNear && !spectating)
                         {
                             //noise starts increasing exponentially after 40% range
--- a/Barotrauma/BarotraumaClient/ClientSource/Networking/Voip/VoipClient.cs
+++ b/Barotrauma/BarotraumaClient/ClientSource/Networking/Voip/VoipClient.cs
@@ -89,6 +89,7 @@ namespace Barotrauma.Networking
         {
             byte queueId = msg.ReadByte();
             float distanceFactor = msg.ReadRangedSingle(0.0f, 1.0f, 8);
+            bool isRadio = msg.ReadBoolean();
             VoipQueue queue = queues.Find(q => q.QueueID == queueId);
 
             if (queue == null)
@@ -117,11 +118,7 @@ namespace Barotrauma.Networking
                     float rangeMultiplier = spectating ? 2.0f : 1.0f;
                     WifiComponent senderRadio = null;
 
-                    var messageType = 
-                        !client.VoipQueue.ForceLocal && 
-                        ChatMessage.CanUseRadio(client.Character, out senderRadio) && 
-                        (spectating || (ChatMessage.CanUseRadio(Character.Controlled, out var recipientRadio) && senderRadio.CanReceive(recipientRadio)))
-                            ? ChatMessageType.Radio : ChatMessageType.Default;
+                    var messageType = isRadio ? ChatMessageType.Radio : ChatMessageType.Default;
                     client.Character.ShowTextlessSpeechBubble(1.25f, ChatMessage.MessageColor[(int)messageType]);
 
                     client.VoipSound.UseRadioFilter = messageType == ChatMessageType.Radio && !GameSettings.CurrentConfig.Audio.DisableVoiceChatFilters;
--- a/Barotrauma/BarotraumaClient/ClientSource/Networking/Voip/VoipClient.cs
+++ b/Barotrauma/BarotraumaClient/ClientSource/Networking/Voip/VoipClient.cs
@@ -118,7 +118,8 @@ namespace Barotrauma.Networking
                     float rangeMultiplier = spectating ? 2.0f : 1.0f;
                     WifiComponent senderRadio = null;
 
-                    var messageType = isRadio ? ChatMessageType.Radio : ChatMessageType.Default;
+                    //Checks if the radio noise filter should be applied. CanUseRadio is there because the code after uses senderRadio.
+                    var messageType = ChatMessage.CanUseRadio(client.Character, out senderRadio) && isRadio ? ChatMessageType.Radio : ChatMessageType.Default;
                     client.Character.ShowTextlessSpeechBubble(1.25f, ChatMessage.MessageColor[(int)messageType]);
 
                     client.VoipSound.UseRadioFilter = messageType == ChatMessageType.Radio && !GameSettings.CurrentConfig.Audio.DisableVoiceChatFilters;
--- a/Barotrauma/BarotraumaServer/ServerSource/Networking/Voip/VoipServer.cs
+++ b/Barotrauma/BarotraumaServer/ServerSource/Networking/Voip/VoipServer.cs
@@ -93,7 +93,7 @@ namespace Barotrauma.Networking
             //sender can't speak
             if (sender.Character != null && sender.Character.SpeechImpediment >= 100.0f) { return false; }
 
-            //check if the message can be sent via radio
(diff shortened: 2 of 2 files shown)


BramvanZijp:
Resolves #16978

Barotrauma/BarotraumaServer/ServerSource/Networking/Voip/VoipServer.cs
```diff
@@ -51,29 +51,32 @@ public void SendToClients(List<Client> clients)
...
         }
 
-        private static bool CanReceive(Client sender, Client recipient, out float distanceFactor)
+        private static bool CanReceive(Client sender, Client recipient, out float distanceFactor, out bool isRadio)
         {
             if (Screen.Selected != GameMain.GameScreen) 
             {
                 distanceFactor = 0.0f;
+                isRadio = false;
                 return true;
```
BramvanZijp, line 75: Also I just realised but this looks like its intended to return false based on the way it is written, no?

BramvanZijp: Nevermind, just me misunderstanding what that is doing.

Barotrauma/BarotraumaServer/ServerSource/Networking/Voip/VoipServer.cs
```diff
@@ -98,13 +101,15 @@ private static bool CanReceive(Client sender, Client recipient, out float distan
             {
                 if (recipientSpectating)
                 {
-                    if (recipient.SpectatePos == null) { return true; }
+                    if (recipient.SpectatePos == null) { isRadio = true; return true; }
```
Regalis11, line 104: This is a little unclear to me. Why would the message be considered to be sent via radio if the recipient is spectating?

Barotrauma/BarotraumaServer/ServerSource/Networking/Voip/VoipServer.cs
```diff
@@ -98,13 +101,15 @@ private static bool CanReceive(Client sender, Client recipient, out float distan
...
                     distanceFactor = MathHelper.Clamp(Vector2.Distance(sender.Character.WorldPosition, recipient.SpectatePos.Value) / senderRadio.Range, 0.0f, 1.0f);
+                    isRadio = true;
```
Regalis11, line 106: This is a little unclear to me. Why would the message be considered to be sent via radio if the recipient is spectating?

Barotrauma/BarotraumaServer/ServerSource/Networking/Voip/VoipServer.cs
```diff
@@ -51,29 +51,32 @@ public void SendToClients(List<Client> clients)
...
+                    msg.WriteBoolean(isRadio);
                     queue.Write(msg);
                     
                     netServer.Send(msg, recipient.Connection, DeliveryMethod.Unreliable);
                 }
             }
         }
 
-        private static bool CanReceive(Client sender, Client recipient, out float distanceFactor)
+        private static bool CanReceive(Client sender, Client recipient, out float distanceFactor, out bool isRadio)
```
Regalis11, line 69: The new isRadio argument could use a comment - what does it actually mean? Clearly not whether the message was sent or received via a radio, since it's also true in cases where the recipient is spectating and can't have a radio.

Regalis11 (changes requested):
I think the approach looks mostly good, however I found some parts of the code confusing. It's now quite unclear what the `isRadio` variables mean in the server-side code.

Barotrauma/BarotraumaServer/ServerSource/Networking/Voip/VoipServer.cs
BramvanZijp: Because the sender IS still transmitting via radio in the specified example. It confirms that the sender is not forcing local chat, and that they are capable of sending on radio, before reaching that check. 

However that leads me to asking: Where the hell does it check for a spectator hearing LOCAL chat if their camera is close enough, which is something I cannot answer so I will add a bugfix for that. 
EDIT: I was blind, nvm.

Barotrauma/BarotraumaServer/ServerSource/Networking/Voip/VoipServer.cs
BramvanZijp: Because the sender IS still transmitting via radio in the specified example. It confirms that the sender is not forcing local chat, and that they are capable of sending on radio, before reaching that check. 

In this specific case, the distance even considers the range of the senders radio.

However that leads me to asking: Where the hell does it check for a spectator hearing LOCAL chat if their camera is close enough, which is something I cannot answer so I will add a bugfix for that.

Barotrauma/BarotraumaServer/ServerSource/Networking/Voip/VoipServer.cs
BramvanZijp: It effectively checks if the client should apply the radio voice filter or not, which is the only reason why that check was there. For the case of spectators, if they are hearing it due to it being sent on radio, that should, IMO, still apply that radio effect. Ill add a comment to clarify this.

Barotrauma/BarotraumaServer/ServerSource/Networking/Voip/VoipServer.cs
Regalis11: > Where the hell does it check for a spectator hearing LOCAL chat if their camera is close enough

On line 121

BramvanZijp: > > Where the hell does it check for a spectator hearing LOCAL chat if their camera is close enough
> 
> On line 121

Yeah, just saw that after adding the comment. 
As for the original issue, given the spectator is hearing it because it is being sent via radio, I feel like isRadio SHOULD be true, so that the spectator hears the radio sound filter.

BramvanZijp:
Had to add one of the CanUseRadio checks back to VoipClient because that outputs the senderRadio variable, which the client does need for the radio messages.

Furthermore, clarifying comments were added to more clearly explain isRadio's purpose.
I have not had a chance to test the changes with anyone yet, but if I find somebody willing to help, I will.

BramvanZijp:
> Had to add one of the CanUseRadio checks back to VoipClient because that outputs the senderRadio variable, which the client does need for the radio messages.
> 
> Furthermore, clarifying comments were added to more clearly explain isRadio's purpose. I have not had a chance to test the changes with anyone yet, but if I find somebody willing to help, I will.

If we want to instead include senderRadio.range as another part of the packet and then keep CanUseRadio serverside only, we COULD, though I am not 100% sure if thats really worth it. (As somebody not knowledgable on the networking part, I am not sure how much bigger we really want to make each message to each client, because IIRC they are sent VERY rapidly. If adding an extra number is a non-issue, however, that WOULD be the ideal solution, but I worry it may cause issues as that would put a hard cap on the max range radios can be given due to the byte size.)

Barotrauma/BarotraumaClient/ClientSource/Networking/Voip/VoipClient.cs
```diff
@@ -117,11 +118,8 @@ public void Read(IReadMessage msg)
                     float rangeMultiplier = spectating ? 2.0f : 1.0f;
                     WifiComponent senderRadio = null;
 
-                    var messageType = 
-                        !client.VoipQueue.ForceLocal && 
-                        ChatMessage.CanUseRadio(client.Character, out senderRadio) && 
-                        (spectating || (ChatMessage.CanUseRadio(Character.Controlled, out var recipientRadio) && senderRadio.CanReceive(recipientRadio)))
-                            ? ChatMessageType.Radio : ChatMessageType.Default;
+                    //Checks if the radio noise filter should be applied. CanUseRadio is there because the code after uses senderRadio.
+                    var messageType = ChatMessage.CanUseRadio(client.Character, out senderRadio) && isRadio ? ChatMessageType.Radio : ChatMessageType.Default;
```
Regalis11: It's still unclear to me why the clients need to determine if the message should be received via the radio here. The server already tells them that information, should the server not have authority on it? 

If the intention is to only use `CanUseRadio` to fetch the `senderRadio`, it seems incorrect to also make it determine the message type.

I'm also unsure how this would now help with the issue you initially reported. If the client is again fully in charge of whether the message should be interpreted as a radio message or not, won't you have the same issue again with your server-side mod?

Regalis11 (changes requested):
The comments clarified this a bit, but one of the newer changes seems strange and potentially incorrect to me.

Barotrauma/BarotraumaClient/ClientSource/Networking/Voip/VoipClient.cs
BramvanZijp: CanUseRadio is a check that neither my mod nor radio broadcast minds.

However I do agree this goes against the intent of the PR, so I will add a workaround by falling back to the default range of a headset IF the client cannot establish a radio and thus range, which should not be possible to happen in vanilla.

BramvanZijp:
Updated the code to simply fetch the radio client-side, and assume the default range value of a headset if that fails (Which it will not unless mods directly impact the outcome of CanUseRadio).

BramvanZijp:
Am I correct in assuming that the code itself probably looks good now, so its just a matter of getting it tested & discussed internally?

Regalis11:
Yeah, I can't spot any further issues in the code on a quick glance at least! I'll have a better look at this again when we're no longer as swamped with the DLC community testing that just started, and then move this to our private repo so we can test it internally. :)

BramvanZijp:
> Yeah, I can't spot any further issues in the code on a quick glance at least! I'll have a better look at this again when we're no longer as swamped with the DLC community testing that just started, and then move this to our private repo so we can test it internally. :)

Alright, sounds good.
Though I will ask if you could dismiss the review requesting changes, since the requested changes have been applied.

Regalis11 (approved):
Looks mostly good to me now. However, I think the way isRadio is set by the clients could still use some improvements: it's set to true separately in each of the branches of the "can use radio" code block, which is unnecessary and somewhat confusing. Reading a line like `if (recipient.SpectatePos == null) { isRadio = true; return true; }` makes me wonder why being a spectator has anything to do with whether the message is a radio message or not, and then on a closer look it turns out it doesn't - it's just set to true in all the different branches of the if-else for no clear reason, instead of doing it once when we've determined it's a radio message.

I'll approve this regardless and adjust that before moving this to a PR in our private repo for review and testing. :) Thank you for the contribution!

BramvanZijp:
> Looks mostly good to me now. However, I think the way isRadio is set by the clients could still use some improvements: it's set to true separately in each of the branches of the "can use radio" code block, which is unnecessary and somewhat confusing. Reading a line like `if (recipient.SpectatePos == null) { isRadio = true; return true; }` makes me wonder why being a spectator has anything to do with whether the message is a radio message or not, and then on a closer look it turns out it doesn't - it's just set to true in all the different branches of the if-else for no clear reason, instead of doing it once when we've determined it's a radio message.
> 
> I'll approve this regardless and adjust that before moving this to a PR in our private repo for review and testing. :) Thank you for the contribution!

Appreciate it. Is there a specific way for me to know if/when these changes make it into the basegame, or should I assume it to be alongside the next major update?

Regalis11:
It unfortunately didn't make it in time for the upcoming Spring Update, but will be included in the one after that (coming out some time closer to the summer)!

PR #17377, opened 2026-04-26
Fix negative-amount wallet transfer exploit

Astro1247:
Wallet transfers accept an Amount field sent by client with no lower bound check. Sending a NetWalletTransfer packet with a negative value causes TryDeduct to behave incorrectly: the balance check passes (any non-negative balance satisfies Balance >= negative), and the subsequent subtraction inverts sign - effectively adding money to the sender's wallet and draining the target or adding money if target wallet does not have enough. The exploit is trivially reproducible by crafting a single TRANSFER_MONEY packet and currently is being used in the wild on public servers by malicious actors to ruin campaigns alongside with crash exploits and mass buying items using it.

The fix adds a one-line guard in ServerReadMoney that rejects any transfer with Amount <= 0 before any wallet logic runs. No changes to the shared network struct or wire format are needed - the existing UI already prevents sending non-positive values, so this only blocks crafted packets.


evilfactory:
Thank you for the contribution! I have now moved the PR to our internal repo for testing, closing

PR #17405, opened 2026-04-30
Console command additions

HawkinsThe71st:
I wanted to try my hand at adding a few minor additions to some console commands.

### give/spawnitem
Added quality parameter which allows spawning in items with quality.

### teleportsub
Added parameter to pick which submarine to teleport.

### spawn/spawncharacter
Added name parameter which allows renaming spawned human characters instead of a random name.

### showseed
Displays map seed if used in campaign mode.

### DIFF
--- a/Barotrauma/BarotraumaClient/ClientSource/DebugConsole.cs
+++ b/Barotrauma/BarotraumaClient/ClientSource/DebugConsole.cs
@@ -4205,6 +4205,7 @@ namespace Barotrauma
                 }
                 else
                 {
+                    if (GameMain.GameSession?.GameMode is CampaignMode campaign) { NewMessage("Map seed: " + GameMain.GameSession.Campaign.Map.Seed); };
                     NewMessage("Level seed: " + Level.Loaded.Seed);
                     NewMessage("Level generation params: " + Level.Loaded.GenerationParams.Identifier);
                     NewMessage("Adjacent locations: " + (Level.Loaded.StartLocation?.Type.Identifier ?? "none".ToIdentifier()) + ", " + (Level.Loaded.StartLocation?.Type.Identifier ?? "none".ToIdentifier()));
--- a/Barotrauma/BarotraumaClient/ClientSource/DebugConsole.cs
+++ b/Barotrauma/BarotraumaClient/ClientSource/DebugConsole.cs
@@ -4205,7 +4205,7 @@ namespace Barotrauma
                 }
                 else
                 {
-                    if (GameMain.GameSession?.GameMode is CampaignMode campaign) { NewMessage("Map seed: " + GameMain.GameSession.Campaign.Map.Seed); };
+                    if (GameMain.GameSession?.Map is Map map) { NewMessage("Map seed: " + map.Seed); };
                     NewMessage("Level seed: " + Level.Loaded.Seed);
                     NewMessage("Level generation params: " + Level.Loaded.GenerationParams.Identifier);
                     NewMessage("Adjacent locations: " + (Level.Loaded.StartLocation?.Type.Identifier ?? "none".ToIdentifier()) + ", " + (Level.Loaded.StartLocation?.Type.Identifier ?? "none".ToIdentifier()));
--- a/Barotrauma/BarotraumaServer/ServerSource/DebugConsole.cs
+++ b/Barotrauma/BarotraumaServer/ServerSource/DebugConsole.cs
@@ -1800,10 +1800,10 @@ namespace Barotrauma
                     Submarine submarineToTeleport = Submarine.MainSub;
                     if (args.Length > 1)
                     {
-                        foreach(Submarine sub in Submarine.Loaded.Where(s => (!s.Info.IsBeacon && !s.Info.IsOutpost && !s.Info.IsRuin && !s.Info.IsWreck)))
-                        {
-                            if (sub.Info.Name == args[1]) { submarineToTeleport = sub; }
-                        }
+                    foreach(Submarine sub in Submarine.Loaded.Where(s => (!s.Info.IsBeacon && !s.Info.IsOutpost && !s.Info.IsRuin && !s.Info.IsWreck)))
+                    {
+                        if ((sub.Info.Name+"_"+sub.TeamID) == args[1]) { submarineToTeleport = sub; }
+                    }
                     }
 
                     if (args.Length == 0 || args[0].Equals("cursor", StringComparison.OrdinalIgnoreCase))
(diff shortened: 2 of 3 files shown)


Barotrauma/BarotraumaShared/SharedSource/DebugConsole.cs
```diff
@@ -3140,6 +3180,14 @@ private static IEnumerable<string> GetItemNameOrIdParams()
             }
         }
 
+        private static IEnumerable<string> GetQualityParams()
+        {
+            yield return "normal";
+            yield return "good";
+            yield return "excellent";
+            yield return "masterwork";
+        }
```
Regalis11: I see no reason for this to be an enumerator, why not make it an immutable collection instead?

Barotrauma/BarotraumaClient/ClientSource/DebugConsole.cs
```diff
@@ -4205,6 +4205,7 @@ void getTextsFromElement(XElement element, List<string> list, string parentName)
                 }
                 else
                 {
+                    if (GameMain.GameSession?.GameMode is CampaignMode campaign) { NewMessage("Map seed: " + GameMain.GameSession.Campaign.Map.Seed); };
```
Regalis11: The variable campaign is unused. It seems odd to check the GameMode is a campaign but then use an entirely different property and not do any null checks on it. This could be

`if (GameMain.GameSession?.Map is Map map) { NewMessage("Map seed: " + map.Seed); };`

Barotrauma/BarotraumaServer/ServerSource/DebugConsole.cs
```diff
@@ -1797,22 +1797,31 @@ string eventDebugStr(ServerEntityEvent ev)
                 (Client client, Vector2 cursorWorldPos, string[] args) =>
                 {
                     if (Submarine.MainSub == null || Level.Loaded == null) { return; }
+                    Submarine submarineToTeleport = Submarine.MainSub;
+                    if (args.Length > 1)
+                    {
+                    foreach(Submarine sub in Submarine.Loaded.Where(s => (!s.Info.IsBeacon && !s.Info.IsOutpost && !s.Info.IsRuin && !s.Info.IsWreck)))
+                    {
+                        if ((sub.Info.Name+"_"+sub.TeamID) == args[1]) { submarineToTeleport = sub; }
+                    }
```
Regalis11: Broken indentation

Barotrauma/BarotraumaServer/ServerSource/DebugConsole.cs
```diff
@@ -1797,22 +1797,31 @@ string eventDebugStr(ServerEntityEvent ev)
                 (Client client, Vector2 cursorWorldPos, string[] args) =>
                 {
                     if (Submarine.MainSub == null || Level.Loaded == null) { return; }
+                    Submarine submarineToTeleport = Submarine.MainSub;
+                    if (args.Length > 1)
+                    {
+                    foreach(Submarine sub in Submarine.Loaded.Where(s => (!s.Info.IsBeacon && !s.Info.IsOutpost && !s.Info.IsRuin && !s.Info.IsWreck)))
+                    {
+                        if ((sub.Info.Name+"_"+sub.TeamID) == args[1]) { submarineToTeleport = sub; }
```
Regalis11: The requirement to add the team to the submarine name isn't clearly communicated anywhere. I also think this is missing a `break`: there's no reason to keep enumerating more subs if a matching one was already found. Or is the intention to choose the last matching submarine? If so, it would be better to enumerate the loaded subs backwards.

Barotrauma/BarotraumaServer/ServerSource/DebugConsole.cs
```diff
@@ -1797,22 +1797,31 @@ string eventDebugStr(ServerEntityEvent ev)
                 (Client client, Vector2 cursorWorldPos, string[] args) =>
                 {
                     if (Submarine.MainSub == null || Level.Loaded == null) { return; }
+                    Submarine submarineToTeleport = Submarine.MainSub;
+                    if (args.Length > 1)
+                    {
+                    foreach(Submarine sub in Submarine.Loaded.Where(s => (!s.Info.IsBeacon && !s.Info.IsOutpost && !s.Info.IsRuin && !s.Info.IsWreck)))
```
Regalis11: 1. Redundant parentheses in the expression. 
2. In my opinion it'd also be better to check whether the sub has a dynamic body rather than checking individual submarine types separately. 
3. This code is repeated several times - it would be really easy to make the mistake of forgetting to modify these 3 duplicate checks if at some point in the future an additional submarine type was added.
4. Why is moving submarines of the type OutpostModule allowed, when moving outposts isn't?

Barotrauma/BarotraumaShared/SharedSource/DebugConsole.cs
```diff
@@ -1323,6 +1329,7 @@ void ToggleGodMode(Character targetCharacter)
                 }
                 else
                 {
+                    if (GameMain.GameSession?.GameMode is CampaignMode campaign) { NewMessage("Map seed: " + GameMain.GameSession.Campaign.Map.Seed); };
```
Regalis11: Same issue here as earlier, null check on GameMode which is then stored in a variable, only to not use that variable for anything and use something else instead.

Barotrauma/BarotraumaShared/SharedSource/DebugConsole.cs
```diff
@@ -3232,7 +3282,30 @@ bool TryGetSpawnPosParam(out string spawnLocation, out int spawnLocationIndex)
...
+                        case "excellent":
+                            itemQuality = 2;
+                            break;
+                        case "masterwork":
+                            itemQuality = 3;
+                            break;
+                        default:
+                            itemQuality = 0;
+                            break;
+                    }
```
Regalis11, line 3292: It seems strange to hard-code these here. This is again a case where a programmer would need to know and remember this check exists here if any changes were done to the qualities or their names.

Regalis11 (changes requested):
Thank you for the contribution! While on a cursory look the changes seem functional, I think there are some code quality issues here that still need some attention.

HawkinsThe71st:
Thank you for the review! Apologies for the poor code quality, this is my first time trying something like this and I'm still quite inexperienced.

Regalis11 (approved):
No problem! :) Overall the code was pretty good though with no functional issues.

Now with your latest revision the only thing I can think of nitpicking about is the name `GetQualityParams` - the name implies it's a function rather than a field, and it's also not that obvious it refers to item qualities when it's just a field in the DebugConsole class with no clear indication it's somehow related to items.

Anyway, I'll move these changes to our private repo, rename that field and pass the PR for our QA team to test. We'll probably be able to get it tested and into the upcoming summer update. Thanks for the contribution again!

Regalis11:
Moved to [a PR in our private repo](https://github.com/FakeFishGames/Barotrauma-development/pull/8052) for internal testing

Regalis11:
This has now been merged in our private repo, coming up in the next unstable build and the upcoming Summer Update.

PR #17514, opened 2026-05-18
Fix/linux nonlatin keyboard layout

TheCyberAngel:
This fixes keyboard bindings on Linux when using non-Latin keyboard layouts such as Ukrainian or Russian.

Previously SDL key events were mapped through layout-dependent keycodes. On non-Latin layouts, keys like WASD could resolve to unmapped Unicode keycodes and become Keys.None, causing movement and default keybindings to break.

The fix adds a scancode-based fallback:

* KeyDown/KeyUp now fall back to physical SDL scancodes if keycode mapping fails
* QwertyToCurrentLayout also falls back to scancodes, preventing reset-to-default bindings from becoming Keys.None

Text input remains layout-aware, so chat and text fields still support non-English input correctly.

Tested on Linux with Ukrainian and Russian keyboard layout:

* gameplay movement works
* rebinding keys works
* resetting controls no longer produces Keys.None
* chat input still supports Kyrylic text


3e849f2e5c:
This has been merged to our private repo now, thank you for the contribution!

PR #17537, opened 2026-05-24
nucleargun Price Reputation Fix

FriedrichNietzsche8:
Item with `nucleargun` identifier has the wrong faction `separatist` instead of the correct `separatists`


Regalis11:
Than you for the contribution! I've now copied this fix to our private repo, fix coming up in the next unstable build and the upcoming Summer Update.

PR #17672, opened 2026-07-06
refactor(loading): refactored and inproved loading speed.

lolrobbe2:
# color:

- removed the unnecesary /255f as in the constructor the exact oposite happend.
- added depthOffset to draw function as to prevent constant add, subtract (default=0 to prevent regressions).
- initialized the hud background color once to prevent the GC from being strangled.

# sound.
- moved the sound system from short variant to the preferred float variant as the constant casting had a significant impact on the loading times and performance.

- this also reduces load on the sound thread wich does not have to also do casting when applying filters.

- also used ArrayPool to prevent GC contention.

- added the FormatMonoF32 and FormatStereoF32 variants

# SubmarineInfo hash

- used filepath+timstamp of the file to calculate the hash as calculating it from the full xml text had a significant impact. this also improved load times by a considerable ammount. (https://github.com/FakeFishGames/Barotrauma/pull/16021)

(in the next PR i will try to improve the file loading more by adtional use of ArrayPool)

### DIFF
--- a/Barotrauma/BarotraumaClient/ClientSource/Characters/AI/Wreck/WreckAI.cs
+++ b/Barotrauma/BarotraumaClient/ClientSource/Characters/AI/Wreck/WreckAI.cs
@@ -1,7 +1,8 @@
 using Barotrauma.Extensions;
-using Microsoft.Xna.Framework;
-using System.Collections.Generic;
 using Barotrauma.Networking;
+using Microsoft.Xna.Framework;
+using Steamworks.Ugc;
+using System.Collections.Generic;
 
 namespace Barotrauma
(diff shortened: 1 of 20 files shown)


Regalis11:
Thank you for the contribution! The same thing as with [Draw Method cleanup/ perf improvments](https://github.com/FakeFishGames/Barotrauma/pull/17681#top) - I think we are going to need more info to start properly looking into this PR, especially considering that this is a large amount of non-trivial and potentially changes to low-level parts of the game's code. Can you provide some profiling data to show how much this improves performance (or whether it actually does improve performance)?

PR #17676, opened 2026-07-07
feat(mem): added buffer pooling for networking code to reduce load on GC

lolrobbe2:
# ArrayPool:
https://learn.microsoft.com/en-us/dotnet/api/system.buffers.arraypool-1?view=net-8.0
by using array pool we can reduce the load on the .NET GC by using ArrayPool wich reuses the buffer instead of Reallocation.
this prevents lag spikes from occuring due to GC overload.

and it is fully thread safe

To faccilitate Easier Use a wrapper class was made called PooledBuffer.

### DIFF
--- a/Barotrauma/BarotraumaClient/ClientSource/Networking/GameClient.cs
+++ b/Barotrauma/BarotraumaClient/ClientSource/Networking/GameClient.cs
@@ -2185,7 +2185,7 @@ namespace Barotrauma.Networking
                             UInt16 updateID = inc.ReadUInt16();
 
                             UInt16 settingsLen = inc.ReadUInt16();
-                            byte[] settingsData = inc.ReadBytes(settingsLen);
+                            PooledBuffer settingsData = inc.ReadBytes(settingsLen);
 
                             bool isInitialUpdate = inc.ReadBoolean();
                             DebugConsole.Log($"Received {(isInitialUpdate ? "initial" : string.Empty)} lobby update ID: {updateID}, last ID: {GameMain.NetLobbyScreen.LastUpdateID}.");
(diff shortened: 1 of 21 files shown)


Regalis11:
Thank you for the contribution! The same thing as with [Draw Method cleanup/ perf improvments](https://github.com/FakeFishGames/Barotrauma/pull/17681#top) - I think we are going to need more info to start properly looking into this PR, especially considering that this is a large amount of non-trivial and potentially volatile changes to low-level parts of the game's code. Can you provide some profiling data to show how much this improves performance (or whether it actually does improve performance)?

lolrobbe2:
this mostly reduces lag spikes from occuring due to the GC being pressured from continues allocations and deallocations from  calling new byte[]. it does this by replacing all these calls in the network code by a pre allocated array from the [SharedArray](https://learn.microsoft.com/en-us/dotnet/api/system.buffers.arraypool-1?view=net-10.0) pool (this is also thread safe).

this should also prevent issues where the server could time out due to a GC spike, potentialy causing disconnects.

general performance should be pretty simmilar or better (have no perfect way of stress testing really, but the server works with the current master to).

i also had to add the serializer.

NOTE: most of the changes are replacing the new byte[] with new PooledBuffer calls where neccesary, no reggression should be noticed if all went well.

Networking stability should be improved however by a large margin.


manny benchmarks of the array pool can be found online to corroborate this. (https://medium.com/@epeshk/the-big-performance-difference-between-arraypools-in-net-b25c9fc5e31d)

lolrobbe2:
the same principle can later also be applied in the loading and parsing of files later on pottentialy

Regalis11:
Thanks, that clarifies the content of the changes a bit!

> manny benchmarks of the array pool can be found online to corroborate this.

Yeah, I don't doubt it's more performant, but how much of a difference does it make in Barotrauma's context, with the way it uses these buffers? I believe when it comes to optimizations, profiling is a must: in order for us to consider some optimization (especially when the changes are as extensive and volatile as this), we need to have data that proves it does what it's intended to do - it's not uncommon that performance problems stem from something else than what one might initially expect, and some optimization turns out to not make much of a difference (or even hinder performance).

PR #17843, opened 2026-08-26
Allow controlling bots in multiplayer

erunehtar:
Adds support for controlling bots in multiplayer campaigns, bringing the feature to parity with single-player.

**Server Configuration:**

- **Optional Feature:** Disabled by default when hosting/starting a server. Must be explicitly enabled in server settings (`AllowControllingBots`).
- Available in Sandbox, Mission, and Campaign game modes (including PermaDeath and Ironman). Not available in PvP.

**Key Features & Behavior:**

- **Bot Control:** Players can take control of bots directly during multiplayer matches if enabled by server settings.
- **Main Character Persistence:** The player's main character remains the sole character tied to respawning.
- **Death Handling:**
  - If the main character dies while controlling a bot, control automatically switches back to the main character to display the death popup.
  - Death popups are suppressed when a controlled bot dies.
- **UI Indicators:** Added an "owner" icon to the crew list to clearly show which bots are actively being controlled and by whom.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,15 @@
+{
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "name": "Launch Barotrauma",
+            "type": "coreclr",
+            "request": "launch",
+            "program": "${workspaceFolder}/Barotrauma/bin/DebugWindows/net8.0/Barotrauma.exe",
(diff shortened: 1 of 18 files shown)


erunehtar:
the feature needs testing, only tried with 2-3 people on server, but we were able to complete the campaign with these changes as-is

